PR #2, opened 2019-07-06
Add rquery paragraph

krlmlr:
@JohnMount: Is this an accurate description?


JohnMount:
That is great, thank you very much.

By the way I am interested in what reception dm gets.  Do people see the problem you are solving or not?

krlmlr:
Thanks. I think the README needs an intro example why we are working with more than one table in the first place. In some software systems this is far from common.

PR #9, opened 2019-07-22
Simplify internal data structure

krlmlr:
so that operations on dm objects are easier to implement.


R/db-helpers.R
```diff
@@ -47,26 +47,22 @@ create_queries <- function(
...
   } else if (is_postgres(dest)) {
     pmap_chr(
       list(
         db_tables,
-        cols_to_set_not_null,
-        cols_db_classes
+        cols_to_set_not_null
       ),
-      ~ glue("ALTER TABLE {..1} ALTER COLUMN {..2} TYPE {..3}, ALTER COLUMN {..2} SET NOT NULL")
+      ~ glue("ALTER TABLE {..1} ALTER COLUMN {..2} SET NOT NULL")
```
TSchiefer: for Postgres your approach works.

I am not sure if for MSSQL one can avoid the need for the data type. I checked a few tutorials for rendering columns NOT NULLABLE, but everywhere it seems the data type is used. 

If we do need it for MSSQL we can either:
- do it like it was done before and store the classes in the 'dm' or
- determine the class on the fly when dealing with MSSQL

R/db-helpers.R
```diff
@@ -88,31 +84,7 @@ queries_set_pk_cols <- function(dest, pk_information) {
...
-        db_child_tables,
-        cols_to_adapt,
-        cols_db_classes
-      ),
-      ~ glue("ALTER TABLE {..1} ALTER COLUMN {..2} TYPE {..3}")
-    )
-  } else {
-    return("")
-  }
+  ""
```
TSchiefer: Your approach here once more works for Postgres, but I am quite certain that for MSSQL it will fail (this was most likely the reason why I added these... but we can leave it out for Postgres)

R/dm.R
```diff
@@ -70,11 +70,34 @@ new_dm <- function(src, tables, data_model) {
...
+  }
+
   structure(
     list(
       src = src,
       tables = tables,
-      data_model = data_model
+      data_model_tables = data_model$tables,
+      data_model_keys = keys,
+      data_model_references = references
```
TSchiefer: Maybe we talked about this already, but it does seem very complicated to always have to change between `data_model` and the three parts needed for our class. For now it seems good though.

Only thing to consider is if we want to store the data types as well, since we seem to need it for MSSQL, but actually we can just get these easily enough from the tables themselves.

R/dm.R
```diff
@@ -134,7 +157,35 @@ cdm_get_tables <- function(x) {
...
+    mutate(type = "integer") %>%
+    left_join(keys, by = c("table", "column")) %>%
+    mutate(key = coalesce(key, 0L)) %>%
+    left_join(references_for_columns, by = c("table", "column"))
+
+  new_data_model(
+    x$data_model_tables,
+    columns,
+    references
+  )
```
TSchiefer: see last comment

R/db-helpers.R
krlmlr: Rewritten to set the `NOT NULL` modifier during creation of the table.

PR #10, opened 2019-07-22
fix function 'create_graph_from_dm()'

TSchiefer:
Had a bug in the function, that would lead to problems, once we have cycles.


krlmlr:
Thanks!

PR #11, opened 2019-07-22
improvements in `cdm_find_conn_tbls()`

TSchiefer:
- returns now all nodes on simple paths
- redundant work was removed (`crossing()` is being filtered first now)


TSchiefer:
Agreed, let's brainstorm. I gotta say though, that I have the feeling, in relational data models the number of possible paths would very rarely be a bottleneck for performance or a source of errors.

I will implement the second part of the changes in another branch if that's ok. That refers to:
- redundant work was removed (crossing() is being filtered first now)

krlmlr:
Closing for now.

PR #16, opened 2019-07-25
reduce to unique paths

TSchiefer:
until we find a better solution we can just as well optimize the suboptimal solution that we have(?)

``` r
library(tidyverse)

m <- c("one", "two", "three")

crossing(m, m)
#> # A tibble: 9 x 2
#>   m     m1   
#>   <chr> <chr>
#> 1 one   one  
#> 2 one   three
#> 3 one   two  
#> 4 three one  
#> 5 three three
#> 6 three two  
#> 7 two   one  
#> 8 two   three
#> 9 two   two

crossing(from = m, to = m) %>% 
  filter(from < to)
#> # A tibble: 3 x 2
#>   from  to   
#>   <chr> <chr>
#> 1 one   three
#> 2 one   two  
#> 3 three two
```

<sup>Created on 2019-07-25 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/select_tables_from_dm.R
+++ b/R/select_tables_from_dm.R
@@ -72,10 +72,10 @@ cdm_find_conn_tbls <- function(dm, ...) {
     abort_vertices_not_connected()
   }
 
-  V_ids <- map_int(table_names, ~ which(V == .x))
-  all_comb <- crossing(table_names, V_ids)
-  ids_vec <- pull(all_comb, V_ids)
-  names_vec <- pull(all_comb, table_names)
+  all_comb <- crossing(from = table_names, to = table_names) %>%
+    filter(from < to)
+  ids_vec <- pull(all_comb, from) %>% map(~(V == .)) %>% map_int(which)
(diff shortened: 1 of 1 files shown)


R/select_tables_from_dm.R
```diff
@@ -72,10 +72,10 @@ cdm_find_conn_tbls <- function(dm, ...) {
     abort_vertices_not_connected()
   }
 
-  V_ids <- map_int(table_names, ~ which(V == .x))
-  all_comb <- crossing(table_names, V_ids)
-  ids_vec <- pull(all_comb, V_ids)
-  names_vec <- pull(all_comb, table_names)
+  all_comb <- crossing(from = table_names, to = table_names) %>%
+    filter(from < to)
+  ids_vec <- pull(all_comb, from) %>% map(~(V == .)) %>% map_int(which)
```
krlmlr: Can we use `map()` instead of `map(map(which))` ?

PR #17, opened 2019-07-26
explain rejected candidates

TSchiefer:
- `enum_pk_candidates()`, `cdm_enum_pk_candidates()` and `cdm_enum_fk_candidates()` explain now rejections in a new column `why` (admittedly rather superficially)


TSchiefer:
(BTW, just wondering if "list" wouldn't be nicer than "enum" )

TSchiefer:
cf. #17

PR #27, opened 2019-08-08
better error description in case of numeric subsetting

### LINKED ISSUE #18: A more helpful error message when using numeric indices on dm objects ?
Thanks for the fantastic package.

I think the following error could be improved :

``` r
library(dm)
library(dplyr,warn.conflicts = FALSE)
#> Warning: package 'dplyr' was built under R version 3.6.1
dm0 <- as_dm(lst(iris,cars))
dm0[[1]]
#> Can't convert a double vector to a string
```

<sup>Created on 2019-07-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

```
dm:::`[[.dm`
#> function (x, name) 
#> {
#>     table <- as_string(name)
#>     tbl(x, table)
#> }
```

My understanding is that `dm` objects are lists rather than environments because you want to pass them by copy, but at the same time you'd rather not allow numeric indices because order doesn't make sense in a DB.

In that case I would suggest something like "Can't subset a dm object by position, either subset by name or use `cdm_get_tables()` to convert to a regular list first."

-------------------

Note that the choice not to allow numeric indices make the operation : `map(dm0, nrow)` fail, which might be unexpected by a user who had previously success running `nrow(dm0[["iris"]])` and `nrow(dm0[["cars"]])` .

TSchiefer:
closes #18

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -272,7 +272,7 @@ print.dm <- function(x, ...) {
 
 #' @export
 `[[.dm` <- function(x, name) {
-  if (class(name) == "integer" || class(name) == "numeric") abort_no_numeric_subsetting()
+  if (is.numeric(name)) abort_no_numeric_subsetting()
   table <- as_string(name)
   tbl(x, table)
 }
@@ -286,7 +286,7 @@ print.dm <- function(x, ...) {
 
 #' @export
 `[.dm` <- function(x, name) {
-  if (class(name) == "integer" || class(name) == "numeric") abort_no_numeric_subsetting()
+  if (is.numeric(name)) abort_no_numeric_subsetting()
   tables <- as_character(name)
   cdm_select_tbl(x, !!!tables)
 }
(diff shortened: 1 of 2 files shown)


R/dm.R
```diff
@@ -272,6 +272,7 @@ print.dm <- function(x, ...) {
 
 #' @export
 `[[.dm` <- function(x, name) {
+  if (class(name) == "integer" || class(name) == "numeric") abort_no_numeric_subsetting()
```
```suggestion
  if (is.numeric(name)) abort_no_numeric_subsetting()
```

`class()` can also have length > 1.

R/dm.R
```diff
@@ -285,6 +286,7 @@ print.dm <- function(x, ...) {
 
 #' @export
 `[.dm` <- function(x, name) {
+  if (class(name) == "integer" || class(name) == "numeric") abort_no_numeric_subsetting()
```
```suggestion
  if (is.numeric(name)) abort_no_numeric_subsetting()
```

krlmlr:
Thanks!

PR #31, opened 2019-08-13
new {datamodelr} print method does not well with tibbles

TSchiefer:
``` r
library(tibble)
t <- tibble(a = rep(0, 3))

# result should be 0
length(unique(t[t$a != 0, "a"]))
#> [1] 1
df <- as.data.frame(t)

# this works as intended
length(unique(df[df$a != 0, "a"]))
#> [1] 0
```

<sup>Created on 2019-08-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


krlmlr:
Thanks!

PR #34, opened 2019-08-21
mention related work in README

jangorecki:
Great to see projects aiming for data warehousing concepts in R!
Adding related project to list of mentioned works in that field so far.


krlmlr:
Thanks!

PR #36, opened 2019-08-21
repair 'browse_docs()'

TSchiefer:
- replace current solution with `browseURL("https://krlmlr.github.io/dm/")`

not sure if this is the right thing to do, but in any case it doesn't work with the current implementation

an alternative might be `pkgdown::preview_site()`


R/pkgdown.R
```diff
@@ -2,14 +2,8 @@
...
 #' @keywords internal
 #' @export
 browse_docs <- function() {
-  path <- get_docs_path()
-  browseURL(path)
-}
-
-get_docs_path <- function() {
-  system.file(package = packageName(), "pkgdown", "index.html")
+  browseURL("https://krlmlr.github.io/dm/")
```
krlmlr:
```suggestion
  utils::browseURL("https://krlmlr.github.io/dm/")
```

krlmlr:
Thanks! `browse_docs()` is a hack when documentation isn't hosted anywhere yet.

PR #43, opened 2019-09-05
F rename cols

### LINKED ISSUE #41: Implement cdm_rename()
Required for #40. Same syntax as `cdm_filter()` .

`cdm_select()` is similar but has the additional complexity that removed columns need to be taken care of. Later.

TSchiefer:
- function renames columns of tables, taking care of keys

closes #41


R/data-model-helpers.R
```diff
@@ -245,3 +245,50 @@ data_model_db_types_to_R_types <- function(data_model) {
   data_model$columns$type <- new_type
   data_model
 }
+
+upd_pks_after_rename <- function(
+  pks,
```
krlmlr: This needs to come right after `function(` .

R/data-model-helpers.R
```diff
@@ -245,3 +245,50 @@ data_model_db_types_to_R_types <- function(data_model) {
...
+  table_name,
+  list_of_renames) {
+
+  pk_of_table <- filter(pks, table == !!table_name) %>% pull(column)
```
krlmlr: We could also express with `left_join()` and `coalesce()` : 

```r
rename_logic <- ...
new_pks <-
  pks %>%
  left_join(rename_logic) %>%
  mutate(table = coalesce(new_table, table))
```

No need to spend time here, though.

R/data-model-helpers.R
```diff
@@ -245,3 +245,50 @@ data_model_db_types_to_R_types <- function(data_model) {
...
+  if (any(list_of_renames == pk_of_table)) {
+    new_pk <- names(list_of_renames[list_of_renames == pk_of_table])
+    pks$column[pks$table == table_name] <- new_pk
+  }
+  pks
+}
+
+upd_fks_after_rename <- function(
```
krlmlr: This is where I think a longer-form data structure with three columns (table, column, role) would fare much better: we'd only need one update routine. TBD.

R/dplyr-dm.R
```diff
@@ -0,0 +1,65 @@
+#' Rename one or more columns of a [`dm`] table
+#'
+#' Rename columns of your [`dm`] with a similar syntax to `dplyr::rename()`.
+#'
+#' @inheritParams cdm_filter
+#' @param ... One or more unquoted expressions separated by commas. You can treat
```
krlmlr: Instead of copy-pasting, would it work to use `@inheritParams dplyr::select` ?

R/dplyr-dm.R
```diff
@@ -0,0 +1,65 @@
...
+
+  fks_upd <-
+    upd_fks_after_rename(
+      cdm_get_data_model_fks(dm),
+      table_name,
+      list_of_renames
+    )
+
+  new_dm2(
+    src = cdm_get_src(dm),
```
krlmlr: Covered by `base_dm` ?

R/dplyr-dm.R
TSchiefer: I slightly changed the text here and there, 2 reasons:
1. We don't have `cdm_select()` yet
2. Some things would not be quite right

don't have a strong opinion though

R/data-model-helpers.R
TSchiefer: This as in `pks,`?

R/data-model-helpers.R
```diff
@@ -245,3 +245,50 @@ data_model_db_types_to_R_types <- function(data_model) {
...
+  fks,
+  table_name,
+  list_of_renames) {
+
+  fks_from_table <- filter(fks, table == !!table_name) %>% pull(column)
+  if (!is_empty(fks_from_table) &&
```
krlmlr: Can you extract this and the block below into a function?

PR #47, opened 2019-09-09
avoid column name conflicts

### LINKED ISSUE #40: Implement cdm_disambiguate_cols()
Goal: rename all columns that exist in more than one table. Prefixed by the table name with a separator, `.` as default.

With optional verbose output (`quiet = FALSE`).

Should also take care of updating pk and fk.

TSchiefer:
function `cdm_disambiguate_cols()`:
- finds ambiguous column names
- replaces them by `paste0(table_name, separator, column_name)`
- optional message with list of renamed columns

closes #40


tests/testthat/helper-src.R
```diff
@@ -105,24 +105,24 @@ data_3_src <- test_load(data_3)
 
 # for table-surgery functions ---------------------------------------------
 data_ts <- tibble(
-  a = as_integer(c(1, 2, 1)),
+  a = vctrs::vec_cast(c(1, 2, 1), to= integer()),
```
krlmlr: `as.integer()` still works without warning. Can you please change?

R/disambiguate.R
```diff
@@ -0,0 +1,45 @@
...
+#' @inheritParams cdm_add_pk
+#' @param sep The character variable used in between the table names and the ambiguous column names
+#' @param quiet Boolean, if `TRUE`, will list the renamed columns in a message
+#'
+#' @examples
+#' cdm_disambiguate_cols(cdm_nycflights13())
+#'
+#' @export
+cdm_disambiguate_cols <- function(dm, sep = ".", quiet = FALSE) {
+  tbl_cols_for_disambiguation <-
```
krlmlr: Use `recipe` as name and refer later as `recipe$table` and `recipe$renames` (instead of `*_for_disambiguation`).

TSchiefer: resolved in https://github.com/krlmlr/dm/tree/f-outstanding-from-disambiguate

PR #48, opened 2019-09-10
Flatten `dm` to wide table

### LINKED ISSUE #13: Multi-joins
with `cdm_join_tbl()` we are able to join two tables of a `dm` which share a direct link (FK).

what if we want to join a further table from the same `dm` in a second step etc.?

A solution would be a new function, which wraps `cdm_join_tbl()` and then returns the original `dm`, but with the result of the join as an additional table. This way, there would be the possibility of adding a secondary join to this table in the pipeline.

Questions would be: 
- what will the PK of the new table be (depends probably on the type of the join and the type of the relation between the joined tables)?
- how to determine the FKs from and to the newly created table?

TSchiefer:
Function `cdm_flatten()`:
1. Applies all filters
2. Disambiguates columns
3. Renames all FK columns so they match the corresponding PK columns
4. Finds an order for joining all tables together
5. Joins all tables together

Mind that this branch is based on https://github.com/krlmlr/dm/tree/f-disambiguate

Closes #13


R/flatten.R
```diff
@@ -0,0 +1,118 @@
...
+  clean_dm <- cdm_apply_filters(dm) %>%
+      cdm_disambiguate_cols(sep = sep, quiet = TRUE) %>%
+      adapt_fk_cols()
+  g <- create_graph_from_dm(clean_dm, directed = TRUE)
+  filtered_tables <- clean_dm %>% cdm_get_tables()
+
+  # chose a "child table" (only outgoing FKs) as a start (otherwise we might
+  # end up with empty rows, if "parent/dim table" contains PK values that are
+  # not present in child table (we might anyway though, depending on `join`-type))
+  initial_LHS_name <- g %>%
```
krlmlr: `names(igraph::topo_sort(g))[[1]]` -- let's reserve the pipe for data frame manipulations.

(What do we do with a `dm` with zero tables?)

R/flatten.R
```diff
@@ -0,0 +1,118 @@
...
+    igraph::topo_sort() %>%
+    names() %>%
+    pluck(1)
+  initial_LHS <- filtered_tables[[initial_LHS_name]]
+
+  # each next table needs to be accessible from the former table; this is not ensured by `topo_sort()`
+  # we achieve this with a breadth-first-search (BFS)
+  order <- igraph::bfs(g, initial_LHS_name) %>%
```
krlmlr: I think a depth-first search (`igraph::dfs()`) is a slightly better choice here, because it would leave the chains intact.

R/flatten.R
```diff
@@ -0,0 +1,118 @@
...
+    extract2("order") %>%
+    names()
+
+  # early return in case of only one table...
+  if (length(order) == 1) return(initial_LHS)
```
krlmlr: The zero-table dm is the important corner case here, `length(order)` should be handled transparently.

R/flatten.R
```diff
@@ -0,0 +1,118 @@
...
+  # since we want to work with `reduce2()`, the `.init`-parameter is the first table
+  # in `order` and the next table is always the join partner
+  ordered_table_list <- filtered_tables[order] %>%
+    extract(2:length(order))
+
+  # get the right column names for each join (should all be adapted by `adapt_fk_cols()`)
+  by <- get_by_for_flatten(clean_dm, order)
```
krlmlr: Can we make use of the edge list optionally returned by `igraph::dfs()` ?

tests/testthat/test-flatten.R
```diff
@@ -0,0 +1,9 @@
+test_that("`cdm_flatten()` does the right thing", {
```
krlmlr: Can you please add a test for a one-table dm and a zero-table dm?

R/flatten.R
```diff
@@ -0,0 +1,118 @@
...
+#' @return A wide table resulting of consecutively joining all tables together.
+#'
+#' @examples
+#' cdm_nycflights13() %>%
+#' cdm_select_tbl(-weather) %>%
+#' cdm_flatten()
+#'
+#' @export
+cdm_flatten <- function(dm, join = left_join, sep = ".") {
+  if (!are_all_vertices_connected(create_graph_from_dm(dm), src_tbls(dm))) {
```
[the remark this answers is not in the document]
TSchiefer: as discussed, I will write a comment

R/flatten.R
TSchiefer: Guess in the result the only difference would be the order of the columns. But you are right, the order will most likely make more sense thematically with DFS.

R/flatten.R
```diff
@@ -0,0 +1,118 @@
...
+  # we achieve this with a breadth-first-search (BFS)
+  order <- igraph::bfs(g, initial_LHS_name) %>%
+    extract2("order") %>%
+    names()
+
+  # early return in case of only one table...
+  if (length(order) == 1) return(initial_LHS)
+  # since we want to work with `reduce2()`, the `.init`-parameter is the first table
+  # in `order` and the next table is always the join partner
+  ordered_table_list <- filtered_tables[order] %>%
```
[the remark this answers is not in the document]
TSchiefer: Do you want the line `if (length(order) == 1) return(initial_LHS)` to be removed? (just asking, because if it stays the code below doesn't matter anyway in case of just one table)

R/flatten.R
```diff
@@ -0,0 +1,118 @@
...
+#' @param lhs The table on the left hand side of the join
+#' @param rhs The table on the right hand side of the join
+#' @param join The type of join to be performed, see \code{\link[dplyr]{join}}
+#'
+#' @return The resulting table of the join
+#'
+#' @family Flattening functions
+#'
+#' @export
+cdm_join_tbl <- function(dm, lhs, rhs, join = semi_join) {
```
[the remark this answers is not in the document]
TSchiefer: Good point, only arguments that come to mind is, that 
1. the function name might seem more intuitive to ppl, who want to join 2 tbls from `dm`
2. the `cdm_select_tbl()` call is not needed

I will implement for now according to your suggestion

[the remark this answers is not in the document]
TSchiefer: `cdm_select_tbl()` needs the table names, and `{{ lhs }}` does not seem to do that trick. I would use `as_name(substitute(lhs))` if that is alright, but I am open to suggestions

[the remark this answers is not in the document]
TSchiefer: Sorry, I meant `as_name(enexpr(lhs))`, not with `substitute()`

R/flatten.R
TSchiefer: Throw an error if trying to flatten an empty list of tables? we could call in the beginning `is_empty(cdm_get_tables(dm))`. Or maybe better: 
```r
list_of_tables <- cdm_get_tables(dm)
if (is_empty(cdm_get_tables(dm))) ...
if (nrow(list_of_tables == 1)) return(...)
```
Then we'd have all (or most) corner cases sorted

R/flatten.R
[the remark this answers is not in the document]
krlmlr: I prefer to avoid redundant code where possible. I'm fine with a comment that indicates that the operations afterwards are a no-op in this case.

R/flatten.R
krlmlr: A 0x0 tibble seems like the logical result. Do we really need to special-case `nrow() == 1` ?

R/flatten.R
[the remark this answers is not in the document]
TSchiefer: I'm rather certain that latest when the joins are to be made an error occurs if a join should be made between a table and an empty list. Otherwise put, I am not sure how to avoid the early return line.

R/flatten.R
krlmlr: This corner case no longer occurs with the current interface.

R/flatten.R
krlmlr: No longer necessary.

PR #49, opened 2019-09-10
return named list with `cdm_nrow()`?

TSchiefer:
Description of the function promises a named list, but the sum of the row numbers was returned instead.


krlmlr:
Thanks!

Issue #57, opened 2019-09-14
Improve internal data structure again


krlmlr:
To make it easy to maintain internal consistency, I think the internal data structure needs to be a tibble with the following columns:

- Table name (character)
- Table object (data frame/tbl/...) (list)
- Color (character)
- Filter expressions (list)
- Primary keys (list of character)
- Child tables (inverse of foreign keys; list of tibbles of the form "table name + list of character")

With this, it seems that we only need two "restore" functions:

- Fix table names after a table select/rename
- Fix column names after a column select/rename

These "restore" functions accept a tidyselect-like recipe that describes how to handle renames. These can be passed on to `dplyr::recode()` to implement the actual recoding.

PR #67, opened 2019-09-19
remove src from `dm` since implicitly available

### LINKED ISSUE #38: Get rid of src component
It's not really needed. We only need to check `same_src()` for all input tables.

TSchiefer:
- remove `src` from `dm`
-  add some checks for same src of tables
- add 2 classed errors
- `cdm_get_src()` for local `dm` returns `.GlobalEnv`

closes #38

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -282,7 +282,7 @@ as_dm.default <- function(x) {
 
 tbl_src <- function(x) {
   if (is.data.frame(x)) {
-    "local"
+    NULL
   } else if (inherits(x, "tbl_sql")) {
     x$src
   } else {
@@ -304,8 +304,10 @@ format.dm <- function(x, ...) {
 #' @import cli
 print.dm <- function(x, ...) {
   cat_rule("Table source", col = "green")
+  src <- cdm_get_src(x)
+  if (is_null(src)) src <- "local"
 
-  db_info <- strsplit(format(cdm_get_src(x)), "\n")[[1]][[1]]
+  db_info <- strsplit(format(src), "\n")[[1]][[1]]
 
   cat_line(db_info)
 
--- a/R/dm.R
+++ b/R/dm.R
@@ -492,5 +492,5 @@ cdm_reset_all_filters <- function(dm) {
 all_same_source <- function(tables) {
   # Use `NULL` if `tables` is empty
   first_table <- tables[1][[1]]
-  is.null(detect(tables[-1], ~ same_src(., first_table)))
+  is.null(detect(tables[-1], ~ !same_src(., first_table)))
 }
--- a/R/dm.R
+++ b/R/dm.R
@@ -148,8 +148,7 @@ validate_dm <- function(x) {
 #' @export
 cdm_get_src <- function(x) {
   tables <- cdm_get_tables(x)
-  if (!all_same_source(tables)) abort_not_same_src()
-  tbl_src(tables[[1]])
+  if (!is_empty(tables)) tbl_src(tables[[1]]) else default_local_src()
 }
 
 #' Get tables component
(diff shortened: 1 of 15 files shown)


R/dm.R
```diff
@@ -91,26 +90,24 @@ new_dm <- function(src, tables, data_model) {
...
+new_dm2 <- function(tables = cdm_get_tables(base_dm),
                     data_model_tables = cdm_get_data_model_tables(base_dm),
                     pks = cdm_get_data_model_pks(base_dm),
                     fks = cdm_get_data_model_fks(base_dm),
                     filter = cdm_get_filter(base_dm),
                     base_dm) {
-
-  stopifnot(!is.null(src))
   stopifnot(!is.null(tables))
+  if (!all_same_source(tables)) abort_not_same_src()
```
krlmlr: This constructor should be fast, and not perform unnecessary checks. This check should be part of `validate_dm()`.

See https://adv-r.hadley.nz/s3.html for principles we should abide by.

R/dm.R
```diff
@@ -151,7 +148,9 @@ validate_dm <- function(x) {
 #'
 #' @export
 cdm_get_src <- function(x) {
-  unclass(x)$src
+  tables <- cdm_get_tables(x)
+  if (!all_same_source(tables)) abort_not_same_src()
```
krlmlr: A getter should never err, this is the job of the validator.

R/dm.R
```diff
@@ -151,7 +148,9 @@ validate_dm <- function(x) {
...
+  tbl_src(tables[[1]])
```
krlmlr: If `is_empty(tables)`, we could return `dplyr::src_df(env = .GlobalEnv)` .

R/dm.R
```diff
@@ -270,29 +269,24 @@ as_dm.default <- function(x) {
...
   data_model <- datamodelr::dm_from_data_frames(proxies)
 
-  new_dm(src, x, data_model)
+  new_dm(x, data_model)
 }
 
 tbl_src <- function(x) {
   if (is.data.frame(x)) {
-    src_df(env = new_environment(x))
+    NULL
```
krlmlr: `dplyr::src_df(env = .GlobalEnv)` -- can you extract a function for the default local source?

R/dm.R
```diff
@@ -310,8 +304,10 @@ format.dm <- function(x, ...) {
 #' @import cli
 print.dm <- function(x, ...) {
   cat_rule("Table source", col = "green")
+  src <- cdm_get_src(x)
+  if (is_null(src)) src <- "local"
```
krlmlr: Would be covered by `cdm_get_src()`

R/dm.R
```diff
@@ -493,3 +484,7 @@ cdm_rename_tbl <- function(dm, ...) {
     .init = dm
   )
 }
+
+all_same_source <- function(tables) {
+  all(map_lgl(tables, same_src, tables[[1]]))
```
krlmlr: - `detect()` aborts on first match
- Introduce pattern to deal with 0-length list

```suggestion
  # Use `NULL` if `tables` is empty
  first_table <- tables[1][[1]]
  is.null(detect(tables[-1], ~ same_src(., first_table)))
```

vignettes/dm-class-and-basic-operations.Rmd
```diff
@@ -35,19 +35,18 @@ Let's first have a look at the new class:
 The `dm` class contains a set of tables as well as information about their primary keys and the relations between the tables. 
 You can have your `dm` object on different data sources, such as a local environment or a database.
 
-`dm` is a wrapper around two classes, which are independent from each other:
-
-1. part of the object is a [`src` object](https://dplyr.tidyverse.org/reference/src.html) from {dplyr}
-2. the other part is a class `data_model` object
+`dm` is a wrapper around the two `data_model` object
```
krlmlr: We're moving away from this. It's more important to understand that we handle tables, relationships, and now also filters.

The text works "as is", let's wait until #57 before revisiting.

PR #68, opened 2019-09-20
detailed 'cdm_enum_fk_candidates()'

### LINKED ISSUE #24: Explain in more detail why column is not a foreign key
Mention which values are not contained in the referred primary key. In #17 I already added printing of duplicate values for primary key candidates.

TSchiefer:
- checks now for class compatibility
- if compatible, checks if subset
- lists in column `why`:
    1. nothing if candidate
    1. class `...` doesn't match pk-col class `...`
    1. following values are not in pk-col: ...

fixes #24

### DIFF
--- a/R/foreign-keys.R
+++ b/R/foreign-keys.R
@@ -196,29 +196,53 @@ cdm_enum_fk_candidates <- nse_function(c(dm, table, ref_table), ~ {
   }
   ref_tbl <- tbl(dm, ref_table_name)
   tbl <- tbl(dm, table_name)
+  rows_tbl <- count(tbl) %>% pull()
   tbl_colnames <- colnames(tbl)
 
   tibble(
     column = tbl_colnames,
-    why = map_chr(column, ~check_fk(dm, tbl, .x, ref_tbl, ref_table_name, ref_tbl_pk))
+    why = map_chr(column, ~check_fk(dm, tbl, table_name, .x, rows_tbl, ref_tbl, ref_table_name, ref_tbl_pk))
   ) %>%
     mutate(candidate = ifelse(why == "", TRUE, FALSE)) %>%
     select(column, candidate, why) %>%
-    mutate(arrange_col = str_sub(why, 1, 3)) %>%
-    arrange(desc(candidate), desc(arrange_col), column) %>%
+    mutate(arrange_col = as.integer(str_extract(why, "^[0-9]*"))) %>%
+    arrange(desc(candidate), arrange_col, column) %>%
     select(-arrange_col)
 })
 
-check_fk <- function(dm, t1, colname, t2, t2_name, pk) {
-  names(pk) <- colname
-  test <- tryCatch(anti_join(
-    select(t1, !!sym(colname)), select(t2, !!sym(pk)), by = pk) %>%
-      utils::head(MAX_COMMAS + 1) %>%
-      pull(), error = identity)
-  if (is_condition(test)) {
-    return(conditionMessage(test))
+check_fk <- function(dm, t1, t1_name, colname, rows_t1, t2, t2_name, pk) {
+
+  t1_join <- t1 %>% select(value = !!sym(colname))
+  t2_join <- t2 %>% select(value = !!sym(pk)) %>% mutate(match = 1L)
+
+  res_tbl <- tryCatch(
+    left_join(t1_join, t2_join, by = "value") %>%
+      mutate(mismatch_or_null = if_else(is.na(match), value, NULL)) %>%
+      count(mismatch_or_null) %>%
+      filter(!is.na(mismatch_or_null)) %>%
+      arrange(desc(n)),
+    error = identity)
+  if (is_condition(res_tbl)) {
+    return(conditionMessage(res_tbl))
   }
-  if (is_empty(test)) return("")
-  test_formatted <- commas(format(test, trim = TRUE, justify = "none"))
-  glue("values not in {tick(glue('{t2_name}${pk}'))}: {test_formatted}")
+  # "na.rm = TRUE", to suppress standard warning message: "Missing values are always removed in SQL."
+  # Second `tryCatch()` is necessary, because on some DBs (e.g. Postgres) mismatched types don't give an immediate error,
+  # but only once the SQL is executed (`summarize(...)`)
+  rows_missing <- tryCatch(summarize(res_tbl, sum(n, na.rm = TRUE)) %>% pull(),
+                           error = identity)
+  if (is_condition(rows_missing)) {
+    return(conditionMessage(rows_missing))
+  }
+
+  if (is.na(rows_missing) || rows_missing == 0) return("")
+  percentage_missing <- as.character(round((rows_missing / rows_t1) * 100, 1))
+  vals_extended <- res_tbl %>%
+    head(MAX_COMMAS + 1) %>%
+    mutate(num_mismatch = paste0(mismatch_or_null, " (", n, ")")) %>%
+    # FIXME: this fails on SQLite, why?
+    # mutate(num_mismatch = glue("{as.character(mismatch_or_null)} ({as.character(n)})")) %>%
+    pull()
+  vals_formatted <- commas(format(vals_extended, trim = TRUE, justify = "none"))
+  glue("{as.character(rows_missing)} entries ({percentage_missing}%) of ",
+       "{tick(glue('{t1_name}${colname}'))} not in {tick(glue('{t2_name}${pk}'))}: {vals_formatted}")
 }
--- a/R/foreign-keys.R
+++ b/R/foreign-keys.R
@@ -152,18 +152,24 @@ cdm_rm_fk <- function(dm, table, column, ref_table) {
 
 #' Find foreign key candidates in a table
 #'
-#' Which columns are candidates for a foreign key of a table, referencing
+#' Which columns are good candidates as a foreign key of a table, referencing
 #' the primary key column of another [`dm`] object's table?
 #'
 #' @inheritParams cdm_add_fk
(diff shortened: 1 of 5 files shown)


R/foreign-keys.R
```diff
@@ -189,20 +192,54 @@ cdm_enum_fk_candidates <- nse_function(c(dm, table, ref_table), ~ {
...
+  # which columns contain subset of values of PK-col?
+  col_candidates <- cols_class_match[map_lgl(
+    cols_class_match,
+    ~ is_subset(tbl, !!.x, ref_tbl, !!ref_tbl_pk))]
+  tibble_candidates <- tibble(column = col_candidates,
+    candidate = TRUE,
+    why = "")
+
+  # which columns of same class contain additional values and which ones?
+  set_of_pk_vals <- pull(arrange(distinct(ref_tbl, !!ensym(ref_tbl_pk))))
```
krlmlr: Can you please use the following idiom to compute mismatches:

```r
tryCatch(
  anti_join(...) %>%
     utils::head(MAX_COMMAS + 1) %>%
     collect(),
  error = identity
)
```

Then, check for `is_condition(.)` and, if yes, use `conditionMessage(.)` as reason.

tests/testthat/helper-src.R
```diff
@@ -8,10 +8,10 @@ message("for ccheck_cardinality...()")
 d1 <- tibble::tibble(a = 1:5, b = letters[1:5])
 d2 <- tibble::tibble(a = c(1, 3:6), b = letters[1:5])
 d3 <- tibble::tibble(c = 1:5)
-d4 <- tibble::tibble(c = c(1:5, 5))
+d4 <- tibble::tibble(c = c(1:5, 5L))
 d5 <- tibble::tibble(a = 1:5)
 d6 <- tibble::tibble(c = 1:4)
-d7 <- tibble::tibble(c = c(1:5, 5, 6))
+d7 <- tibble::tibble(c = c(1:5, 5L, 6L))
```
krlmlr: Can you tweak the test so that we check integer/numeric compatibility?

TSchiefer: Just so I got it right: we want a test that compares numeric columns with integer columns. I just checked, what `anti_join()` does locally:
``` r
suppressPackageStartupMessages(library(dplyr))
library(tibble)
t1 <- tibble(a = as.numeric(1:5))
t1
#> # A tibble: 5 x 1
#>       a
#>   <dbl>
#> 1     1
#> 2     2
#> 3     3
#> 4     4
#> 5     5
t2 <- tibble(a = 1:5)
t2
#> # A tibble: 5 x 1
#>       a
#>   <int>
#> 1     1
#> 2     2
#> 3     3
#> 4     4
#> 5     5
anti_join(t1, t2, by = "a")
#> # A tibble: 0 x 1
#> # … with 1 variable: a <dbl>
```

<sup>Created on 2019-09-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
But on DB there might/will be differences. So is the test supposed to show the differences between DB- and local behaviour?

krlmlr: I'd expect this behavior for all databases.

R/foreign-keys.R
```diff
@@ -152,31 +152,34 @@ cdm_rm_fk <- function(dm, table, column, ref_table) {
...
 #' @param ref_table A table with a primary key.
 #'
-#' @description Which columns are foreign candidates of a table, referencing the primary key column of another [`dm`] object's table?
-#' `cdm_enum_fk_candidates()` checks first, if `ref_table` has a primary key set. Then it determines
-#' for each column of `table`, if this column contains only a subset of values of the primary key column of
-#' `ref_table` and is therefore a candidate for a foreign key from `table` to `ref_table`.
+#' @details `cdm_enum_fk_candidates()` checks first, if `ref_table` has a primary key set. Then
+#' it determines for each column of `table`, if it has the same class as the primary key of `ref_table`.
+#' If this is the case a check is performed, if the column contains only a subset of values of the
+#' primary key column of `ref_table`. If this is `TRUE`, this column is a candidate for a foreign key
```
krlmlr: Can you use bullets or an enumeration to describe the rules?

```suggestion
#' primary key column of `ref_table`.
#' This check uses [anti_join()] which implicitly checks type compatibility.
#' If this is `TRUE`, this column is a candidate for a foreign key
```

R/foreign-keys.R
```diff
@@ -185,23 +188,32 @@ cdm_enum_fk_candidates <- nse_function(c(dm, table, ref_table), ~ {
...
+why <- function(dm, t1, colname, t2, t2_name, pk) {
+  names(pk) <- colname
+  test <- tryCatch(anti_join(
+    select(t1, !!sym(colname)), select(t2, !!sym(pk)), by = pk) %>%
+      utils::head(MAX_COMMAS + 1) %>%
+      pull(), error = identity)
+  if (is_condition(test)) {
+    return(conditionMessage(test))
+  }
+  if (is_empty(test)) return("") else {
```
krlmlr: We don't need the `else` here.

R/foreign-keys.R
```diff
@@ -185,23 +188,32 @@ cdm_enum_fk_candidates <- nse_function(c(dm, table, ref_table), ~ {
...
+    why = map_chr(column, ~why(dm, tbl, .x, ref_tbl, ref_table_name, ref_tbl_pk))
+  ) %>%
+    mutate(candidate = ifelse(why == "", TRUE, FALSE)) %>%
+    select(column, candidate, why) %>%
+    mutate(arrange_col = str_sub(why, 1, 3)) %>%
+    arrange(desc(candidate), desc(arrange_col), column) %>%
+    select(-arrange_col)
 })
+
+why <- function(dm, t1, colname, t2, t2_name, pk) {
```
krlmlr:
```suggestion
check_fk <- function(dm, t1, colname, t2, t2_name, pk) {
```

tests/testthat/test-foreign-key-functions.R
```diff
@@ -159,23 +159,89 @@ test_that("cdm_rm_fk() works as intended?", {
...
+    "minute",         FALSE,      "Can't join on 'minute' x 'faa' because of incompatible types (numeric / character)",
+    "month",          FALSE,      "Can't join on 'month' x 'faa' because of incompatible types (integer / character)",
+    "sched_arr_time", FALSE,      "Can't join on 'sched_arr_time' x 'faa' because of incompatible types (integer / character)",
+    "sched_dep_time", FALSE,      "Can't join on 'sched_dep_time' x 'faa' because of incompatible types (integer / character)",
+    "year",           FALSE,      "Can't join on 'year' x 'faa' because of incompatible types (integer / character)",
+    "time_hour",      FALSE,      "cannot join a POSIXct object with an object that is not a POSIXct object"
+  )
+
+  expect_identical(
+    cdm_enum_fk_candidates(cdm_nycflights13(), flights, airports),
```
krlmlr:
```suggestion
    cdm_enum_fk_candidates(cdm_nycflights13(), flights, airports) %>%
      mutate(why = if_else(why != "", "<reason>", ""),
```

R/foreign-keys.R
```diff
@@ -185,23 +188,32 @@ cdm_enum_fk_candidates <- nse_function(c(dm, table, ref_table), ~ {
...
+  names(pk) <- colname
+  test <- tryCatch(anti_join(
+    select(t1, !!sym(colname)), select(t2, !!sym(pk)), by = pk) %>%
```
krlmlr: ~~We need to remove dupes on the server: anti_join %>% distinct %>% arrange %>% head %>% collect %>% pull~~

We also might want the total count and the mismatch percentage to detect weak relationships:

```r
t1_join <- t1 %>% select(value = !!sym(colname))
t2_join <- t2 %>% select(value = !!sym(pk)) %>% mutate(match = 1L)

left_join(t1_join, t2_join, by = "value") %>%
  mutate(mismatch_or_null = if_else(is.na(match), value, NULL)) %>%
  count(mismatch_or_null) %>%
  ... %>%
  collect()
```

Example result for `flights$dest`:

1334 rows (8.5%) of the values not in `airports$faa`: SJU (280), BQN (110), 3 more, …

R/foreign-keys.R
```diff
@@ -185,23 +194,55 @@ cdm_enum_fk_candidates <- nse_function(c(dm, table, ref_table), ~ {
...
+check_fk <- function(dm, t1, t1_name, colname, rows_t1, t2, t2_name, pk) {
+
+  t1_join <- t1 %>% select(value = !!sym(colname))
+  t2_join <- t2 %>% select(value = !!sym(pk)) %>% mutate(match = 1L)
+
+  res_tbl <- tryCatch(
+    left_join(t1_join, t2_join, by = "value") %>%
+      mutate(mismatch_or_null = if_else(is.na(match), value, NULL)) %>%
+      count(mismatch_or_null) %>%
+      filter(!is.na(mismatch_or_null)) %>%
```
krlmlr: This only removes one row, and I think we need it to compute percentages?

krlmlr:
```r
left_join(t1_join, t2_join, by = "value") %>%
  # if value is NULL, this also counts as a match -- consistent with fk semantics
  mutate(mismatch_or_null = if_else(is.na(match), value, NULL)) %>%
  count(mismatch_or_null) %>%
  ungroup() %>% # dbplyr problem?
  mutate(n_mismatch = sum(if_else(is.na(value, 0, 1), na.rm = TRUE))) %>%
  mutate(n_total = sum(n, na.rm = TRUE)) %>%
  arrange(desc(n)) %>%
  head(MAX_COMMAS + 1L) %>%
  collect()
```

PR #71, opened 2019-09-23
treatment of `semi_`- and `anti_join()` in `cdm_flatten_to_tbl()`

TSchiefer:
- in case of no filter in component, for the two join-methods the logic of `cdm_get_filtered_tables()` is used
- for `join = semi_join`, the result is the rows of `start` whose references are available in all parent tables
- for `join = anti_join` it's the opposite (`anti_join()` of original table with result of `semi_join()`); is basically those rows whose lookup-values that are missing at least once


TSchiefer:
closing, since we decided on a different implementation

PR #82, opened 2019-10-04
check referential integrity of a `dm`

### LINKED ISSUE #56: Provide way to check referential integrity
After b23b969d, we need an easy way to check integrity of all pk and fk constraints, with an option to show where integrity is violated:

- one function that returns a data frame with results of the integrity check
- one function that throws an error if integrity is violated, calls the first function

What's a good name?

TSchiefer:
`cdm_check_constraints()` does the following:
- checks if all PKs are unique keys
- checks if all values in FK-columns are a subset of their respective referenced columns
- returns a tibble:
     1. `table`: the table in the `dm`
     1. `kind`: "PK" or "FK"
     1. `column`: column of `table`
     1. `is_key`: logical
     1. `problem`: in case `is_key = FALSE` the reason for that

Closes #56.


R/error-helpers.R
```diff
@@ -450,3 +450,12 @@ error_what_a_weird_object <- function(class) {
   paste0("Don't know how to determine table source for object of class ",
          class)
 }
+
+abort_rename_plz <- function() {
+ abort(error_rename_plz(), .subclass = cdm_error_full("rename_plz"))
+}
+
+error_rename_plz <- function() {
+  paste0("`check_key()` can currently not deal with cases, when both variables `n` and `nn` are involved. ",
```
krlmlr: Where is this function used?

```suggestion
  paste0("`check_key()` can't currently deal with cases when both variables `n` and `nn` are involved. ",
```

R/key-helpers.R
```diff
@@ -1,3 +1,38 @@
+#' Validate your [`dm`]: are all key constraints met?
+#'
+#' This function returns a concise list, reporting for each key constraint if it is met (`TRUE`) or violated (`FALSE`).
```
krlmlr: Can we return a data frame?

R/key-helpers.R
```diff
@@ -27,8 +62,9 @@ check_key <- function(.data, ...) {
   data_q <- enquo(.data)
   .data <- eval_tidy(data_q)
   args <- exprs(...)
```
krlmlr: We could do `names(args) <- set_names(paste0("...", seq_along(args)))`, then we're sure that the `n` column is always returned. Can you please isolate these changes to a separate PR?

TSchiefer:
Should we issue a warning in case of violations? 

(I would think not, since this action is a pretty conscious decision and people will have at least a brief look at the result.)

R/key-helpers.R
```diff
@@ -62,6 +93,15 @@ is_unique_key <- nse_function(c(.data, column), ~ {
   duplicate_rows
 })
 
+is_key <- nse_function(c(.data, column), ~ {
+  col_expr <- ensym(column)
+  col_name <- as_name(col_expr)
+
+  duplicate_rows <- .data %>%
+    count(value = !!col_expr) %>%
+    filter(n != 1)
```
krlmlr: Does this work for remote data sources?

R/key-helpers.R
```diff
@@ -173,3 +213,31 @@ is_subset <- function(t1, c1, t2, c2) {
...
+  tbls <- map(set_names(table_names), ~ tbl(dm, .)) %>%
+    map2(syms(pks$pk_col), ~select(.x, !!.y))
+  tbl_is_pk <- map_dfr(tbls, enum_pk_candidates) %>%
+    rename(is_key = candidate, problem = why)
+
+  tibble(
+    table = table_names,
+    kind = "PK",
+    column = pks$pk_col
+  ) %>% left_join(tbl_is_pk, by = "column")
```
krlmlr:
```suggestion
  ) %>%
    left_join(tbl_is_pk, by = "column")
```

R/key-helpers.R
```diff
@@ -173,3 +213,31 @@ is_subset <- function(t1, c1, t2, c2) {
...
+check_fk_constraints <- function(dm) {
+  fks <-  left_join(cdm_get_all_fks(dm), cdm_get_all_pks(dm), by = c("parent_table" = "table"))
+  pts <- pull(fks, parent_table) %>% map(tbl, src = dm)
+  cts <- pull(fks, child_table) %>% map(tbl, src = dm)
+  fks_tibble <- mutate(fks, t1 = cts, t2 = pts) %>%
+    select(t1, t1_name = child_table, colname = child_fk_col, t2, t2_name = parent_table, pk = pk_col)
+  mutate(
+    fks_tibble, problem = pmap_chr(fks_tibble, check_fk),
+    is_key = if_else(problem == "", TRUE, FALSE),
+    kind = "FK") %>%
```
krlmlr:
```suggestion
    kind = "FK"
  ) %>%
```

R/key-helpers.R
TSchiefer: Actually, this function isn't used anymore in this version, so I deleted it. Thanks for drawing attention to it.

Regarding your question: the function `is_unique_key()` uses the same logic (it is used for `enum_pk_candidates()`). I just verified that it does work on Postgres.

TSchiefer:
added tests on DBs now (with option  'set_key_constraints = FALSE', since otherwise DB complains already by itself)

krlmlr:
Good work!

I think we need an own `print()` method that highlights the problems. Let's consider this in a future issue.

``` r
library(dm)
cdm_check_constraints(cdm_nycflights13())
#> # A tibble: 6 x 5
#>   table   kind  column  is_key problem                                     
#>   <chr>   <chr> <chr>   <lgl>  <chr>                                       
#> 1 flights FK    tailnum FALSE  50094 entries (14.9%) of `flights$tailnum` …
#> 2 airlin… PK    carrier TRUE   ""                                          
#> 3 airpor… PK    faa     TRUE   ""                                          
#> 4 planes  PK    tailnum TRUE   ""                                          
#> 5 flights FK    carrier TRUE   ""                                          
#> 6 flights FK    origin  TRUE   ""
```

<sup>Created on 2019-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Thanks!

PR #83, opened 2019-10-04
repair behaviour for named character vector as param for 'check_key()'

TSchiefer:
`check_key()` now deals correctly with named column lists

### DIFF
--- a/R/key-helpers.R
+++ b/R/key-helpers.R
@@ -27,6 +27,7 @@ check_key <- function(.data, ...) {
   data_q <- enquo(.data)
   .data <- eval_tidy(data_q)
   args <- exprs(...)
+  names(args) <- set_names(paste0("...", seq_along(args)))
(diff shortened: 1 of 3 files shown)


R/key-helpers.R
```diff
@@ -27,8 +27,9 @@ check_key <- function(.data, ...) {
   data_q <- enquo(.data)
   .data <- eval_tidy(data_q)
   args <- exprs(...)
-
-  if (any(!!args %in% "n")) count_col <- "nn" else count_col <- "n"
+  if (count_problem(args)) {
+    count_col <- "nn"
+  } else count_col <- "n"
```
krlmlr: Can you do

```r
names(args) <- paste0("...", seq_along(args))
count_col <- "n"
```

?

TSchiefer: good idea

PR #85, opened 2019-10-08
add library call to vignette

TSchiefer:
needed, so it works

### DIFF
--- a/vignettes/dm-introduction-relational-data-models.Rmd
+++ b/vignettes/dm-introduction-relational-data-models.Rmd
@@ -28,6 +28,7 @@ knit_print.grViz <- function(x, ...) {
 
 library(tidyverse)
 library(dm)
+library(nycflights13)
 ```
 
 Multiple, linked tables are a common concept within computer science. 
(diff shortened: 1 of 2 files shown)


vignettes/dm-introduction-relational-data-models.Rmd
```diff
@@ -75,9 +76,6 @@ This can result in long and inflated pipe chains full of `left_join()`, `anti_jo
 In classical {dplyr} notation you need three `left_join()` calls to merge the `flights` table gradually to `airlines`, `planes` and `airports` tables to create one wide data frame.
 
 ```{r}
```
krlmlr: It's better to keep these visible and also load nycflights13 here (but also add to the setup chunk so that the startup messages are hidden).

vignettes/dm-introduction-relational-data-models.Rmd
```diff
@@ -132,7 +130,8 @@ Which columns of the `airlines` table can serve as a primary key?
...
 dm %>% 
-  cdm_enum_pk_candidates(airports)
+  cdm_enum_pk_candidates(airports) %>% 
+  arrange(desc(candidate))
```
krlmlr: I'm confused -- why do we arrange here but not below?

TSchiefer: That's because `cdm_enum_fk_candidates()` arranges the tibble and `cdm_enum_pk_candidates()` doesn't. But I agree, we should have consistent behaviour.

I will change this for `cdm_enum_pk_candidates()` in another PR, if that's ok.

krlmlr: Let's do it in this PR.

TSchiefer: `cdm_enum_pk_candidates()` arranges the output in an equivalent way to `cdm_enum_fk_candidates()` now

krlmlr:
Thanks!

PR #86, opened 2019-10-08
function 'cdm_get_con()'

### LINKED ISSUE #84: Implement cdm_get_con()
Like `cdm_get_src()` but returns a database connection (or an error if a data frame or other object).

TSchiefer:
like `cdm_get_src()` but returns a database connection (or an error if a data frame or other object).

fixes #84


R/dm.R
```diff
@@ -147,10 +147,25 @@ validate_dm <- function(x) {
...
+#'
+#' `cdm_get_con()` returns the connection object (`con`-part of \pkg{dplyr} source component) of a `dm`
+#' object.
+#'
+#' @rdname dm
+#'
+#' @export
+cdm_get_con <- function(x) {
+  con <- cdm_get_src(x)[["con"]]
+  if (is_null(con)) abort_no_con_if_local()
```
krlmlr: Can we extract a `src` variable and use `inherits(src, "src_dbi")` instead?

R/test-dm.R
```diff
@@ -24,7 +24,7 @@ cdm_test_load <- function(x,
 # internal helper functions:
 # validates, that object `dm` is of class `dm` and that `table` is character and is part of the `dm` object
 check_correct_input <- function(dm, table) {
-  if (!is_dm(dm)) abort("`dm` has to be of class `dm`")
+  if (!is_dm(dm)) abort_is_not_dm(class(dm))
```
krlmlr: Is it worthwhile to extract a `check_dm()` helper?

R/dm.R
TSchiefer: `class(src)` is `"src_local" "src"`; implemented

R/dm.R
```diff
@@ -147,10 +147,25 @@ validate_dm <- function(x) {
...
+#'
+#' `cdm_get_con()` returns the connection object (`con`-part of \pkg{dplyr} source component) of a `dm`
+#' object.
+#'
+#' @rdname dm
+#'
+#' @export
+cdm_get_con <- function(x) {
+  src <- cdm_get_src(x)
+  if (inherits(src, "src_local")) abort_no_con_if_local()
```
krlmlr: I think this will only work for DBI.

```suggestion
  if (!inherits(src, "src_dbi")) abort_con_only_for_dbi()
```

R/test-dm.R
krlmlr: Yes, let's do it as part of this PR.

PR #87, opened 2019-10-08
PR comments from #78

TSchiefer:
- tweaks
- some outsourcing from `cdm_flatten_to_tbl_impl()`


R/utils.R
```diff
@@ -18,3 +18,30 @@ tick <- function(x) {
...
+    recipe <-
+      compute_disambiguate_cols_recipe(red_dm, tables, sep = ".")
+    explain_col_rename(recipe)
+    # prepare `dm` by disambiguating columns (on a reduced dm)
+    clean_dm <-
+      col_rename(red_dm, recipe)
+    # the column names of start_tbl need to be updated, since taken from `dm` and not `clean_dm`,
+    # therefore we need a named variable containing the new and old names
+    renames <-
+      recipe %>% filter(table == !!start) %>% pull() %>% flatten_chr()
```
krlmlr: Not sure which is better:

```suggestion
      recipe$...[recipe$table == "start"][[1]]
```

TSchiefer: After trying out some things the following seemed to be at least not completely without grace...
```r
pluck(recipe[recipe$table == start, "renames"], 1, 1)
```

krlmlr: `[` is similar to the pipe because you can read left to right. Why not:

```r
recipe$renames[recipe$table == "start"][[1]][[1]]
```

(Do we need the second `[[1]]` ?)

TSchiefer: Ok, this works (actually without the second `[[1]]`, otherwise the names are lost)

TSchiefer: just realized that it doesn't work for empty lists (when no renaming is necessary), since `list()[[1]]` throws an error.

TSchiefer: what about `pluck(recipe$renames[recipe$table == start], 1)`? (works for empty list)

R/utils.R
```diff
@@ -18,3 +18,39 @@ tick <- function(x) {
...
+      pluck(recipe$renames[recipe$table == start], 1)
+    start_tbl <- start_tbl %>% rename(!!!renames)
+  } else {
+    # for `anti_join()` and `semi_join()` no renaming necessary
+    clean_dm <- red_dm
+    renames <- character(0)
+  }
+  tables <- cdm_get_tables(clean_dm)
+  tables[[start]] <- start_tbl
+  new_dm2(
```
krlmlr: This should be `new_dm3()` now.

R/error-helpers.R
```diff
@@ -462,3 +466,43 @@ abort_unique_table_names_or_table_names <- function() {
 error_unique_table_names_or_table_names <- function() {
   "Can supply either `table_names` or `unique_table_names = TRUE`, not both."
 }
+
+
+# error conditions evaluation for 'cdm_flatten_to_tbl_impl()' -------------
+
+check_flatten_to_tbl <- function(
```
krlmlr: This is used only once, and so belongs in `flatten.R`.

krlmlr:
Thanks!

PR #91, opened 2019-10-11
add/remove tibbles to/from dm

### LINKED ISSUE #90: Implement cdm_add_tbl() and cdm_rm_tbl()
For adding, use `def[0, ]` as boilerplate for the new row in the `def`. Use `vctrs::vec_rbind()` for row-binding tibbles.

Removing is implemented for symmetry, the documentation should point to `cdm_select_tbl()` .

Use `@rdname` to document in the same `.Rd` file.

Rewrite all examples and documentation to use this new idiom.

TSchiefer:
three new functions:
1. `cdm_add_tbl()`: adds one tibble; naming can happen optionally in 2nd parameter, otherwise name is kept (doesn't work for pipe and purrr-functions)
1. `cdm_add_tbls()`: adds one or more tibbles at once, using tidyverse naming syntax (`new_table_name = tbl`).
1. `cdm_rm_tbls()`: remove one or more tibbles from `dm` by handing them unquoted to ellipsis

also adapted a few minor things to the new `dm` structure implementation and added tests for the new functions

fixes #90


TSchiefer:
Remark: adding the `cdm_add_tbls()` or `cdm_rm_tbls()` to the help page of `cdm_select_tbl()` is problematic, since the ellipsis has a different meaning in each case. We could of course keep it very vague and add a more specific description in the details for each function

TSchiefer:
Oh, and please feel free to choose one of the two implementations to add tibbles and tweak them according to your likes and needs. I guess they could coexist, but we do not necessarily need both of them.

R/table_management.R
```diff
@@ -0,0 +1,109 @@
...
+
+  cdm_select_tbl(dm, -one_of(!!!table_names))
+
+}
+
+
+check_new_tbls <- function(dm, tbls, name) {
+  orig_tbls <- cdm_get_tables(dm)
+  # are all new tables on the same source as the original ones?
+  if (!all_same_source(flatten(list(cdm_get_tables(dm), tbls)))) abort_not_same_src()
```
krlmlr:
```suggestion
  if (has_length(cdm_get_tables(dm)) && !all_same_source(c(cdm_get_tables(dm)[1], tbls))) abort_not_same_src()
```

R/table_management.R
```diff
@@ -0,0 +1,109 @@
...
+  # test if a name "." is part of the new names, indicating a piped table that is not explicitly named
+  # or if table names of the kind ".x" or "..1" are present, which would indicate a `map()`-type operation
+  if ("." %in% name || any(str_detect(name, "^\\.[x-z]$")) || any(str_detect(name, "^\\.\\.[0-9][0-9]?$"))) {
```
krlmlr: We can offer some sort of name repair to deal with that problem.

R/draw-dm.R
```diff
@@ -116,7 +116,7 @@ color_quos_to_display <- function(...) {
 #' @export
 cdm_get_colors <- nse_function(c(dm), ~ {
   cdm_get_def(dm) %>%
-    select(table = name, display) %>%
+    select(table, display) %>%
```
krlmlr: Do we have a test here?

krlmlr:
Thanks!

PR #96, opened 2019-10-21
provide empty 'dm' object

TSchiefer:
- Empty object can be created via `dm()` and `new_dm()` (without arguments, similar to `numeric()`, `list()`, etc.)
- Allows for creation of a `dm` from scratch, adding tables successively with `cdm_add_tbl()`
- Works locally and on DB
- vignettes have NOT been adapted

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -600,3 +600,18 @@ all_same_source <- function(tables) {
   first_table <- tables[1][[1]]
   is.null(detect(tables[-1], ~ !same_src(., first_table)))
 }
+
+# creates an empty `dm`-object, `src` is defined by implementation of `cdm_get_src()`.
+empty_dm <- function() {
+  new_dm3(
+    tibble(
+      table = character(),
+      data = list(),
+      segment = logical(),
+      display = character(),
+      pks =vctrs::list_of(new_pk()),
+      fks = vctrs::list_of(new_fk()),
+      filters = vctrs::list_of(new_filter())
+    )
+  )
+}
--- a/R/dm.R
+++ b/R/dm.R
@@ -62,6 +62,7 @@ dm <- nse_function(c(src, data_model = NULL), ~ {
 #' @rdname dm
 #' @export
 new_dm <- function(tables, data_model) {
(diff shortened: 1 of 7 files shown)


R/select-tbl.R
```diff
@@ -49,6 +49,8 @@ tidyselect_table_names <- function(dm) {
 }
 
 cdm_select_tbl_impl <- function(dm, selected) {
+
+  if (is_empty(selected)) return(empty_dm())
```
krlmlr: Do we really need the shortcut here?

TSchiefer: We get an error otherwise:
```r
Error in dot_call(capture_dots, frame_env = frame_env, named = named,  : 
  attempt to set an attribute on NULL
```

krlmlr: Where does that error occur? Also with `selected = character()`?

krlmlr:
```suggestion
  # Required to avoid error further below
  if (is_empty(selected)) return(empty_dm())
```

TSchiefer: in `filter_recode_table(selected)` when `selected` is `named character(0)`

R/dm.R
```diff
@@ -4,7 +4,7 @@
 #' The `dm` class wraps [dplyr::src] and adds a description of table relationships
 #' based on [datamodelr::datamodelr-package].
 #'
-#' `dm()` coerces its inputs.
+#' `dm()` coerces its inputs. If called without parameters, an empty `dm` object is created.
```
```suggestion
#' `dm()` coerces its inputs. If called without arguments, an empty `dm` object is created.
```

R/dm.R
```diff
@@ -56,11 +57,14 @@ dm <- nse_function(c(src, data_model = NULL), ~ {
 #' Low-level constructor
 #'
 #' `new_dm()` only checks if the inputs are of the correct class.
+#' If called without parameters, an empty `dm` object is created.
```
```suggestion
#' If called without arguments, an empty `dm` object is created.
```

krlmlr:
Thanks! Need to add a test that `cdm_select_tbl(dm)` is empty.

PR #98, opened 2019-10-21
Define filtering operation

### LINKED ISSUE #70: Define filtering operation
- referential integrity of the result guaranteed
- filter then flatten is the same as flatten with inner join then filter (?)

In the documentation in `?cdm_filter` and `vignette("dm-filtering")` .

TSchiefer:
- update doc of `cdm_filter()`
- update vignette "Filtering in Relational Data Models"

fixes #70

### DIFF
--- a/R/filter-dm.R
+++ b/R/filter-dm.R
@@ -1,6 +1,6 @@
 #' Filtering a [`dm`] object
 #'
-#' Filtering one table of a [`dm`] object has an effect on all tables connected to this table
+#' Filtering one table of a [`dm`] object affects all tables connected to this table
 #' via one or more steps of foreign key relations. Firstly, one or more filter conditions for
 #' one or more tables can be defined using `cdm_filter()`, with a syntax similar to `dplyr::filter()`.
 #' These conditions will be stored in the [`dm`] and not immediately executed. With `cdm_apply_filters()`
@@ -9,10 +9,19 @@
 #'
 #' @details `cdm_filter()` allows you to set one or more filter conditions for one table
 #' of a [`dm`] object. These conditions will be stored in the [`dm`] for when they are needed.
-#' Once executed, the filtering the will affect all tables connected to the filtered one by
-#' foreign key constraints, leaving only the rows with the corresponding key values. The filtering
-#' implicitly takes place, once a table is requested from the [`dm`] by using one of `tbl()`, `[[.dm()`, `$.dm()`.
+#' The conditions are only evaluated in one of the following scenarios:
+#' 1. Calling `cdm_apply_filters()` or `compute()` (method for `dm` objects) on a `dm`: each filtered table potentially
+#' reduces the rows of all other tables connected to it by foreign key relations (cascading effect), only leaving the rows
+#' with the corresponding key values. Isolated (in terms of foreign key constraints) and unfiltered tables are left untouched.
+#' This results in a new `dm` class object.
+#' 1. Calling one of `tbl()`, `[[.dm()`, `$.dm()`: the remaining rows of the requested table are calculated based on the
+#' filter conditions and the foreign key conditions (similar to 1. but only for one table)
 #'
+#' It should be mentioned, that if an UNFILTERED table had a potential effect on other tables -- since it lacks key
+#' values from foreign tables --, these other tables are in general NOT affected by them. If you want to explore this effect
+#' you can set a filter condition like `1 == 1` to enforce this effect.
+#'
+#' Several functions of the {dm} package will throw an error if unevaluated filter conditions exist when they are called.
 #' @rdname cdm_filter
 #'
 #' @inheritParams cdm_add_pk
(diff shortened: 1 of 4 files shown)


vignettes/dm-filtering.Rmd
```diff
@@ -94,47 +94,65 @@ tbl(dm, "airports")
...
-  cdm_filter(airports, name == "John F Kennedy Intl") %>%
+rows_per_table <- 
+  filtered_dm %>% 
+  cdm_apply_filters() %>% 
   cdm_nrow()
+rows_per_table
+sum(rows_per_table)
 ```
 
+The total number of rows drops from  367 687 to 138 786 (the only unaffected table is the disconnected `weather` table).
```
krlmlr: Is it too risky to compute the number automatically?

```suggestion
The total number of rows drops from `r nrow(dm$flights)` to `r cdm_nrow[["flights"]]` (the only unaffected table is the disconnected `weather` table).
```

R/filter-helpers.R
```diff
@@ -5,6 +5,7 @@
 #' @param dm A [`dm`] object
 #' @export
 cdm_nrow <- function(dm) {
+  check_no_filter(dm)
```
krlmlr: Why is this important?

R/filter-dm.R
```diff
@@ -9,10 +9,19 @@
...
-#' implicitly takes place, once a table is requested from the [`dm`] by using one of `tbl()`, `[[.dm()`, `$.dm()`.
+#' The conditions are only evaluated in one of the following scenarios:
+#' 1. Calling `cdm_apply_filters()` or `compute()` (method for `dm` objects) on a `dm`: each filtered table potentially
+#' reduces the rows of all other tables connected to it by foreign key relations (cascading effect), only leaving the rows
+#' with the corresponding key values. Isolated (in terms of foreign key constraints) and unfiltered tables are left untouched.
+#' This results in a new `dm` class object.
+#' 1. Calling one of `tbl()`, `[[.dm()`, `$.dm()`: the remaining rows of the requested table are calculated based on the
+#' filter conditions and the foreign key conditions (similar to 1. but only for one table)
 #'
+#' It should be mentioned, that if an UNFILTERED table had a potential effect on other tables -- since it lacks key
```
krlmlr: This is confusing and a bit too detailed, maybe add an example that highlights this effect?

R/filter-dm.R
```diff
@@ -9,10 +9,19 @@
 #'
 #' @details `cdm_filter()` allows you to set one or more filter conditions for one table
 #' of a [`dm`] object. These conditions will be stored in the [`dm`] for when they are needed.
-#' Once executed, the filtering the will affect all tables connected to the filtered one by
-#' foreign key constraints, leaving only the rows with the corresponding key values. The filtering
-#' implicitly takes place, once a table is requested from the [`dm`] by using one of `tbl()`, `[[.dm()`, `$.dm()`.
+#' The conditions are only evaluated in one of the following scenarios:
+#' 1. Calling `cdm_apply_filters()` or `compute()` (method for `dm` objects) on a `dm`: each filtered table potentially
+#' reduces the rows of all other tables connected to it by foreign key relations (cascading effect), only leaving the rows
+#' with the corresponding key values. Isolated (in terms of foreign key constraints) and unfiltered tables are left untouched.
```
krlmlr:
```suggestion
#' with the corresponding key values.
#' Tables that are not connected to any table with an active filter are left unchanged.
```

R/filter-helpers.R
TSchiefer: I would think, if a user adds filter conditions and then calls `cdm_nrow()`, they are interested in the number rows after filtering. But since the function just counts the rows in the unfiltered state (using `cdm_get_tables()`) I figured it would make sense to add this function to those that don't work when a filter condition exists.

R/filter-dm.R
TSchiefer: I tried explaining it a bit better now and added an example (1f9c06975719cadde3b2a1badd44be03d9b38ff8).

We could also leave the explanatory text away and let the example speak for itself(?)

krlmlr:
Sounds fair.

TSchiefer:
Ok, removing the text

TSchiefer:
wait a second: Does "sounds fair" refer to 

- [ ] the removal of the explanatory text or
- [ ] adding `cdm_nrow()` to the functions that don't accept filtered `dm`

R/filter-dm.R
krlmlr: I'm in favor of removing the text and only explain by example.

R/filter-helpers.R
krlmlr: Yeah, sure, we need `cdm_apply_filters()` to compute accurately.

krlmlr:
"Sounds fair" was misplaced by the e-mail reply system -- please follow the most recent comments.

R/filter-helpers.R
TSchiefer: Sorry, got to ask once more to make sure: should we have in `cdm_nrow()` rather `check_no_filter(dm)` or `cdm_apply_filters(dm)`?

krlmlr:
Checking is good, because applying is very costly.

PR #99, opened 2019-10-21
function for zooming to a table of a 'dm'

TSchiefer:
- function `cdm_zoom_to_tbl()` (as of yet unexported)
- print method for `dm` that has a zoomed on table
- function `cdm_zoom_out()` (as of yet unexported)

exporting doesn't make sense just yet I would say


R/dm.R
```diff
@@ -98,7 +98,8 @@ new_dm <- function(tables, data_model) {
     as.character(data_model_tables$display),
     keys,
     references,
-    filter = new_filters()
+    filter = new_filters(),
+    zoom = new_zoom()
```
krlmlr: Can you please get rid of `new_dm2()` as part of this pull request? I remember it's called only from here.

R/dm.R
```diff
@@ -481,6 +496,18 @@ print.dm <- function(x, ...) {
   invisible(x)
 }
 
+print_zoomed <- function(x) {
+  # so far only 1 table can be zoomed on
+  zoom <- cdm_get_zoomed_tbl(x)
+  tbl <- zoom %>% pull(zoom) %>% pluck(1)
+  name <- zoom %>% pull(table)
+  # FIXME: how do I get the text to be in grey colour (like above `tibble` print output)?
```
krlmlr: Please implement `tbl_sum.dm()`, see `dplyr:::tbl_sum.grouped_df` .

R/dm.R
```diff
@@ -481,6 +496,18 @@ print.dm <- function(x, ...) {
   invisible(x)
 }
 
+print_zoomed <- function(x) {
```
krlmlr: Ideally we'd implement `format()` and forward `print()` to it.

R/zoom.R
```diff
@@ -0,0 +1,22 @@
+cdm_zoom_to_tbl <- function(dm, table) {
+  # for now only one table can be zoomed on
+  zoom <- as_string(ensym(table))
+  check_correct_input(dm, zoom)
+  zoom_tbl <- tbl(dm, zoom)
```
krlmlr: Do we accept filtered dm-s? What happens if the dm is already zoomed?

R/zoom.R
```diff
@@ -0,0 +1,22 @@
...
+
+  new_dm2(
```
krlmlr: Please use `new_dm3()` .

R/zoom.R
```diff
@@ -0,0 +1,22 @@
...
+    zoom = tibble(table = zoom, zoom = list(zoom_tbl)),
+    base_dm = dm
+  )
+}
+
+is_zoomed <- function(dm) {
+  !all(cdm_get_zoomed_tbl(dm) %>% pull(zoom) %>% map_lgl(is_null))
```
krlmlr: Can you implement using `cdm_get_zoomed_tbl()` ?

R/zoom.R
```diff
@@ -0,0 +1,22 @@
...
+}
+
+cdm_zoom_out <- function(dm) {
+  new_dm2(
```
krlmlr: Please use `new_dm3()`.

R/zoom.R
TSchiefer: 1. filtered dm-s: this is a tricky question, since it very much depends on what the user is planning to do with the zoomed table. I do think there are plenty scenarios, where things could take unintended turns, if we accept filtered dm-s and the user isn't aware that the original (unfiltered) table is used for the zoom. I think I would go for a warning. What do you think?
2. Currently, the originally zoomed table will be just dropped and forgotten, if another table is zoomed on. This would be a more than annoying behaviour, if that happened with a carefully manipulated zoomed table. So maybe let's only allow to zoom on a table if the `dm` is unzoomed?

krlmlr:
Can we make both an error for now?

TSchiefer:
Sure, sounds good

R/zoom.R
TSchiefer: I would think I kind of did. The name for the function `cdm_get_zoomed_tbl()` is maybe a bit misleading: it does not return only the tibble itself, but a tibble with two columns and one row: the original table name and a list with the zoomed table.

I see your point and I think we need two functions:
1. the function as it is currently implemented (it's very similar to `cdm_get_filter()`) and it should be exported eventually
2. an internal function (e.g. `get_zoomed_tbl()`), that just returns the zoomed table if one exists, and `NULL` otherwise. <- this one I can then use for `is_zoomed()`

krlmlr: Sorry, missed that. How about `nrow(...) > 0` ?

R/dm.R
krlmlr: Please inline so that `new_dm()` calls `new_dm3()` directly.

R/zoom.R
TSchiefer: I was now thinking of introducing a subclass `zoomed_dm`, similar to `grouped_df`. Would that be an option too? Then the printing could work even more similarly to the printing of tibbles. Would that be ok with you?

krlmlr: A subclass seems "cleaner", but let's implement without a subclass first and later think where it would help us.

TSchiefer: I think one important argument for introducing the subclass is, that the {dplyr} verbs would work with the subclass "zoomed_dm", but not with "dm".

R/dm.R
TSchiefer: I implemented `tbl_sum.zoomed_df()`, see comment above the function for explanation.

R/dm.R
TSchiefer: Not sure how the `format()`-method(s) should look like. I will postpone it, unless you describe how it should look, if that's ok.

R/zoom.R
TSchiefer: FYI: It seemed just so much more straightforward to implement it using the subclass that I went for it. Also, as mentioned in my last comment in this thread, I really do believe it makes a lot of sense to have this subclass.

krlmlr:
We still need a nice error message when trying with an unzoomed dm.

TSchiefer:
Error message for trying what with an unzoomed dm? when zooming out? (I'd say we could let that one slip if a user tries it with an unzoomed one)

krlmlr:
Error when trying to mutate something in an unzoomed dm.

TSchiefer:
Right, I was thinking that we could do this in a later step. But what about something along the lines of: "Please specify the table first that you want to manipulate, using `cdm_zoom_to_tbl()`."

R/error-helpers.R
```diff
@@ -483,3 +483,13 @@ abort_con_only_for_dbi <- function() {
...
 }
+
+# no zoom allowed for `cdm_zoom_to_tbl()` ---------------------------------
+
+abort_no_zoom_allowed <- function() {
+  abort(error_no_zoom_allowed(), .subclass = cdm_error_full("no_zoom_allowed"))
+}
+
+error_no_zoom_allowed <- function() {
+  "`cdm_zoom_to_tbl()` only works for unzoomed `dm`"
```
krlmlr: What can the user do here?

R/dm.R
```diff
@@ -485,6 +480,29 @@ print.dm <- function(x, ...) {
   invisible(x)
 }
 
+print.zoomed_dm <- function(x, ..., n = NULL, width = NULL, n_extra = NULL) {
+
+  df <- get_zoomed_tbl(x)
```
krlmlr: This part up to and including the `format()` call inside `cat_line()` can be part of a new `format.zoomed_dm()` .

PR #102, opened 2019-10-24
support {tidyselect} for flatten/squash

### LINKED ISSUE #101: Support tidyselect in cdm_squash_to_tbl() and cdm_flatten_to_tbl()
``` r
library(dm)

dm <- cdm_nycflights13()

dm %>% cdm_flatten_to_tbl(flights, -planes)
#> Tables `-planes` not in `dm` object. Available table names: `airlines`,
#> `airports`, `flights`, `planes`, `weather`
```

<sup>Created on 2019-10-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

TSchiefer:
- support select helpers, deselecting
- tests

fixes #101

### DIFF
--- a/R/flatten.R
+++ b/R/flatten.R
@@ -16,7 +16,7 @@
 #' the order of the joins. If empty, all tables that can be reached are included.
 #' If this includes tables which aren't direct neighbours of `start`,
 #' it will only work with `cdm_squash_to_tbl()` (given one of the allowed join-methods).
-#' `tidyselect` is supported.
+#' `tidyselect` is supported, cf. [dplyr::select()].
 #' @family flattening functions
 #'
 #' @details With the `...` left empty, this function joins all the tables of your [`dm`]
--- a/R/flatten.R
+++ b/R/flatten.R
@@ -16,6 +16,7 @@
 #' the order of the joins. If empty, all tables that can be reached are included.
 #' If this includes tables which aren't direct neighbours of `start`,
 #' it will only work with `cdm_squash_to_tbl()` (given one of the allowed join-methods).
+#' `tidyselect` is supported.
 #' @family flattening functions
 #'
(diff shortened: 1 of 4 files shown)


R/flatten.R
```diff
@@ -76,8 +77,9 @@ cdm_squash_to_tbl <- function(dm, start, ..., join = left_join) {
 
 cdm_flatten_to_tbl_impl <- function(dm, start, ..., join, join_name, squash) {
   check_correct_input(dm, start)
-  list_of_pts <- as.character(enexprs(...))
-  check_correct_input(dm, list_of_pts)
+  vars <- setdiff(tidyselect_table_names(dm), start)
+  list_of_pts <- tidyselect::vars_select(vars, ...)
```
krlmlr: What's the wording of the error message when the table isn't found?

R/flatten.R
```diff
@@ -16,6 +16,7 @@
 #' the order of the joins. If empty, all tables that can be reached are included.
 #' If this includes tables which aren't direct neighbours of `start`,
 #' it will only work with `cdm_squash_to_tbl()` (given one of the allowed join-methods).
+#' `tidyselect` is supported.
```
krlmlr: Can you please add a link to a relevant help page?

R/flatten.R
TSchiefer: `Error in is_character(x) : object 'dim_5' not found`...

better might be a `tryCatch()` with `abort_table_not_in_dm(setdiff(tables, src_tbls(dm)), dm))` in case of an error, but it seems non-trivial to find the wrong columns in case the user uses deselection or select helpers

TSchiefer: An alternative would be to extend the {tidyselect} error message (also in a `tryCatch()`), adding which tables are available.

TSchiefer: Implemented in 4fc8eb8

PR #106, opened 2019-10-25
reorder tables in 'dm' using 'cdm_select_tbl()' 

TSchiefer:
- order in selection -> order of tables in `dm`

### DIFF
--- a/R/select-tbl.R
+++ b/R/select-tbl.R
@@ -54,10 +54,13 @@ cdm_select_tbl_impl <- function(dm, selected) {
   if (is_empty(selected)) return(empty_dm())
   check_correct_input(dm, selected)
 
+  order_tbl <- tibble(table = names(selected))
+
   def <-
     cdm_get_def(dm) %>%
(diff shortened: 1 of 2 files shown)


R/select-tbl.R
```diff
@@ -72,8 +72,8 @@ filter_recode_table_fks <- function(def, selected) {
 }
 
 filter_recode_table <- function(data, selected) {
-  data %>%
-    filter(table %in% !!selected) %>%
+  reduced_selected <- tibble(table = selected[selected %in% data$table])
```
krlmlr: Would that work:

```suggestion
  reduced_selected <- tibble(table = unname(selected))
```

We're checking inputs earlier on.

TSchiefer: produces an error in the tests. I think the problem is that `filter_recode_table()` is used in a few different places, among others in `cdm_rm_pk()`. I wouldn't want to dig too deep, cause it seems like a waste of time. If you want I can try to figure it out though.

PR #107, opened 2019-10-25
update/insert zoomed table

TSchiefer:
- function `cdm_update_zoomed_tbl()` for replacing table with zoomed table
- function `cdm_insert_zoomed_tbl()` for adding zoomed table to `dm`
- basic tests

I was wondering if we should also make a copy of the keys of the zoomed table available in column `zoom`, maybe as a 3-columned tibble in the `zoom` column of `def` with the columns:
1. `data`
2. `pks`
3. `fks`

When manipulations are being done with `data`, the keys can then (hopefully always) be tracked and accordingly inserted/updated in the end.

On a related topic: I realized that the column `fks` in `def` contains the keys pointing to the respective table. I was wondering if if might not be beneficial to store here the keys pointing away from the table. Mainly because it would make the tracking of the key relations slightly easier (otherwise, when zooming to a table, we'd need to first search which keys are pointing away from it in order to store it in the `zoom` column)


krlmlr:
Can you please expand by implementing `cdm_select_tbl()` and `cdm_rename_tbl()` as zoom + select + update? I think we need another column, perhaps named `selected`, in the def.

TSchiefer:
I implemented `select.zoomed_dm()` in b7ac68a and with `cdm_update_zoomed_tbl()` the original tibble is replaced by the manipulated one, so your suggestion would work.

So far `cdm_update_zoomed_tbl()` does not touch keys yet, but it needs to do so in the future. I should probably for now let the key entries for the updated table be the result of two functions (e.g. `update_zoomed_pk()`, `update_zoomed_fks()`), which will produce empty entries for now and can be adapted later on.

PR #108, opened 2019-10-25
methods for {dplyr}-verbs

### LINKED ISSUE #89: activate()-style API
With #57 and part of #54 we're now ready to start looking into an API that extracts a table from a dm but keeps the context, so that it can be put back.

```r
dm %>%
  activate(flights) %>%
  filter(hour < 6) %>%
  ...
```

## How do we put back a table?

- Overwrite an existing table, removing all obsolete keys
- Create a new table and add primary and foreign keys where we know that they haven't been altered

## How are the verbs called?

We temporarily remove a piece of a mosaic and reinsert it or perhaps insert a replica. Is reinsertion different from adding a new version? Maybe `zoom_to_tbl()`, `update_tbl()`, `insert_tbl()` ?

## Roadmap

- [x] `zoom_to_tbl()` and `update_tbl()`, with `print()` and `format()` method
- [ ] `select.dm()` and `rename.dm()`, with a nice error message if not activated
- [ ] `filter.dm()`
- [ ] `mutate.dm()` and `transmute.dm()`
- [ ] `summarise.dm()`
- [ ] `group_by.dm()`, `ungroup.dm()`

Independently:

- [x] `cdm_add_tbl()` (useful for #77)
- [ ] `insert_tbl()`

### LINKED ISSUE #112: internal: 'def' (and other) changes for 'activate()'-API
changes related to `def`:

1. filters: 2nd column, boolean `zoomed`
1. keep all keys that could be tracked (when inserting new table; this does not mean, that, when deselecting key-columns, they are automatically re-added to the tibble)

other changes:

1. implement `compute()` also for `zoomed_dm`
1. error for `select.dm()` & `rename.dm()` when filters exist for the original and/or zoomed table

TSchiefer:
methods for `dm` and `zoomed_dm`for:
- `group_by()`
- `ungroup()`
- `summarise()`
- `filter()`
- `mutate()`
- `transmute()`
- `select()`
- `rename()`

The key handling is not yet in place. We should probably sort this out before merging?

Would be nice, if I could get a preliminary feedback though.

closes #112 
fixes #89


krlmlr:
Looks good. Initial remarks:

- ellipsis should sufficient for all methods
- filter() only sets the filter
- new column in the def?

TSchiefer:
As far as I understand, we first need to discuss how we want to store/track the key columns of the zoomed table, before I can implement `cdm_select_tbl()` in the way you suggested.

I agree with you, that we need a new column in `def`. However, I think we need information about three things in that new column: 
1. the current pk-column of the zoomed table
2. the current columns pointing away from the zoomed table as FKs
3. the filter conditions for the zoomed table

(I actually would even prefer to add three new columns to `def`)

Also, I still think that it would be better to implement `def$fks` in such a way, that it contains the FKs pointing away from a table, and not to the table: then we could just copy the FKs from the base table to the zoomed table.

krlmlr:
Thanks!

PR #114, opened 2019-11-01
Integrate {datamodelr}-code in {dm}

TSchiefer:
- necessary code integrated in file "datamodelr-code.R"
- adding Darko Bergant as co-author (email-address taken from his github-page)
- minimal changes to existing code

fixes #111

PR #115, opened 2019-11-01
rename prefix "cdm" -> "dm"

TSchiefer:
should wait for #114


krlmlr:
Thanks, let's postpone, #117 contains a script that creates results consistent with this PR and can be applied at any time.

PR #116, opened 2019-11-01
function to disentangle multiple relations between 2 tables

### LINKED ISSUE #55: New operation: disentangle
dm in, dm, out.

Implementation:

- dfs or bfs

TSchiefer:
- only works for case of different columns in child tables (CT) pointing to same parent tables (PT) (case of one column of CT pointing to different PTs would be also conceivable(?))
- sets colors based on old PT
- creates copies of old PT with names: `paste0(child_fk_col, ".", PT)`
- "cdm"-prefix in "disentangle.R" and "test-disentangle.R" needs to be fixed after #115 

fixes #55


TSchiefer:
I forgot earlier: of course we can drop the `child_table` argument and let the function find the tables itself, where disentangling is necessary. But I was also thinking that the user knows which tables are candidates for this. So I don't have a clear opinion about this. What do you think?

krlmlr:
This turns out to be more complicated than anticipated. Let's shelve for now.

TSchiefer:
Maybe let us revisit this later on

PR #119, opened 2019-11-04
avoid 'cdm_get_data_model()' in vignettes

TSchiefer:
This is just a hotfix, so R_CMD_check doesn't throw an error. We need to rework the vignettes as soon as we're at a stable state


TSchiefer:
Is it ok to merge this, so the checks stop failing? We will need to rework the vignettes anyway

krlmlr:
Thanks!

PR #122, opened 2019-11-04
Bugfix: `cdm_select_tbl()` with multi-FKs between 2 tables

TSchiefer:
currently on master:
``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter
cdm_nycflights13(cycle = TRUE) %>% cdm_draw()
```

``` r
cdm_select_tbl(cdm_nycflights13(cycle = TRUE), flights, airports) %>% cdm_draw()
```

<sup>Created on 2019-11-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

in this branch:
``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter
cdm_nycflights13(cycle = TRUE) %>% cdm_draw()
```

``` r
cdm_select_tbl(cdm_nycflights13(cycle = TRUE), flights, airports) %>% cdm_draw()
```

<sup>Created on 2019-11-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/select-tbl.R
+++ b/R/select-tbl.R
@@ -72,16 +72,19 @@ filter_recode_table_fks <- function(def, selected) {
 }
 
 filter_recode_table_def <- function(data, selected) {
+  # We want to keep the order mentioned in `selected` here.
+  # data$table only contains unique values by definition.
   idx <- match(selected, data$table, nomatch = 0L)
+
   data[idx, ] %>%
     mutate(table = recode(table, !!!prep_recode(selected)))
 }
 
 filter_recode_fks_of_table <- function(data, selected) {
+  # data$table can have multiple entries, we don't care about the order
   idx <- data$table %in% selected
   data[idx, ] %>%
     mutate(table = recode(table, !!!prep_recode(selected)))
-
 }
 
 prep_recode <- function(x) {
--- a/R/select-tbl.R
+++ b/R/select-tbl.R
@@ -72,7 +72,7 @@ filter_recode_table_fks <- function(def, selected) {
 }
 
 filter_recode_table_def <- function(data, selected) {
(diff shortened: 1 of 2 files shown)


R/select-tbl.R
```diff
@@ -72,7 +72,7 @@ filter_recode_table_fks <- function(def, selected) {
 }
 
 filter_recode_table <- function(data, selected) {
-  idx <- match(selected, data$table, nomatch = 0L)
+  idx <- map(unname(selected), ~which(. == data$table)) %>% flatten_int()
```
krlmlr: The `match()` was a daft idea.

```suggestion
  idx <- data$table %in% selected
```

TSchiefer:
Argh, no, sorry, it doesn't work with `%in%` since we need the indices for ensuring the order...

TSchiefer:
I can't think of a nicer solution than the one you already saw, that works in all cases. Please feel free to suggest something though.

R/select-tbl.R
```diff
@@ -72,7 +72,7 @@ filter_recode_table_fks <- function(def, selected) {
 }
 
 filter_recode_table <- function(data, selected) {
-  idx <- match(selected, data$table, nomatch = 0L)
+  idx <- map(unname(selected), ~which(data$table == .)) %>% flatten_int()
```
krlmlr: - create an `ordered()` from `data$table` with `levels = selected`
- remove all `NA`
- sort

krlmlr: Alternatively, implement two functions, one for `def`, one for the fks?

krlmlr:
Thanks!

PR #125, opened 2019-11-06
instant filter for filter-table

### LINKED ISSUE #113: Use hidden column to implement filters
Advantages: All dplyr operations are allowed on filtered tables.

Disadvantage: Hidden technical column.

A filter translates to

```r
mutate(.dm_filter = if_else(..., 1, 0))
```

or

```r
mutate(.dm_filter = if_else(.dm_filter == 1, if_else(..., 1, 0), 0))
```

Every time we show such a table to the user, we use

```r
. %>% filter(.dm_filter != 0) %>% select(-.dm_filter)
```

### LINKED ISSUE #124: Apply filters right away on the data
We memoize filters in a separate structure, because we don't want them to have an effect on all related tables right away -- this leads to SQL that's far too complex.

We might be able to apply the filters in the individual tables, though. We still record the filter expression for `cdm_apply_filters()` and for display, but each `cdm_filter(...)` or `. %>% zoom_to_tbl(...) %>% filter()` applies the filter on the table.

Would that work?

Ideally, `cdm_filter()` would be the same as zoom + `filter()` .

TSchiefer:
when setting a filter (either with `cdm_filter()` or `cdm_zoom_to_tbl()` - `filter()`) the table that is filtered will be actually filtered immediately. In addition, the filter-expression is stored.

1. users can still access their filter conditions (also available in `print()`-method)
2. no problems anymore with changing column names etc.
3. effect on all other tables needs to be invoked via `cdm_apply_filters()` or `compute.dm()`

fixes #124 
fixes #113 ?

### DIFF
--- a/R/dplyr.R
+++ b/R/dplyr.R
@@ -26,7 +26,6 @@ ungroup.zoomed_dm <- function(x, ...) {
 
 #' @export
 summarise.zoomed_dm <- function(.data, ...) {
-  if (nrow(cdm_get_filter(.data) %>% filter(table == !!orig_name_zoomed(.data)))) abort_no_filters_rename_select()
   tbl <- get_zoomed_tbl(.data)
   # groups are "selected"; key tracking will continue for them
   groups <- set_names(map_chr(groups(tbl), as_string))
@@ -68,7 +67,6 @@ mutate.dm <- function(.data, ...) {
 
 #' @export
 mutate.zoomed_dm <- function(.data, ...) {
-  if (nrow(cdm_get_filter(.data) %>% filter(table == !!orig_name_zoomed(.data)))) abort_no_filters_rename_select()
   tbl <- get_zoomed_tbl(.data)
   mutated_tbl <- mutate(tbl, ...)
   # all columns that are not touched count as "selected"; names of "selected" are identical to "selected"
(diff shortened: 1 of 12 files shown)


R/dplyr.R
```diff
@@ -47,12 +46,18 @@ filter.dm <- function(.data, ...) {
 
 #' @export
 filter.zoomed_dm <- function(.data, ...) {
-  quos <- enquos(...)
-  if (is_empty(quos)) {
+  filter_exprs <- enexprs(...)
```
krlmlr: We still need `enquo()` here?

R/filter-dm.R
```diff
@@ -61,19 +61,16 @@ cdm_filter <- function(dm, table, ...) {
...
+    cdm_update_zoomed_tbl()
 }
 
-set_filter_for_table <- function(dm, table, quos, zoomed) {
+set_filter_for_table <- function(dm, table, filter_exprs, zoomed) {
   def <- cdm_get_def(dm)
 
   i <- which(def$table == table)
-  def$filters[[i]] <- vctrs::vec_rbind(def$filters[[i]], new_filter(quos, zoomed))
+  def$filters[[i]] <- vctrs::vec_rbind(def$filters[[i]], new_filter(filter_exprs, zoomed))
```
krlmlr: `map(..., quo_get_expr)` ?

R/dplyr.R
krlmlr: If only to do the work of collecting the expressions only once.

R/dplyr.R
```diff
@@ -47,12 +46,18 @@ filter.dm <- function(.data, ...) {
...
+  if (is_empty(filter_exprs)) {
     return(.data)
   } # valid table and empty ellipsis provided
 
-  set_filter_for_table(.data, orig_name_zoomed(.data), quos, TRUE)
+  tbl <- get_zoomed_tbl(.data)
+  filtered_tbl <- filter(tbl, ...)
```
krlmlr: In general, once we `enquo()` or `enexpr()`, we should use this and not pass on the ellipsis.

```suggestion
  filtered_tbl <- filter(tbl, !!!filter_quos)
```

krlmlr:
Thanks!

PR #129, opened 2019-11-06
copy tables to a 'dm'

TSchiefer:
implementation for `copy_to.dm()`, based on `cdm_add_tbl_impl()`

- adds table to `dm` (also if the `srcs` don't match)


R/dm.R
```diff
@@ -583,9 +583,35 @@ src_tbls.dm <- function(x) {
 }
 
 #' @export
-copy_to.dm <- function(dest, df, name = deparse(substitute(df)), overwrite = FALSE, ...) {
-  # TODO: How to add a table to a dm?
-  abort("`dm` objects are immutable, please use ...")
+copy_to.dm <- function(dest, df, name = deparse(substitute(df)), overwrite = FALSE, temporary = TRUE, repair = "unique", quiet = FALSE, ...) {
+  df_list <- if (
+    inherits(df, "list") &&
```
krlmlr: We don't need to support lists of data frames here.

R/dm.R
```diff
@@ -583,9 +583,35 @@ src_tbls.dm <- function(x) {
...
+  # if `df_list` is not named and has different length than `name` throw an error
+  if (is_named(df_list)) {
+    if (length(df_list) != length(name)) name <- names(df_list)
+  } else {
+    if (length(df_list) != length(name)) abort_one_name_for_each_table()
+  }
+  # src: if `df` on a different src:
+  # if `df_list` is on DB and `dest` is local, collect `df_list`
+  # if `df_list` is local and `dest` is on DB, copy `df_list` to respective DB
+  if (!all_same_source(append(cdm_get_tables(dest), df_list))) {
```
krlmlr: Can we call `copy_to()` unconditionally and let the implementation decide if a copy needs to be made?

TSchiefer: The idea was, that if we have a local `dm` and want to add a DB table to it, then it should collect it from the DB and add it.

Do you think we should not mix the two? (copy_to and collect)

krlmlr: `copy_to()` works for DB -> DF too:

``` r
print(
  dplyr::copy_to(
    dplyr::src_df(env = rlang::new_environment()),
    dbplyr::memdb_frame(a = 1)
  )
)
#> # A tibble: 1 x 1
#>       a
#>   <dbl>
#> 1     1
```

<sup>Created on 2019-11-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

TSchiefer: that's nice, didn't know, thanks

R/dm.R
```diff
@@ -583,9 +583,20 @@ src_tbls.dm <- function(x) {
...
-  # TODO: How to add a table to a dm?
-  abort("`dm` objects are immutable, please use ...")
+copy_to.dm <- function(dest, df, name = deparse(substitute(df)), overwrite = FALSE, temporary = TRUE, repair = "unique", quiet = FALSE, ...) {
+  if (!(inherits(df, "data.frame") || inherits(df, "tbl_dbi"))) abort_only_data_frames_supported()
+  if (overwrite) abort_no_overwrite()
+  if (length(name) != 1) abort_one_name_for_copy_to(name)
+  # src: if `df` on a different src:
+  # if `df_list` is on DB and `dest` is local, collect `df_list`
+  # if `df_list` is local and `dest` is on DB, copy `df_list` to respective DB
+  df <- copy_to(cdm_get_src(dest), df, unique_db_table_name(name), temporary = temporary)
```
```suggestion
  df <- copy_to(cdm_get_src(dest), df, unique_db_table_name(name), temporary = temporary, ...)
```

krlmlr:
Thanks!

PR #132, opened 2019-11-08
handling adding tbls with duplicate names

TSchiefer:
- `repair`-argument for `cdm_insert_zoomed_tbl()`

this was by the way the notorious "prime time"-blocker


R/add-tbl.R
```diff
@@ -24,15 +24,25 @@ cdm_add_tbl <- function(dm, ..., repair = "check_unique") {
...
-  dm <- cdm_select_tbl_impl(dm, selected)
+  cdm_add_tbl_impl(dm, new_tables, names_list$new_names)
+}
 
-  new_names <- all_names[seq2(length(old_names) + 1, length(all_names))]
-  cdm_add_tbl_impl(dm, new_tables, new_names)
+figure_out_names <- function(old_names, new_names, repair = "check_unique") {
+  all_names <- tryCatch(
+    vctrs::vec_as_names(c(old_names, new_names), repair = repair),
+    error = function(e) {
```
krlmlr: Can we use a more specific error class here?

R/zoom.R
```diff
@@ -67,10 +68,16 @@ get_zoomed_tbl <- function(dm) {
 #' @param new_tbl_name Name of the new table
 #'
 #' @export
-cdm_insert_zoomed_tbl <- function(dm, new_tbl_name) {
+cdm_insert_zoomed_tbl <- function(dm, new_tbl_name, repair = "check_unique") {
```
krlmlr: Should we use `"unique"` as default here and for `cdm_add_tbl()` ? Can we allow `new_tbl_name = NULL` ? Do we also expose the `quiet` argument?

R/add-tbl.R
```diff
@@ -24,15 +24,25 @@ cdm_add_tbl <- function(dm, ..., repair = "check_unique") {
...
-  new_old_names <- all_names[seq_along(old_names)]
-
-  selected <- set_names(old_names, new_old_names)
-  dm <- cdm_select_tbl_impl(dm, selected)
+  cdm_add_tbl_impl(dm, new_tables, names_list$new_names)
+}
 
-  new_names <- all_names[seq2(length(old_names) + 1, length(all_names))]
-  cdm_add_tbl_impl(dm, new_tables, new_names)
+figure_out_names <- function(old_names, new_names, repair = "check_unique") {
```
krlmlr: Maybe `repair_table_names()` ?

R/zoom.R
TSchiefer: Somehow I am not convinced by `"unique"` as default. But of course (together with accepting `new_tbl_name = NULL`) it would have the advantage, that `cdm_insert_zoomed_tbl()` would work when only providing the `dm` argument. Maybe discuss tomorrow?

no strong opinion about exposing `quiet`.

R/add-tbl.R
TSchiefer: Why is `cdm_error_full("need_unique_names")` not specific enough?

krlmlr: This is about `error = ...`, is there a more specific error we can catch?

R/zoom.R
TSchiefer: Implemented it with `"unique"` as default and for cdm_add_tbl(). Allowed `new_tbl_name = NULL`. Exposed the `quiet` argument.

PR #139, opened 2019-11-12
'compute.dm()' applies filters and then computes

### LINKED ISSUE #135: compute() should call compute() on the tables?
after calling `cdm_apply_filters()` .

TSchiefer:
- added `compute()` for all tables after filtering
- test

fixes #135


krlmlr:
Thanks!

Issue #69, opened 2019-09-21
Redesign constructors and validators


krlmlr:
- `dm()` should be replaced by logic to extract pk/fk information from a datamodelr object
- `new_dm()` becomes `dm()`, no longer accepts datamodelr object
- `new_dm2()` becomes `new_dm()` -- a thin wrapper around `tibble()` after #57. Need to revisit interface carefully. Maybe extract `patch_dm()` to update parts of an existing dm?

krlmlr:
Do we need to do anything else here?

TSchiefer:
Extended `validate_dm()` coming up

PR #140, opened 2019-11-12
avoid 'data_model' arguments

### LINKED ISSUE #118: adapt 2nd parameter of 'new_dm()'
Since we do not support the class `data_model` anymore, we should not require it as an argument to `dm()` and `new_dm()`

TSchiefer:
remove `data_model`-argument from

1. `dm()`
2. `new_dm()`

only place where the `data_model`-argument plays a larger role is in `cdm_learn_from_db()`. If we merge this PR, we need a new issue to refactor `cdm_learn_from_db()`.

fixes #118


TSchiefer:
Just saw the overlaps between #118 and #69, so not sure how useful this one here is.

R/dm.R
```diff
@@ -1,13 +1,18 @@
...
+#' `dm()` creates a `dm` object from one or multiple [tbl] objects
+#' (tibbles or lazy data objects).
 #'
+#' @param ... Tables to add to the `dm`.  If no names are provided, the tables
+#'   are auto-named.
+#' @param .name_repair Options for name repair.
+#'   Forwarded as `repair` to [vctrs::vec_as_names()].
 #' @param src A \pkg{dplyr} table source object.
-#' @param data_model A \pkg{datamodelr} data model object, or `NULL`.
+#' @param table_names A character vector of the table names to include.
```
bbecane:
```suggestion
 @param table_names A character vector of the names of the tables to include.
```

krlmlr: Thanks. Please keep the `#'` .

R/dm.R
```diff
@@ -37,118 +44,82 @@
...
-  new_dm(tables, data_model)
+  new_dm(tbls)
 })
 
 #' Low-level constructor
 #'
-#' `new_dm()` only checks if the inputs are of the correct class.
-#' If called without arguments, an empty `dm` object is created.
+#' `new_dm()` doesn't perform any checks on the input.
+#' You may need to double-check the returned object with `validate_dm()`.
```
bbecane: Is  "If called without arguments, an empty `dm` object is created." not true any more?

krlmlr: It is still true, but also holds for `dm()`. I wasn't sure it's worth pointing out here.

krlmlr: I'll add it back.

krlmlr: Done.

krlmlr:
Thanks, @bbecane. I saw only one change, is this correct?

bbecane:
Correct

krlmlr:
Thanks!

PR #145, opened 2019-11-13
chose tables in 'cdm_apply_filters()'

### LINKED ISSUE #126: apply filters only to selected tbls?
Maybe we could support a second argument to `cdm_apply_filters()`, giving the possibility to specify which tables we want to use `tbl()` for. Might save time in some situations (only interested in effect of filtering table "a" on table "b", but not other tables).

TSchiefer:
- new argument `...` to `cdm_apply_filters()` to specify, which tables should be filtered
- tests

closes #126

### DIFF
--- a/R/filter-dm.R
+++ b/R/filter-dm.R
@@ -34,7 +34,8 @@
 #'   See `vignette("programming", package = "dplyr")`
 #'   for an introduction to these concepts.
 #'
-#' For `cdm_apply_filters()`: Unquoted names of the tables to apply the filters to.
+#' For `cdm_apply_filters()`: Unquoted names of the tables to apply the filters to. {tidyselect}-helpers are supported.
+#' In case `...` is left empty, the filters will be applied to all tables.
 #'
 #' @examples
 #' library(dplyr)
@@ -100,7 +101,7 @@ cdm_apply_filters <- function(dm, ...) {
   vars <- tidyselect_table_names(dm)
   selected <- tidyselect::vars_select(vars, ...)
   # in case of empty ellipsis all tables should be in selection
-  if (is_empty(selected)) selected <- names(dm)
+  if (is_empty(enexprs(...))) selected <- names(dm)
   new_def <- cdm_get_def(dm) %>%
     mutate(data = map(table, ~if_else(.x %in% selected, list(tbl(dm, .x)), data[.x == table])) %>% flatten)
 
--- a/R/filter-dm.R
+++ b/R/filter-dm.R
@@ -116,14 +116,14 @@ set_filter_for_table <- function(dm, table, filter_exprs, zoomed) {
 #' @export
 cdm_apply_filters <- function(dm, ...) {
   check_not_zoomed(dm)
+  def <- cdm_get_def(dm)
   vars <- tidyselect_table_names(dm)
-  selected <- tidyselect::vars_select(vars, ...)
   # in case of empty ellipsis all tables should be in selection
-  if (is_empty(enexprs(...))) selected <- names(dm)
-  new_def <- cdm_get_def(dm) %>%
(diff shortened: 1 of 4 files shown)


R/filter-dm.R
```diff
@@ -89,13 +91,16 @@ set_filter_for_table <- function(dm, table, filter_exprs, zoomed) {
...
   check_not_zoomed(dm)
-  def <- cdm_get_def(dm)
-
-  def$data <- map(def$table, ~ tbl(dm, .))
-
-  cdm_reset_all_filters(new_dm3(def))
+  vars <- tidyselect_table_names(dm)
+  selected <- tidyselect::vars_select(vars, ...)
+  # in case of empty ellipsis all tables should be in selection
+  if (is_empty(selected)) selected <- names(dm)
```
krlmlr: Only `if (missing(...))` (before even calling tidyselect)?

R/filter-dm.R
```diff
@@ -22,14 +22,16 @@
...
-#' @param ... Logical predicates defined in terms of the variables in `.data`, passed on to [dplyr::filter()].
+#' @param ... For `cdm_filter()`: Logical predicates defined in terms of the variables in `.data`, passed on to [dplyr::filter()].
 #' Multiple conditions are combined with `&` or `,`. Only rows where the condition evaluates
 #' to TRUE are kept.
 #'
 #' The arguments in ... are automatically quoted and evaluated in the context of
 #' the data frame. They support unquoting and splicing. See `vignette("programming", package = "dplyr")`
 #' for an introduction to these concepts.
 #'
+#' For `cdm_apply_filters()`: Unquoted names of the tables to apply the filters to.
```
krlmlr: Can we split documentation files?

R/filter-dm.R
```diff
@@ -89,13 +91,16 @@ set_filter_for_table <- function(dm, table, filter_exprs, zoomed) {
...
-
-  def$data <- map(def$table, ~ tbl(dm, .))
-
-  cdm_reset_all_filters(new_dm3(def))
+  vars <- tidyselect_table_names(dm)
+  selected <- tidyselect::vars_select(vars, ...)
+  # in case of empty ellipsis all tables should be in selection
+  if (is_empty(selected)) selected <- names(dm)
+  new_def <- cdm_get_def(dm) %>%
+    mutate(data = map(table, ~if_else(.x %in% selected, list(tbl(dm, .x)), data[.x == table])) %>% flatten)
```
krlmlr: `def$data[selected] <- map(def$data[selected], ...)`

R/filter-dm.R
```diff
@@ -89,13 +91,16 @@ set_filter_for_table <- function(dm, table, filter_exprs, zoomed) {
...
+
+  cdm_reset_all_filters(new_dm3(new_def))
```
krlmlr: Is this still correct?

TSchiefer: Very valid point. In my opinion, it should be still correct and we should add something like the following to the documentation: "A call to `cdm_apply_filters()` resets all filter conditions that are stored in the `dm`, independent if the filters are only applied to a subset of the tables"

But I'm open to suggestions

krlmlr: I forgot about the duality here. In this case let's shelve this for now until we have a better understanding of filtering and applying filters.

TSchiefer: It somehow doesn't seem very likely to me, that we're going to move away far from our current API regarding the filtering, apart from #184, which doesn't affect this issue though. I might be wrong, but it seems to me that it's just a question of deciding whether we want the filter conditions still to be stored or dropped after `cdm_apply_filters()`

krlmlr:
Let's shelve for now.

PR #146, opened 2019-11-13
bugfix: 'mutate.zoomed_dm()' when no keys are tracked

TSchiefer:
problem was `set_names(NULL)` would throw an error

- helper function `set_names_or_null()`
- test for the failing case

@krlmlr this patch is necessary for the demo.

### DIFF
--- a/R/dplyr.R
+++ b/R/dplyr.R
@@ -68,9 +68,11 @@ mutate.dm <- function(.data, ...) {
 #' @export
 mutate.zoomed_dm <- function(.data, ...) {
   tbl <- get_zoomed_tbl(.data)
-  mutated_tbl <- mutate(tbl, ...)
+  quos <- enquos(...)
+  mutated_tbl <- mutate(tbl, !!!quos)
   # all columns that are not touched count as "selected"; names of "selected" are identical to "selected"
-  selected <- set_names(setdiff(names(get_tracked_keys(.data)), names(enquos(..., .named = TRUE))))
+  # in case no keys are tracked, `set_names(NULL)` would throw an error
+  selected <- set_names(setdiff(names2(get_tracked_keys(.data)), names(quos)))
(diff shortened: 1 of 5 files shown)


R/dplyr.R
```diff
@@ -70,7 +70,8 @@ mutate.zoomed_dm <- function(.data, ...) {
   tbl <- get_zoomed_tbl(.data)
   mutated_tbl <- mutate(tbl, ...)
   # all columns that are not touched count as "selected"; names of "selected" are identical to "selected"
-  selected <- set_names(setdiff(names(get_tracked_keys(.data)), names(enquos(..., .named = TRUE))))
+  # in case no keys are tracked, `set_names(NULL)` would throw an error
+  selected <- set_names_or_null(setdiff(names(get_tracked_keys(.data)), names(enquos(..., .named = TRUE))))
```
krlmlr: If we need to `enquos()`, please extract and call `mutate(!!!quos)` above.

```suggestion
  selected <- set_names(setdiff(names2(get_tracked_keys(.data)), names(quos)))
```

Why would `get_tracked_keys()` return `NULL` anyway, should it return `character()` in this case?

TSchiefer: Good point, `get_tracked_keys()` now returns an empty named character, when no keys are tracked

R/dplyr.R
```diff
@@ -68,9 +68,11 @@ mutate.dm <- function(.data, ...) {
 #' @export
 mutate.zoomed_dm <- function(.data, ...) {
   tbl <- get_zoomed_tbl(.data)
-  mutated_tbl <- mutate(tbl, ...)
+  quos <- enquos(...)
```
krlmlr: Same as in `mutate.tbl_df()`:

```suggestion
  quos <- enquos(..., .named = TRUE)
```

PR #148, opened 2019-11-13
demo

TSchiefer:
which other functions should be included in the demo?

- [ ] `cdm_insert_zoomed_tbl()`
- [ ] `cdm_enum_fk_candidates()`
- [ ] `cdm_squash_to_tbl()`
- [ ] `decompose_table()`
- [ ] `cdm_nrow()`
- [ ] more examples of {dplyr}-workflow for `zoomed_dm`


R/dplyr.R
```diff
@@ -70,7 +70,8 @@ mutate.zoomed_dm <- function(.data, ...) {
   tbl <- get_zoomed_tbl(.data)
   mutated_tbl <- mutate(tbl, ...)
   # all columns that are not touched count as "selected"; names of "selected" are identical to "selected"
-  selected <- set_names(setdiff(names(get_tracked_keys(.data)), names(enquos(..., .named = TRUE))))
+  # in case no keys are tracked, `set_names(NULL)` would throw an error
+  selected <- set_names_or_null(setdiff(names(get_tracked_keys(.data)), names(enquos(..., .named = TRUE))))
```
krlmlr: Need to rebase after merging #146.

demo/00Index
```diff
@@ -1,2 +1,3 @@
 berlin	Demo for presentation at the 16th Berlin R meetup, July 2019
 zurich	Demo for presentation at the Zurich R meetup, September 2019
+demo Demo for presentation at Energie360, November 2019
```
krlmlr:
```suggestion
zurich-2 Demo for presentation at Energie360, November 2019
```

TSchiefer: To avoid failing R CMD check I would need to replace the entry for zurich.R in the index file, if  I overwrite zurich.R. Guess we can also sort this out then later

TSchiefer:
this is slowly getting somewhere. #156 could be also used for this demo if merged in time

krlmlr:
Thanks! I decided to reorganize so that creation comes first with existing tables, and then computation of the surrogate key comes in a separate step.

PR #152, opened 2019-11-14
Bugfix: key tracking for 'slice()'-method

### LINKED ISSUE #151: slice() can invalidate primary keys
if positive indexes are repeated.

TSchiefer:
- now drops potential primary key if positive indices are repeated
- now only accepts integer and numeric vectors without any decimal place different from `0`

fixes #151

### DIFF
--- a/R/dplyr.R
+++ b/R/dplyr.R
@@ -159,19 +159,17 @@ slice.dm <- function(.data, ...) {
 #' @export
 slice.zoomed_dm <- function(.data, idx) {
   if (is_missing(idx)) return(.data)
-
+  # we want to exclude errors triggered from {vctrs}, since hard to interpret and catching would need case differentiation, i.e. check first
   if (!inherits(idx, "integer")) {
-    # numeric vector only accepted if identical to itself as integer
-    if (!inherits(idx, "numeric") || !all(as.integer(idx) == idx)) abort_need_int(class(idx))
+    if (!inherits(idx, "numeric") || !all(as.integer(idx) == idx)) abort_need_int(class(idx)) else idx <- as.integer(idx)
   }
+  idx <- vctrs::vec_as_index(idx, nrow(.data))
   # FIXME: will be easier if we have an extra row in `def` for the zoomed table
-  if (cdm_has_pk(.data, !!orig_name_zoomed(.data))) {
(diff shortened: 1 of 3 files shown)


R/dplyr.R
```diff
@@ -157,8 +157,21 @@ slice.dm <- function(.data, ...) {
 }
 
 #' @export
-slice.zoomed_dm <- function(.data, ...) {
-  replace_zoomed_tbl(.data, slice(get_zoomed_tbl(.data), ...))
+slice.zoomed_dm <- function(.data, idx) {
+  if (is_missing(idx)) return(.data)
+
+  if (!inherits(idx, "integer")) {
```
krlmlr: I forgot to mention that we can use `vctrs::vec_as_index()` to create a positive index vectors from all inputs.

R/dplyr.R
```diff
@@ -157,8 +157,21 @@ slice.dm <- function(.data, ...) {
...
+    # numeric vector only accepted if identical to itself as integer
+    if (!inherits(idx, "numeric") || !all(as.integer(idx) == idx)) abort_need_int(class(idx))
+  }
+  # FIXME: will be easier if we have an extra row in `def` for the zoomed table
+  if (cdm_has_pk(.data, !!orig_name_zoomed(.data))) {
```
krlmlr: Can we init `tracked_keys <- NULL` here and return `replace_zoomed_tbl(..., tracked_keys)` unconditionally?

R/dplyr.R
TSchiefer: would that mean that different from `dplyr::slice()` we would accept logical vectors too(?), or should we actively check against that?

TSchiefer: I would opt for not accepting logical vectors, since people might get into habits that don't work with `dplyr::slice()`

TSchiefer: I kept the `class()`-checks, for reasons mentioned in the comment in line 162

TSchiefer:
Yup, that way we don't lose the usual `dplyr::slice()` functionality, I like that.

I'd say a message in case `.keep_pk = NULL` AND the original zoomed table has a PK (or: in case `.keep_pk = NULL` AND the original zoomed table has a PK AND that PK is still being tracked)

PR #157, opened 2019-11-14
autoselect RHS 'by' columns

### LINKED ISSUE #153: automatically select key column of RHS for join-methods
for methods for `..._join.zoomed_dm` the user should not need to add the `by`-column for the RHS explicitly in argument `select`

TSchiefer:
when user selects columns of the RHS in a join but doesn't include the `by` columns in the selection, we re-add them.

depending on if `by = NULL` or not there is a different behavior:
- if `by = NULL`, we check if the automatically detected `by`-column is renamed in parameter `select` and track that.
- if `by` is explicitly given, we check if its value is in the `names(selected)`

If `by = NULL` and the user chooses to rename another column with the same name as the `by` column would be, we throw an error (there are limits :) )

fixes #153


R/dplyr.R
```diff
@@ -261,17 +261,36 @@ prepare_join <- function(x, y, by, selected, suffix, copy, disambiguate = TRUE)
   if (quo_is_null(select_quo)) {
     selected <- tidyselect::vars_select(all_cols_y, everything())
   } else {
+    # FIXME: should we complain, if the user wishes to rename two or more columns to the same name?
```
krlmlr: Is this the responsibility of tidyselect?

R/dplyr.R
```diff
@@ -261,17 +261,36 @@ prepare_join <- function(x, y, by, selected, suffix, copy, disambiguate = TRUE)
...
-    if (!any(selected == by)) abort_need_to_select_rhs_by(y_name, unname(by))
-
     if (!all(names(by) %in% get_tracked_keys(x))) abort_fk_not_tracked(x_orig_name, y_name)
+    # here we need to check if the RHS `by`-column taken directly from the `dm` key relations
+    # is part of the values of `selected` (even if it is renamed)
+    if (has_length(setdiff(by, selected))) {
+      message(glue("Adding missing `by` variable(s) of RHS table: {commas(tick(setdiff(by, selected)))}"))
+      # re-add RHS-`by` column if not selected:
+      selected <- c(selected, set_names(setdiff(by, selected)))
+      if (anyDuplicated(names(selected))) abort_duplicated_cols_introduced(names(selected[anyDuplicated(names(selected))]))
```
krlmlr: Good catch!

Maybe we could:

```suggestion
      if (anyDuplicated(names(selected))) abort_duplicated_cols_introduced(names(selected)))
```

and compute the duplicates in `abort_...()` (using `unique(x[duplicated(x)])` )?

R/dplyr.R
```diff
@@ -261,17 +261,36 @@ prepare_join <- function(x, y, by, selected, suffix, copy, disambiguate = TRUE)
...
 
   if (is_null(by)) {
+    rename_rhs_by <- TRUE
     by <- get_by(x, x_orig_name, y_name)
-    if (!any(selected == by)) abort_need_to_select_rhs_by(y_name, unname(by))
-
     if (!all(names(by) %in% get_tracked_keys(x))) abort_fk_not_tracked(x_orig_name, y_name)
+    # here we need to check if the RHS `by`-column taken directly from the `dm` key relations
+    # is part of the values of `selected` (even if it is renamed)
+    if (has_length(setdiff(by, selected))) {
```
krlmlr: Can you extract `setdiff(by, selected)` to a variable? (Can we make the behaviors in the two cases more similar?)

R/dplyr.R
TSchiefer: `tidyselect::vars_select()` allows this:

``` r
dplyr::select(tibble::as_tibble(iris), name = Species, name = Sepal.Length)
#> # A tibble: 150 x 2
#>    name    name
#>    <fct>  <dbl>
#>  1 setosa   5.1
#>  2 setosa   4.9
#>  3 setosa   4.7
#>  4 setosa   4.6
#>  5 setosa   5  
#>  6 setosa   5.4
#>  7 setosa   4.6
#>  8 setosa   5  
#>  9 setosa   4.4
#> 10 setosa   4.9
#> # … with 140 more rows
```

<sup>Created on 2019-11-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/dplyr.R
TSchiefer: not sure: in the once case we need `setdiff(by, selected)` and in the case we need `setdiff(by, names(selected))` (at least if we stick to how I understood the situation in the first place)

krlmlr:
Looks like we followed a wrong route. I've added a comment to the original issue.

PR #169, opened 2019-11-20
Some full stops added to params

bbecane:
Incomplete, other files need to be checked as well.

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -8,7 +8,7 @@
 #' `dm()` creates a `dm` object from one or multiple [tbl] objects
 #' (tibbles or lazy data objects).
 #'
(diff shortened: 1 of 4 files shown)


R/dm.R
```diff
@@ -8,7 +8,7 @@
 #' `dm()` creates a `dm` object from one or multiple [tbl] objects
 #' (tibbles or lazy data objects).
 #'
-#' @param ... Tables to add to the `dm`.  If no names are provided, the tables
+#' @param ... Tables to add to the `dm`. If no names are provided, the tables
```
```suggestion
#' @param ... Tables to add to the `dm` object. If no names are provided, the tables
```

bbecane: A dummy comment

PR #173, opened 2019-11-21
extended 'validate_dm'

### LINKED ISSUE #69: Redesign constructors and validators

TSchiefer:
- more checks
- all problems found in this function raise an error of class `cdm_error_full("dm_invalid")`

closes #69

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -187,9 +187,9 @@ validate_dm <- function(x) {
   if (any(table_names == "")) abort_dm_invalid("Not all tables are named.")
   check_col_classes(def)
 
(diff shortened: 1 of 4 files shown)


krlmlr:
Thanks!

PR #188, opened 2019-11-26
'check_key()' with {tidyselect}

TSchiefer:
...since we were already bragging about that in the documentation

### DIFF
--- a/R/key-helpers.R
+++ b/R/key-helpers.R
@@ -64,10 +64,9 @@ check_key <- function(.data, ...) {
   .data <- eval_tidy(data_q)
 
   cols_avail <- colnames(.data)
-  cols_chosen <- tidyselect::vars_select(cols_avail, ...)
   # if no column is chosen, all columns are used for the check
-  if (is_empty(cols_chosen)) cols_chosen <- cols_avail
-  names(cols_chosen) <- set_names(paste0("...", seq_along(cols_chosen)))
+  cols_chosen <- if (is_empty(enexprs(...))) cols_avail else tidyselect::vars_select(cols_avail, ...)
+  if (has_length(cols_chosen)) names(cols_chosen) <- set_names(paste0("...", seq_along(cols_chosen)))
 
   duplicate_rows <-
     .data %>%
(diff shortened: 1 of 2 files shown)


R/key-helpers.R
```diff
@@ -62,16 +62,20 @@ cdm_check_constraints <- function(dm) {
 check_key <- function(.data, ...) {
   data_q <- enquo(.data)
   .data <- eval_tidy(data_q)
-  args <- exprs(...)
-  names(args) <- set_names(paste0("...", seq_along(args)))
+
+  cols_avail <- colnames(.data)
+  cols_chosen <- tidyselect::vars_select(cols_avail, ...)
+  # if no column is chosen, all columns are used for the check
+  if (is_empty(cols_chosen)) cols_chosen <- cols_avail
```
krlmlr: `if (missing(...))` ?

TSchiefer: that throws an error in case a tidyselect-helper function is used, but it does work with `if (is_empty(enexprs(...)))`, see new commits

R/key-helpers.R
```diff
@@ -62,16 +62,19 @@ cdm_check_constraints <- function(dm) {
 check_key <- function(.data, ...) {
   data_q <- enquo(.data)
   .data <- eval_tidy(data_q)
-  args <- exprs(...)
-  names(args) <- set_names(paste0("...", seq_along(args)))
+
+  cols_avail <- colnames(.data)
+  # if no column is chosen, all columns are used for the check
+  cols_chosen <- if (is_empty(enexprs(...))) cols_avail else tidyselect::vars_select(cols_avail, ...)
```
krlmlr: Do we even need the special behavior for an empty argument list here? Can the user ask `check_key(everything())` ?

TSchiefer: 1. not necessarily (it would be a quick way to check if the combination of all columns is a unique key though)
2. yes

krlmlr: Let's keep default semantics then.

TSchiefer: Just to make sure: this does mean I can merge without further changes, right?

krlmlr: "Default semantics" means that we don't need to special-case behavior of `...` (but we need to watch out what happens with an empty selection). I'll tweak.

krlmlr:
Thanks!

PR #197, opened 2019-11-28
set colors using {tidyselect}

### LINKED ISSUE #162: Support `tidyselect` for setting colors?
Something like `cdm_set_colors(dm, ends_with("lookup") = "blue")`

TSchiefer:
- second function for setting colors, currently `cdm_set_colors2()`
- syntax: `cdm_set_colors2(cdm_nycflights13(), blue = starts_with("air"), green = contains("h"))`

(in #162 the syntax was suggested like `cdm_set_colors(dm, ends_with("lookup") = "blue")`, but this doesn't seem to be a proper function call...)

Maybe we could offer both variants for setting colors?

closes #162


TSchiefer:
BTW, the reason I requested your review for a WIP branch was to get an initial feedback on how to proceed from here, since it turned out to be a bit different from what I originally excpected.

I reckon that it's not a good idea to have just one function that recognizes, if the tables are the names in the ellipsis or the colors (this for examples would be exceedingly tricky: `cdm_set_colors(color_dm, orange = blue, blue = orange)`)

I think the questions are now: 
1. should we scrap the idea of {tidyselect}?
2. should we replace the fall-through syntax with the {tidyselect} one?
3. should we keep both? If yes, how could we call the different functions? Or would it be one function and you'd chose in a boolean parameter what the names in the ellipsis are?

krlmlr:
Thanks. I don't mind moving to the new notation altogether. The old notation was rather nonstandard anyway. We can keep the old implementation for now and forward to it if `all(names(quos) %in% src_tbls(dm))` and `!all(names(quos) %in% colors)`, with a warning.

krlmlr:
Should `dm_get_colors()` also return the colors in the names and the table names in the value, for symmetry?

TSchiefer:
Yes, I think it's a good idea to change the implementation of `dm_get_colors()` in the way you suggested

R/draw-dm.R
```diff
@@ -146,6 +146,47 @@ dm_get_all_columns <- function(x) {
 #'   dm_draw()
 #' @export
 dm_set_colors <- function(dm, ...) {
+  avail_cols <- dm_get_available_colors()$dm
+  quos <- enquos(...)
+  # If old fall-through syntax is being used:
```
krlmlr: Let's remove this compatibility code.

R/draw-dm.R
```diff
@@ -146,6 +146,47 @@ dm_get_all_columns <- function(x) {
...
+    all(names(quos) %in% src_tbls(dm))) {
+    warning(paste0(
+      "When setting colors with `dm_set_colors()`, ",
+      "the syntax is now as follows: `blue = table_name1, ",
+      "orange = table_name2` etc. (tidyselect supported) ",
+      "Forwarding the function call ",
+      "to the old implementation of `dm_set_colors()`."
+    ))
+    return(dm_set_colors2(dm, !!!quos))
+  }
```
krlmlr: Can we make this an error instead of forwarding?

R/draw-dm.R
```diff
@@ -146,6 +146,47 @@ dm_get_all_columns <- function(x) {
...
+
+  avail_tables <- src_tbls(dm)
+  selected_tables <- tidyselect::vars_select(avail_tables, ...) %>%
+    # names will get integer suffixes if duplicated ("blue1", "blue2", ...)
```
krlmlr: Do you have an example? I'd rather remove the `set_names()` call below.

R/draw-dm.R
```diff
@@ -146,6 +146,47 @@ dm_get_all_columns <- function(x) {
...
+    set_names(gsub("[0-9]*", "", names(.)))
+  sel_colors <- names(selected_tables)
+  if (!all(sel_colors %in% colors$dm)) {
+    abort_wrong_color(paste0("`", colors$dm, "` ", colors$nb))
```
krlmlr: Should formatting be the responsibility of `abort_wrong_color()` ?

R/draw-dm.R
```diff
@@ -146,6 +146,47 @@ dm_get_all_columns <- function(x) {
...
+  }
+
+  display_df <- tibble(
+    table = selected_tables,
+    new_display = colors$datamodelr[match(sel_colors, colors$dm)]
```
krlmlr: Can we drop support for datamodelr colors, and instead support a much wider range of colors, and also arbitrary RGB colors?

R/draw-dm.R
TSchiefer:
``` r
avail_tables <- c("t1", "t2")
tidyselect::vars_select(avail_tables, blue = dplyr::starts_with("t"))
#> blue1 blue2 
#>  "t1"  "t2"

# it works when only using 1-element variables:
tidyselect::vars_select(avail_tables, blue = t1, blue = t2)
#> blue blue 
#> "t1" "t2"

# it doesn't work when using multi-element vectors:
tidyselect::vars_select(avail_tables, blue = c(t1, t2))
#> blue1 blue2 
#>  "t1"  "t2"
```

<sup>Created on 2020-01-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

But I agree, that it's currently not a good solution, mainly if at some point we also allow hex color codes that may also contain digits. Not sure how to make fail-safe.

krlmlr: The dev version of tidyselect has `eval_select()` with options to control naming, we should be able to use it.

TSchiefer: I just uploaded some changes: now it behaves differently depending on {tidyselect}-version. The only thing is, that in case someone has an old {tidyselect}-version, a NOTE will be produced by R CMD check:
```r
> checking dependencies in R code ... NOTE
  Missing or unexported object: ‘tidyselect::eval_select’
```
cause it's a new function...

R/draw-dm.R
TSchiefer: Now there is a much larger choice of colors (`mauve` is still missing though, sorry ;) ). Basically all hex colors can be used for setting the header color or alternatively you can use the color names given by `grDevices::colors()`.

I still left the {datamodelr} colortables untouched, cause the legacy `cdm_set_colors()` still uses them. But the implementation for the new function is independent of them.

R/draw-dm.R
```diff
@@ -114,38 +111,86 @@ dm_get_all_columns <- function(x) {
...
-#'   weather = "dark_blue"
-#' )
 #' dm_nycflights13(color = FALSE) %>%
-#'   dm_set_colors(!!!new_colors) %>%
+#'   dm_set_colors(!!!dm_get_colors(dm_nycflights13())) %>%
 #'   dm_draw()
 #' @export
 dm_set_colors <- function(dm, ...) {
+  # only those tables for which a color is set (others get default color when drawing)
+  quos <- enquos(...) %>% extract(names(.) != "")
```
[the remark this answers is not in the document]
TSchiefer: true, throwing an error now

R/draw-dm.R
```diff
@@ -114,38 +111,86 @@ dm_get_all_columns <- function(x) {
...
+        quos,
+        function(quos_sel) tidyselect::vars_select(avail_tables, !!quos_sel)
+      )
+    }
+
+  # create "color-vector" of appropriate repetitions for each color
+  num_for_each_col <- map_int(selected_tables, length)
+  sel_colors <- rep(hexcols, num_for_each_col)
+
+  display_df <- tibble(
```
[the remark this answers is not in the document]
TSchiefer: thanks, that's much nicer

R/draw-dm.R
```diff
@@ -114,38 +111,86 @@ dm_get_all_columns <- function(x) {
...
-#'     airports = ,
-#'     airlines = ,
-#'     planes = "yellow",
-#'     weather = "dark_blue"
+#'     darkblue = starts_with("air"),
+#'     "#5986C4" = flights
 #'   ) %>%
 #'   dm_draw()
 #'
 #' # Splicing is supported:
```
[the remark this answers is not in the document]
TSchiefer: I wasn't sure, how you imagine showing the output. I implemented it like this:
```r
#' # Splicing is supported:
#' nyc_cols <- dm_get_colors(dm_nycflights13())
#'
#' nyc_cols
#' #>    #FFA07A    #FFA07A    #ADD8E6    #FFA07A    #90EE90
#' #> "airlines" "airports"  "flights"   "planes"  "weather"
#'
#' dm_nycflights13(color = FALSE) %>%
#'   dm_set_colors(!!!nyc_cols) %>%
#'   dm_draw()
```

TSchiefer:
R CMD check still failing because of `tidyselect::eval_select()` being only available in dev version of {tidyselect}, causes a note in R CMD check. This problem will of course be solved once {tidyselect} is updated on CRAN. But can we somehow circumvent this?

R/datamodelr-code.R
```diff
@@ -272,16 +274,25 @@ bdm_create_graph_list <- function(
...
   g_labels <-
     sapply(names(tables), function(x) {
-      dot_html_label(
-        tables[[x]],
-        title = x,
-        palette_id = data_model$tables[data_model$tables$table == x, "display"],
-        col_attr = col_attr,
-        columnArrows = columnArrows
-      )
+      if (legacy) {
```
krlmlr: Can we add a `legacy` argument to the `dot_html_label()` function, and from there forward as necessary?

R/datamodelr-code.R
```diff
@@ -428,26 +440,51 @@ to_html_table <- function(x,
...
+      header_bgcolor = "#EFEBDD",
+      header_font = "#000000",
+      bgcolor = "#FFFFFF"
+    )
+  } else {
+    header_bgcol_rgb <- col2rgb(palette_id)[, 1]
+    bodycol_rgb <- as.integer(header_bgcol_rgb + (255 - header_bgcol_rgb) * 0.87)
+    bodycol <- rgb(bodycol_rgb[1], bodycol_rgb[2], bodycol_rgb[3], maxColorValue = 255)
+    # if header background too dark, use white font color
+    header_font <- if (sum(header_bgcol_rgb) < 300) "#FFFFFF" else "#000000"
```
krlmlr:
```suggestion
    header_font <- if (is_dark_color(header_bcgol_rgb)) "#FFFFFF" else "#000000"
```

with a suitable `is_dark_color()` function.

R/datamodelr-code.R
```diff
@@ -428,26 +440,51 @@ to_html_table <- function(x,
...
+    line_color_rgb <- as.integer(header_bgcol_rgb / 3.)
+    line_color <- rgb(line_color_rgb[1], line_color_rgb[2], line_color_rgb[3], maxColorValue = 255)
```
krlmlr: Same here, `line_color <- line_color_for_bg(header_...)`

R/draw-dm.R
```diff
@@ -114,38 +111,83 @@ dm_get_all_columns <- function(x) {
     unnest(value)
 }
 
-#' dm_set_colors()
+#' `dm_set_colors()`
 #'
 #' `dm_set_colors()` allows to define the colors that will be used to display the tables of the data model.
+#' The colors can either be either specified with hex color codes or using the names of the built-in R colors.
+#' An overview of the colors corresponding to the standard color names can be found at
```
krlmlr:
```suggestion
#' An overview of the colors corresponding to the standard color names can be found at
#' the bottom of
```

R/utils.R
```diff
@@ -33,6 +33,25 @@ is_syntactic <- function(x) {
   x == make.names(x)
 }
 
+is_hex_color <- function(x) {
```
krlmlr: Maybe move this and the to-be-implemented color helpers to a new file  `colors.R` ?

krlmlr:
For tidyselect, try:

```r
usethis::use_dev_package("tidyselect")
```

TSchiefer:
> Thanks, looks good. What's the difference between `[c]dm_draw()` and between `[c]dm_nycflights()`?

The `cdm_`-variants work with the colors in the old way (the whole original implementation is still used) and the `dm_`-variants use the new syntax and colors. As long as you don't mix `cdm_` and `dm_`, things will work.

``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter
cdm_nycflights13() %>% cdm_set_colors(airlines = , airports = "yellow") %>% cdm_draw()
#> Warning: `cdm_nycflights13()` is deprecated as of dm 0.1.0.
#> Please use `dm_nycflights13()` instead.
#> This warning is displayed once per session.
#> Call `lifecycle::last_warnings()` to see where this warning was generated.
#> Warning: `cdm_set_colors()` is deprecated as of dm 0.1.0.
#> Please use `dm_set_colors()` instead.
#> This warning is displayed once per session.
#> Call `lifecycle::last_warnings()` to see where this warning was generated.
#> Warning: `cdm_draw()` is deprecated as of dm 0.1.0.
#> Please use `dm_draw()` instead.
#> This warning is displayed once per session.
#> Call `lifecycle::last_warnings()` to see where this warning was generated.
```



``` r

dm_nycflights13() %>% dm_set_colors(yellow3 = starts_with("air")) %>% dm_draw()
```



<sup>Created on 2020-01-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Can I `dm_draw()` a data model where colors are defined with `cdm_set_colors()` ?

TSchiefer:
No, the colors in `def` are then called `accent1` (etc.), and `dm_draw()` just understands hex color codes and the built-in color names

krlmlr:
Can `cdm_set_colors()` forward to `dm_set_colors()`? I don't really care what `cdm_get_colors()` returns.

krlmlr:
Also, for now `dm_nycflights()` should return colors that are consistent with the original implementation -- we have other material that uses these colors.

TSchiefer:
Should work to translate the input of `cdm_set_colors()` to something that `dm_set_colors()` understands, gonna do that.

And good idea about keeping the original colors of `dm_nycflights13()`, also gonna change that.

krlmlr:
Can you please merge master and ensure the build is green?

R/colors.R
```diff
@@ -0,0 +1,36 @@
...
+  # and https://www.w3.org/WAI/WCAG21/Techniques/general/G18.html
+  # you should calculate luminance first and then compare its contrast to white and black
+  #
+  # But after evaluating results from this approach, it seems more natural to use a formula like this:
+  sum(rgb / 255. * c(0.3, 0.5, 0.2)) < 0.6
+  # the weights are inspired by the sources mentioned above (only the differences between them are much smaller here)
+}
+
+is_hex_color <- function(x) {
+  grepl("^#[A-Fa-f0-9]{6}$", x)
```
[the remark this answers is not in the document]
TSchiefer: What is the added value of `?:`? If I leave it out, like this:
```r
grepl("^#[A-Fa-f0-9]{6}([A-Fa-f0-9]{2})?$", x)
```
it seems to work fine as well.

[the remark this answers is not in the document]
krlmlr: `(?:)` indicates a non-capturing group, important for replace operations but otherwise the same.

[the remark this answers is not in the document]
TSchiefer: Ok, thanks, very good to know. Since in our case it's the same, I will leave it out if that's ok.

TSchiefer:
In the R 3.3  Travis test we get again the error:
`Authors@R field gives persons with no valid roles` (cynkra and E360)

also seems as if tests fail, even though locally everything seems to be alright, but then again https://travis-ci.com/krlmlr/dm/jobs/276220267 also did succeed.

Not sure what is happening, it would be nice if you could have a look.

krlmlr:
Thanks!

krlmlr:
I'll take care of the builds. We can't avoid the NOTE for older R versions with the fnd role, but that's fine.

Issue #150, opened 2019-11-13
Applying filters on data frames is slow


krlmlr:
Seen in a complex-ish data model with ~10 tables.

Are we joining too much?

Also, `cdm_apply_filters()` should abort in the presence of cycles?

PR #198, opened 2019-11-29
test for cycle in 'get_all_filtered_connected()'

TSchiefer:
For the test we check in an induced subgraph of all connected tables/nodes (related to one "start" table) if `N(E(graph)) >= N(V(graph))`

The only problem is, that we are sometimes too strict: assume a graph `(t1 - t2, t2 - t3, t3 - t4, t4 - t2)` and we filter `t1` and we request `t2`, then it fails cause the subgraph consists of `(t1, t2, t3, t4)`. But, if we want to resolve this properly, we might need to work with something like `igraph::all_simple_paths()` for many pairs of tables, which would slow down things quite a bit. So I tend to think, that this solution here is not so bad.

Tests are included in this PR.

addresses 2nd part of issue #150

### DIFF
--- a/R/filter-dm.R
+++ b/R/filter-dm.R
@@ -160,6 +160,12 @@ get_all_filtered_connected <- function(dm, table) {
   # Using only nodes with finite distances (=in the same connected component)
   # as target. This avoids a warning.
   target_tables <- names(finite_distances)
+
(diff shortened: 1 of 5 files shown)


krlmlr:
Thanks!

PR #208, opened 2019-12-03
update doc and tweak implementation of 'check_key()'

TSchiefer:
minor changes...

### DIFF
--- a/R/key-helpers.R
+++ b/R/key-helpers.R
@@ -71,7 +71,9 @@ check_key <- function(.data, ...) {
   duplicate_rows <-
     .data %>%
     count(!!!syms(cols_chosen)) %>%
+    select(n) %>%
     filter(n > 1) %>%
+    head(1) %>%
     collect()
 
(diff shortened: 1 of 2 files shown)


R/key-helpers.R
```diff
@@ -72,7 +71,6 @@ check_key <- function(.data, ...) {
   duplicate_rows <-
     .data %>%
     count(!!!syms(cols_chosen)) %>%
-    count(n) %>%
```
krlmlr: What's the reason for this change?

TSchiefer: for each group we get a count of how many times it appears (n = ?). So we just need to check if `n > 1`, we don't benefit of counting how many times we get `n = 1`, if I am not mistaken

krlmlr: The `count(n)` compresses what's returned from the database. We could do something like `. %>% select(n) %>% filter(n > 1) %>% head(1)` to achieve minimum overhead for both database and data frame.

krlmlr:
Thanks!

PR #210, opened 2019-12-03
pkgdown references update

### LINKED ISSUE #230: output colnames for data frame producing funs
For functions that return a data frame we need to check if the output column names make sense.

### LINKED ISSUE #231: collect all non-dm functions in a separate sub-topic
in _pkgdown.yml

Still needs to be determined, if only one such sub-topic for all such functions, or should there be more of those sub-topics, depending on the respective role of the non-dm functions.

### LINKED ISSUE #232: stricter handling of `dm`-class in functions
Each function either accepts a `dm` or a `zoomed_dm`, never both.

### LINKED ISSUE #233: rename/merge several functions
- [x] `dm_learn_from_db`-> `dm_from_src`
- [x] `check_` -> `examine_` for cardinality and constraints
- [ ] `dm_copy_to` -> `copy_to` (will become a method)
- [x] `*_zoomed_tbl` -> `*_zoomed`
- [x] `dm_zoom_out` -> `dm_discard_zoomed`

TSchiefer:
- completing list of exported functions in "_pkgdown.yml"
- slightly tweaking where certain functions belong (topic)
- `dm_zoom_out()` -> `dm_discard_zoomed()`
- `dm_update_zoomed_tbl()` -> `dm_update_zoomed()` 
- `dm_insert_zoomed_tbl()` -> `dm_insert_zoomed()`

fixes #230
fixes #231
fixes #232
closes #233


pkgdown/_pkgdown.yml
```diff
@@ -83,6 +100,11 @@ reference:
   contents:
   - dm_nycflights13
   - dm_nrow
+  - dm_disambiguate_cols
+  - dm_paste
+  - dm_select
+  - dm_rename
+  - browse_docs
```
krlmlr: Do we still need this?

TSchiefer:
potentially `dm_select()`, `dm_rename()` and `dm_disambiguate_cols()` could get their own category, something like "Managing the columns of the tables of the `dm`"

pkgdown/_pkgdown.yml
```diff
@@ -42,26 +54,30 @@ reference:
...
   - dm_learn_from_db
 
-- title: Utilizing foreign key relations
-  contents:
-  - dm_is_referenced
-  - dm_get_referencing_tables
-
 - title: Table surgery
   contents:
   - decompose_table
```
krlmlr: Could we do `dm_separate_tbl()` and `dm_unite_tbl()` instead, without corresponding non-table functions?

pkgdown/_pkgdown.yml
```diff
@@ -25,6 +33,7 @@ reference:
   - dm_get_all_pks
   - dm_rm_pk
   - dm_enum_pk_candidates
+  - enum_pk_candidates
```
krlmlr: Can you please collect all non-dm functions in a separate sub-topic?

pkgdown/_pkgdown.yml
```diff
@@ -8,14 +8,22 @@ reference:
...
+  - dm_from_src
   - new_dm
-  - is_dm
   - as_dm
+  - is_dm
+  - validate_dm
   - dm_get_src
   - dm_get_con
   - dm_get_tables
   - dm_get_filter
```
krlmlr: `dm_get_filters()` ?

pkgdown/_pkgdown.yml
```diff
@@ -25,6 +33,7 @@ reference:
   - dm_get_all_pks
```
krlmlr: For functions that return a data frame we also need to check if the output column names make sense.

pkgdown/_pkgdown.yml
```diff
@@ -83,6 +100,11 @@ reference:
   contents:
   - dm_nycflights13
   - dm_nrow
+  - dm_disambiguate_cols
+  - dm_paste
+  - dm_select
+  - dm_rename
```
krlmlr: Move this to a separate topic, close to "Filtering"?

pkgdown/_pkgdown.yml
```diff
@@ -83,6 +100,11 @@ reference:
   contents:
   - dm_nycflights13
   - dm_nrow
+  - dm_disambiguate_cols
+  - dm_paste
```
krlmlr: Move to "Creating ..."?

pkgdown/_pkgdown.yml
```diff
@@ -83,6 +100,11 @@ reference:
   contents:
   - dm_nycflights13
   - dm_nrow
+  - dm_disambiguate_cols
```
krlmlr: Move to "Flattening"?

pkgdown/_pkgdown.yml
TSchiefer: The `browse_docs()`? Not sure

pkgdown/_pkgdown.yml
TSchiefer: sounds good! The `dm_unite_tbl()` would then be pretty much like the `dm_flatten_to_tbl()` with the result reinserted into the `dm`.

Is this still material for milestone 0.1.0.?

pkgdown/_pkgdown.yml
TSchiefer: yes, we should rename this one

pkgdown/_pkgdown.yml
TSchiefer: see #211

pkgdown/_pkgdown.yml
TSchiefer: The `enum_(p,f)k_candidates()` functions work for `zoomed_dm` as well as for `tibble/data.frame`, just saying :)

So generally functions that don't immediately have anything to do with a `dm` should get their own sub-topic? We already have some functions like that (`check_cardinality()`, `check_key()` etc.)

pkgdown/_pkgdown.yml
TSchiefer: Ok, I can move them. Just out of curiosity, why close to "Filtering"?

pkgdown/_pkgdown.yml
TSchiefer: Should I chuck it out?

pkgdown/_pkgdown.yml
krlmlr: Does the `dm_` prefix mean that the first argument is a dm (or a zoomed dm)?

pkgdown/_pkgdown.yml
krlmlr: Filtering and selection are two basic operations that are invoked first when using a ready-made dm as a consumer.

pkgdown/_pkgdown.yml
TSchiefer: It should IMHO mean, that the first argument is a `dm`, and  (almost always) not a `zoomed_dm`. The latter should normally be treated like a `tibble`.

Having said that, so far it is e.g. possible to set keys for a `zoomed_dm` (independent of which table is zoomed) and probably a few other functions will work too.

If we wanted to be really strict, we could say, that we don't allow anything on a `zoomed_dm` that works on a `dm`. That way it would be easier to be consistent with the prefixes.(?)

pkgdown/_pkgdown.yml
TSchiefer: Name of the new category: "Rename or select columns of tables in a `dm`"

Open to constructive criticism :)

TSchiefer:
Rules for prefix:

1. `dm_` should mean, that the first argument is a `dm` (currently - as mentioned above - it can often also be a `zoomed_dm`, TBD); exceptions: 
  - `dm_from_src`, but I don't think that's a problem.
  - `dm_copy_to`: first argument is `dest`: syntax known from `copy_to`, so might be ok
  - `dm_learn_from_db`: only argument is `dest`; shall we change the name to e.g. `learn_dm_from_db`?
  - `dm_nycflights13`: first argument is `cycle`; we could change the name to e.g. `nycflights13_dm`?
2. `check_` for a function that returns an error if the check isn't passed. Actually for all functions with this prefix, the first argument is a table and the 2nd is a column name (or ellipsis, but column names are expected)

We don't seem to have other rules for prefixes. Should we have one that expects a `zoomed_dm` as the first argument? Should we have one that works for both `zoomed_dm` and `tibble`? Maybe one that only works for `tibble`?

TSchiefer:
BTW, only function, that expects a `dm` as the first argument, but doesn't have a prefix `dm_` is `validate_dm()`. Probably fine though(?)

krlmlr:
What about suffixes?

TSchiefer:
Bit of patience please, WIP ;)

TSchiefer:
Our rules for suffixes:

1. `_to_tbl`: if the result is a `tibble` and the first argument is a `dm`; exceptions:
  - `dm_zoom_to_tbl()`: here the result is something similar to a `tibble`, so it's probably fine
  - functions that produce a table with summarized metainformation about the `dm`, like:
    - `dm_enum_(p,f)k_candidates()`; see 2.; should be alright
    - `dm_check_constraints()`: seems good to me.
2. `candidates`: if key-candidates are to be calculated and presented in `tibble`-form
3. `_(p,f)k(s)`: relates to primary/foreign key
4. `_tbl`: add/remove/select/rename a `tibble` to/from/within a `dm`; exceptions:
  - `dm_(insert,update)_zoomed_tbl()`: should be fine; see 6.
5. `_colors`: get info about or set the colors for the visualization of the `dm`
6. `zoomed_tbl`: insert or update zoomed table; `dm_zoom_out` is a bit of an exception here; possibly we could rename it to `dm_discard_zoomed_tbl()`(?)

krlmlr:
Thanks. Let's work on reducing exceptions. Also, which functions are not covered by the rules (yet)?

TSchiefer:
Not covered by the prefix rules:

- decompose_table
- reunite_parent_child
- reunite_parent_child_from_list
- dm_copy_to
- dm_learn_from_db
- enum_fk_candidates
- enum_pk_candidates
- validate_dm

Not really covered but probably ok are:
- new_dm
- as_dm
- is_dm
- dm

Not covered by the suffix rules:

- dm
- dm_from_src
- new_dm
- as_dm
- is_dm
- validate_dm
- dm_get_src
- dm_get_con
- dm_get_tables
- dm_get_filter
- dm_paste
- dm_is_referenced
- dm_get_referencing_tables
- dm_draw
- dm_zoom_out
- dm_disambiguate_cols
- dm_rename
- dm_select
- dm_filter
- dm_apply_filters
- dm_copy_to
- dm_learn_from_db
- decompose_table
- reunite_parent_child
- reunite_parent_child_from_list
- dm_check_constraints
- check_key
- check_if_subset
- check_set_equality
- check_cardinality
- check_cardinality_0_n
- check_cardinality_1_n
- check_cardinality_0_1
- check_cardinality_1_1
- dm_nycflights13
- dm_nrow
- browse_docs

Apart from the already mentioned ones I wouldn't necessarily see much need for changing suffixes

TSchiefer:
Suggestions for reducing exceptions:
1. `dm_learn_from_db`-> `learn_dm_from_db`
1. `dm_nycflights13` -> `nycflights13_dm`
1. `dm_zoom_out` -> `dm_discard_zoomed_tbl`

TSchiefer:
The following two functions could also be renamed:

1. `check_cardinality` -> `assess_cardinality`
1. `dm_check_constraints` -> `dm_assess_constraints`

TSchiefer:
or maybe `examine` instead of `assess`?

krlmlr:
`dm_apply_filters_to_tbl()` looks a bit odd too.

krlmlr:
`dm_` as prefix for functions that construct a dm (if the first argument isn't a dm)?

`_dm` suffix for validators, checkers and coercers.

Let's rename/merge:

- [ ] `dm_learn_from_db`-> `dm_from_src`
- [ ] `check_` -> `examine_` for cardinality and constraints
- [ ] `dm_copy_to` -> `copy_to` (will become a method)
- [ ] `*_zoomed_tbl` -> `*_zoomed`
- [ ] `dm_zoom_out` -> `dm_discard_zoomed`

Let's also describe our nomenclature in a vignette.

TSchiefer:
Sounds good about the new names.

About `dm_apply_filters_to_tbl()`: At first I also thought it's strange, but then again, you put in a `dm` and get a `tbl`, so both prefix (`dm_`) and suffix (`_to_tbl`) rules are fulfilled.

TSchiefer:
One more question for me is if we should forbid everything on a `zoomed_dm` that works on a `dm` and/or vice versa. And if we do not forbid this generally I would say we should once more think about it case by case.

NAMESPACE
```diff
@@ -91,7 +91,7 @@ export(cdm_get_all_pks)
 export(cdm_get_available_colors)
 export(cdm_get_colors)
 export(cdm_get_con)
-export(cdm_get_filter)
+export(cdm_get_filters)
```
krlmlr: `cdm_` functions remain unchanged.

R/dm.R
```diff
@@ -359,20 +359,20 @@ dm_get_data_model_fks <- function(x) {
...
+#' @return For `dm_get_filters()`: A tibble with columns:
 #'
 #' - "table": table that was filtered,
 #' - "filter": the filter expression,
 #' - "zoomed": logical, does the filter condition relate to the zoomed table.
 #'
 #' @export
-dm_get_filter <- function(x) {
+dm_get_filters <- function(x) {
   # FIXME: Obliterate
```
krlmlr: Let's remove this comment.

```suggestion
```

R/zzx-deprecated.R
```diff
@@ -16,7 +16,7 @@ cdm_get_tables <- new_cdm_forward(dm_get_tables)
 #' @rdname deprecated
 #' @keywords internal
 #' @export
-cdm_get_filter <- new_cdm_forward(dm_get_filter)
+cdm_get_filters <- new_cdm_forward(dm_get_filters)
```
krlmlr:
```suggestion
cdm_get_filter <- new_cdm_forward(dm_get_filters)
```

pkgdown/_pkgdown.yml
krlmlr: Agree to be stricter: functions/methods that apply to a zoomed dm don't have a prefix; functions work either for a zoomed, or for a non-zoomed dm.

pkgdown/_pkgdown.yml
krlmlr: Yes, let's get rid of it.

krlmlr:
Can you please copy the naming logic from this discussion into a small `.Rmd` file and add to this pull request?

pkgdown/_pkgdown.yml
TSchiefer: outsourced this comment to issue #230

pkgdown/_pkgdown.yml
TSchiefer: outsourced the point "Can you please collect all non-dm functions in a separate sub-topic?" into issue #231.

outsourced "Agree to be stricter: functions/methods that apply to a zoomed dm don't have a prefix; functions work either for a zoomed, or for a non-zoomed dm." into issue #232

TSchiefer:
Should we also rename `check_key()`, `check_if_subset()`, `check_set_equality()` to `examine_key()`, `examine_if_subset()`, `examine_set_equality()`? 

similar to: `check_` -> `examine_` for cardinality and constraints

I would probably do it, since it reduces the number of different prefixes.

TSchiefer:
I overwrote the method `dbplyr::copy_to.src_sql()` with a {dm}-method, so that it forwards to the {dbplyr}-method when `df` is a tbl of some sort. 

I then needed to adapt the vignettes, and they do work fine when I knit them manually, but in R CMD check there are still errors.

For now I do not check the box "`dm_copy_to` -> `copy_to` (will become a method)"

In other news: I also changed `check_key()`, `check_if_subset()`, `check_set_equality()` to `examine_key()`, `examine_if_subset()`, `examine_set_equality()`. All in one commit, so it can be reversed with minimal effort, should it be necessary.

TSchiefer:
Also: this is the Rmd with the function naming logic: https://github.com/krlmlr/dm/blob/f-pkgfown-ref-upd/vignettes/wip/dm-function-naming-logic.Rmd

R/dm.R
```diff
@@ -659,7 +673,44 @@ compute.zoomed_dm <- function(x, ...) {
 src_tbls.dm <- function(x) {
   # The x argument here is a dm object
   dm <- x
-  names(dm_get_tables(dm))
+  # FIXME: should this be forbidden for `zoomed_dm`? and should we not just take the names from `def`?
```
```suggestion
  check_not_zoomed(x)
```

and move before the assignment. `names(...)` is fine here.

R/dm.R
```diff
@@ -659,7 +673,44 @@ compute.zoomed_dm <- function(x, ...) {
...
+  names(dm_get_tables_impl(dm))
+}
+
+#' @export
+copy_to.src_sql <- function(dest,
```
krlmlr: This turns out to be a bad idea, because this method dispatches only on `dest` and currently doesn't use double dispatch. I'm hesitant to request this feature in dplyr, because it feels rather exotic.

Let's export `dm_copy_to()` again and not override `copy_to()`.

R/error-helpers.R
```diff
@@ -412,7 +412,7 @@ abort_is_not_dm <- function(obj_class) {
 }
 
 error_is_not_dm <- function(obj_class) {
-  glue("Required class `dm` but instead is {tick(obj_class)}")
+  glue("Required class `dm` but instead is {paste(tick(obj_class), collapse = ', ')}")
```
krlmlr:
```suggestion
  glue("Required class `dm` but instead is {format_classes(obj_class)}")
```

with a suitable implementation.

R/error-helpers.R
```diff
@@ -433,8 +433,8 @@ abort_no_zoom_allowed <- function() {
 
 error_no_zoom_allowed <- function() {
```
krlmlr: Can we remove this error and use `error_only_possible_wo_zoom` instead?

R/error-helpers.R
```diff
@@ -473,8 +473,8 @@ abort_only_possible_wo_zoom <- function(fun_name) {
 
 error_only_possible_wo_zoom <- function(fun_name) {
   glue(
-    "You cannot call `{fun_name}()` on a `zoomed_dm`. Consider using one of `dm_update_zoomed_tbl()`, ",
-    "`dm_insert_zoomed_tbl()` or `dm_zoom_out()` first."
+    "You cannot call `{fun_name}()` on a `zoomed_dm`. Consider using one of `dm_update_zoomed()`, ",
+    "`dm_insert_zoomed()` or `dm_zoom_out()` first."
```
krlmlr:
```suggestion
    "`dm_insert_zoomed()` or `dm_discard_zoomed()` first."
```

R/error-helpers.R
```diff
@@ -444,14 +444,14 @@ abort_w_message <- function(msg) {
...
-# no table zoomed, but 'dm_insert_zoomed_tbl()' called ---------------------------------
+# no table zoomed, but 'dm_insert_zoomed()' called ---------------------------------
 
 abort_no_table_zoomed <- function() {
   abort(error_no_table_zoomed(), .subclass = dm_error_full("no_table_zoomed"))
 }
 
 error_no_table_zoomed <- function() {
-  "`dm_insert_zoomed_tbl()` only works for zoomed `dm`"
+  "`dm_insert_zoomed()` only works for zoomed `dm`"
```
krlmlr:
```suggestion
  "`dm_insert_zoomed()` only works for zoomed `dm`."
```

Here and everywhere.

R/filter-helpers.R
```diff
@@ -13,19 +13,20 @@
 #'   dm_nrow()
 #' @export
 dm_nrow <- function(dm) {
+  check_not_zoomed(dm)
   # FIXME: with "direct" filter maybe no check necessary: but do we want to issue
   # a message in case the filters haven't been applied yet?
   check_no_filter(dm)
-  map_dbl(dm_get_tables(dm), ~ as.numeric(pull(collect(count(.)))))
+  map_int(dm_get_tables_impl(dm), ~ as.integer(pull(collect(count(.)))))
```
krlmlr: Protects against > 2³¹ rows.

```suggestion
  map_dbl(dm_get_tables_impl(dm), ~ as.numeric(pull(collect(count(.)))))
```

R/zoom.R
```diff
@@ -102,10 +102,10 @@ get_zoomed_tbl <- function(dm) {
 #' @param new_tbl_name Name of the new table.
 #' @inheritParams vctrs::vec_as_names
 #'
-#' @return For `dm_insert_zoomed_tbl()`, `dm_update_zoomed_tbl()` and `dm_zoomed_out()`: A `dm` object.
+#' @return For `dm_insert_zoomed()`, `dm_update_zoomed()` and `dm_zoomed_out()`: A `dm` object.
```
krlmlr:
```suggestion
#' @return For `dm_insert_zoomed()`, `dm_update_zoomed()` and `dm_discard_zoomed()`: A `dm` object.
```

Here and everywhere.

demo/basel.R
```diff
@@ -351,7 +351,7 @@ dm_with_summary %>%
 dm_flights %>%
   dm_zoom_to_tbl(flights) %>%
   count(carrier, origin) %>%
-  dm_insert_zoomed_tbl(origin_carrier_count) %>%
+  dm_insert_zoomed(origin_carrier_count) %>%
```
krlmlr:
```suggestion
  dm_insert_zoomed("origin_carrier_count") %>%
```

New objects should be quoted by the caller. Maybe add a warning in `dm_insert_zoomed()` (in a new issue/PR)?

pkgdown/_pkgdown.yml
```diff
@@ -43,26 +56,35 @@ reference:
   - dm_get_colors
   - dm_get_available_colors
 
+- title: Zooming to a table of a `dm`
+  contents:
+  - dm_zoom_to_tbl
+  - dm_insert_zoomed_tbl
```
krlmlr:
```suggestion
  - dm_insert_zoomed
```

?

pkgdown/_pkgdown.yml
```diff
@@ -43,26 +56,35 @@ reference:
   - dm_get_colors
   - dm_get_available_colors
 
+- title: Zooming to a table of a `dm`
+  contents:
+  - dm_zoom_to_tbl
```
krlmlr: Maybe:

```suggestion
  - dm_zoom_to
```

?

vignettes/wip/dm-function-naming-logic.Rmd
```diff
@@ -0,0 +1,43 @@
...
+<!-- FIXME: Vignette needs to be reviewed and integrated (https://github.com/krlmlr/dm/issues/185) -->
+
+## Prefixes
+
+1. `dm_`: There are two cases for this prefix:
+    - The function takes a `dm` object as a first argument
+    - The function constructs a `dm` object from other objects
+1. `examine_`: There are two cases for this prefix:
+    - The function throws an error if the respective check is not passed and otherwise returns the first argument invisibly.
+    - The function executes several checks and returns the results in tibble-form.
```
krlmlr: Can we use another prefix here?

vignettes/wip/dm-function-naming-logic.Rmd
```diff
@@ -0,0 +1,43 @@
...
+
+In general the functions are implemented in such a way, that no function can handle both `dm` classes (`dm` and `zoomed_dm`) as an argument.
+A `zoomed_dm` behaves in most situations like a table and there are mostly no explicit functions available for this class, but rather methods (see vignette ["Zooming and manipulating tables"]().
+<!-- FIXME: Vignette needs to be reviewed and integrated (https://github.com/krlmlr/dm/issues/185) -->
+
+## Prefixes
+
+1. `dm_`: There are two cases for this prefix:
+    - The function takes a `dm` object as a first argument
+    - The function constructs a `dm` object from other objects
```
krlmlr: Which functions do this, short of `dm()`?

krlmlr:
Thanks!

vignettes/wip/dm-function-naming-logic.Rmd
TSchiefer: Resolved

pkgdown/_pkgdown.yml
TSchiefer: ok, changed the name from `dm_zoom_to_tbl` to `dm_zoom_to`

vignettes/wip/dm-function-naming-logic.Rmd
TSchiefer: `dm_from_src()`, `dm_nycflights13()` and `dm_learn_from_db()` (didn't we consider changing its name to `learn_dm_from_db()`?)

PR #228, opened 2020-01-07
avoid warnings in R CMD check with dev pkg versions

TSchiefer:
I still had an error in `dm_learn_from_db()` related to stricter {vctrs} rules in its dev version. We need to revisit this function's implementation anyway, but solving this problem seemed beneficial.

In addition there appeared a few warnings in the tests when using dev versions of {tibble}, {rlang}


TSchiefer:
BTW, it looks like the dev version of {tibble} (`2.99.99.9012`) has a dependency of a newer {vctrs} version than is written in its DESCRIPTION file (it doesn't work with version `0.2.1`, but it works with the current dev version)

R/learn.R
```diff
@@ -192,16 +192,15 @@ legacy_new_dm <- function(tables, data_model) {
...
-  # Legacy compatibility
-  pks$column <- as.list(pks$column)
-
   pks <-
     pks %>%
-    nest(pks = -table)
+    # Legacy compatibility
+    mutate(column = vctrs::vec_cast(column, list())) %>%
+    nest(pks = -table) %>%
+    mutate(pks = vctrs::as_list_of(pks))
```
krlmlr: This won't be necessary with https://github.com/tidyverse/tidyr/pull/847 I think.

R/learn.R
```diff
@@ -217,6 +216,7 @@ legacy_new_dm <- function(tables, data_model) {
     fks %>%
     select(-ref_col) %>%
     nest(fks = -ref) %>%
+    mutate(fks = vctrs::as_list_of(fks)) %>%
```
krlmlr: Same.

tests/testthat/test-dm-table-surgery.R
```diff
@@ -69,7 +69,12 @@ test_that("table surgery functions fail in the expected ways?", {
     data_ts_src,
     ~ expect_error(
       decompose_table(., aex_id, a, e, x),
-      "."
+      # FIXME: error class might be subject to future change?
```
krlmlr: - Don't need a FIXME here -- if the test fails it will be obvious why
- Perhaps: `expect_error(..., ".", class = if (packageVersion(...)...) "tidyselect_...")` with suitable wrapping

R/learn.R
krlmlr: ...but it's going to take a while. Can you please create a small wrapper `nest_compat()` and add an issue?

tests/testthat/test-dm-table-surgery.R
TSchiefer: I'm not sure why we'd need the `"."`, since the default case `regexp = NULL` should cover this.

tests/testthat/test-dm-table-surgery.R
```diff
@@ -69,7 +69,7 @@ test_that("table surgery functions fail in the expected ways?", {
     data_ts_src,
     ~ expect_error(
       decompose_table(., aex_id, a, e, x),
-      "."
+      class = if_pkg_version("tidyselect", "0.2.99.9000", "tidyselect_error_subscript_oob_name", "simpleError")
```
krlmlr:
```suggestion
      class = if_pkg_version("tidyselect", "0.2.99.9000", "tidyselect_error_subscript_oob_name", NULL)
```

The dot is indeed unnecessary here.

R/learn.R
```diff
@@ -248,3 +240,12 @@ legacy_new_dm <- function(tables, data_model) {
...
   new_dm3(def)
 }
+
+nest_compat <- function(.data, ...) {
+  # `...` has to be name-variable pair (see `?nest()`) of length 1
+  quos <- enquos(...)
+  stopifnot(length(quos) == 1)
+  new_col <- names(quos)
+  nest(.data, ...) %>%
+    mutate(!!new_col := vctrs::as_list_of(!!sym(new_col)))
```
krlmlr: A bit simpler:

```suggestion
    mutate_at(vars(new_col), vctrs::as_list_of)
```

R/learn.R
```diff
@@ -192,16 +191,14 @@ legacy_new_dm <- function(tables, data_model) {
...
   key_tracker_zoom <- new_key_tracker_zoom()
 
-  # Legacy compatibility
-  pks$column <- as.list(pks$column)
-
   pks <-
     pks %>%
-    nest(pks = -table)
+    # Legacy compatibility
+    mutate(column = vctrs::vec_cast(column, list())) %>%
```
krlmlr: Do we still need this?

R/utils.R
```diff
@@ -32,3 +32,7 @@ tick_if_needed <- function(x) {
 is_syntactic <- function(x) {
   x == make.names(x)
 }
+
+if_pkg_version <- function(pkg, min_version, if_true, if_false) {
```
krlmlr:
```suggestion
if_pkg_version <- function(pkg, min_version, if_true, if_false = NULL) {
```

R/learn.R
TSchiefer: yes, I believe so, it's the difference of having a list of characters and a list of lists of characters in the `pks` column of the resulting `def`

krlmlr:
Thanks!

PR #235, opened 2020-01-14
revert dm::copy_to.src_sql()

TSchiefer:
- method `dm::copy_to.src_sql()` doesn't make much sense
- added internal function `src_tbls_impl()`, since we make use of `src_tbls()` frequently and since our stricter approach this throws errors when used on `zoomed_dm`


R/dm.R
```diff
@@ -677,40 +677,11 @@ src_tbls.dm <- function(x) {
...
-      indexes = indexes, unique_indexes = unique_indexes,
-      set_key_constraints = set_key_constraints, unique_table_names = unique_table_names,
-      table_names = table_names,
-      temporary = temporary
-    )
-  } else {
-    abort_either_dm_of_df(class(df))
-  }
+# this is not exported, but we use it a lot in our own code
+src_tbls.zoomed_dm <- function(x) {
```
```suggestion
src_tbls_impl <- function(x) {
```

R/db-interface.R
```diff
@@ -38,6 +38,7 @@
 #'   as_dm(list(iris = iris)),
 #'   set_key_constraints = FALSE
 #' )
+#' @export
 dm_copy_to <- function(dest, dm, ...,
```
krlmlr:
```suggestion
copy_dm_to <- function(dest, dm, ...,
```

with forwarder?

krlmlr:
Thanks!

PR #242, opened 2020-01-16
WIP: list for PK and FK cols

### LINKED ISSUE #239: Prepare dm_examine_constaints() for compound keys
Only for the interface, let's not change internals for now.

Should return `vctrs::list_of()`, create with:

```r
new_keys <- function(x) {
  vctrs::new_list_of(x, character())
}
```

- [x] `dm_examine_constraints()` should return a list in the `column` column. Can we keep `dm_check_constraints()` as is?
- [x] `dm_get_all_?ks()`
- [x] `dm_enum_?k_candidates()`
- [x] `enum_?k_candidates()`
- [x] `dm_get_pk()` et al.: Adapt documentation to describe that multiple key columns may be possible, perhaps a short vignette?

TSchiefer:
As it turns out, the assumption of single key columns is rather deeply rooted in the code. So far the function `dm_examine_constraints()` works as it should, but tests fail, since helper functions have been adapted, which now produce the wrong type for key-column -info-tibbles.

TBC

closes #239

### DIFF
--- a/R/db-helpers.R
+++ b/R/db-helpers.R
@@ -15,7 +15,6 @@ systime_convenient <- function() {
 
 # Internal copy helper functions
 build_copy_data <- nse(function(dm, dest, table_names, unique_table_names) {
-
   source <-
     dm %>%
     dm_apply_filters() %>%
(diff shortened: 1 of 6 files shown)


TSchiefer:
replaced by #245

PR #261, opened 2020-02-04
dm_paste() produces code for setting colors

### LINKED ISSUE #218: 'dm_paste()' should also produce code for setting colors
maybe optionally?

TSchiefer:
- `dm_paste()` has boolean argument `color`

Currently the tests start failing when I replace
```r
cat(glue_collapse(c(code, code_pks, code_fks), sep = " %>%\n"), "\n")
```
with
```r
cli::cli_code(glue_collapse(c(code, code_pks, code_fks), sep = " %>%\n"))
```
If I run one expectation manually, it works, but the problem is that when using `devtools::test()` or `testthis:::test_this()` it seems that no output can be detected from `dm_paste()`

fixes #218


R/paste.R
```diff
@@ -4,6 +4,8 @@
 #'
 #' @inheritParams dm_add_pk
 #' @param select Boolean, default `FALSE`. If `TRUE` will try to produce code for reducing to necessary columns.
+#' @param color Boolean, default `TRUE`. If `TRUE` will create the code for reproducing the original colors for
+#' the visualization of the tables.
```
krlmlr:
```suggestion
#'    visualizing the tables.
```

R/paste.R
```diff
@@ -23,7 +25,7 @@
 #'
 #' dm_nycflights13() %>%
 #'   dm_paste(select = TRUE)
-dm_paste <- function(dm, select = FALSE, tab_width = 2) {
+dm_paste <- function(dm, select = FALSE, color = TRUE, tab_width = 2) {
```
krlmlr:
```suggestion
dm_paste <- function(dm, select = FALSE, tab_width = 2, color = TRUE) {
```

For better backward compatibility. Perhaps we can always generate colors and (if ever needed) provide `dm_remove_colors()` ?

R/paste.R
```diff
@@ -4,6 +4,8 @@
 #'
 #' @inheritParams dm_add_pk
 #' @param select Boolean, default `FALSE`. If `TRUE` will try to produce code for reducing to necessary columns.
```
krlmlr:
```suggestion
#' @param select Boolean, default `FALSE`. If `TRUE` will produce code for reducing to necessary columns.
```

What can go wrong?

tests/testthat/test-cdm-deprecated.R
```diff
@@ -258,7 +258,7 @@ test_that("cdm_nycflights13() works", {
 test_that("cdm_paste() works", {
   withr::local_options(c(lifecycle_verbosity = "quiet"))
   expect_output(
-    cdm_paste(dm_for_filter, FALSE, 4),
+    cdm_paste(dm_for_filter, FALSE, TRUE, 4),
```
krlmlr: The `cdm_paste()` interface should not be affected.

R/paste.R
TSchiefer: What would you think of leaving the color optional for `dm_paste()`, but anyway provide `dm_remove_colors()` (or maybe `dm_rm_colors()`?)

TSchiefer:
In addition I now changed the output of `dm_get_colors()` from `NA` to `default` in case no color or `default` is set for a table, since:
1. it is nicer for the user
1. we can do this: 
``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter
dm_set_colors(
  dm_nycflights13(color = TRUE), 
  !!!dm_get_colors(dm_nycflights13(color = FALSE))
  ) %>% 
  dm_draw()
```



<sup>Created on 2020-02-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

tests/testthat/test-dm-paste.R
```diff
@@ -52,4 +52,18 @@ test_that("generating code for creation of existing 'dm' works", {
     ),
     fixed = TRUE
   )
+
+  # produce code with colors
+  expect_output(
```
krlmlr: Can you please use `verify_output()` ?

R/paste.R
```diff
@@ -3,7 +3,9 @@
 #' `dm_paste` takes an existing `dm` and produces the code necessary for its creation
 #'
 #' @inheritParams dm_add_pk
-#' @param select Boolean, default `FALSE`. If `TRUE` will try to produce code for reducing to necessary columns.
+#' @param select Boolean, default `FALSE`. If `TRUE` will produce code for reducing to necessary columns.
+#' @param color Boolean, default `TRUE`. If `TRUE` will create the code for reproducing the original colors for
```
krlmlr:
```suggestion
```

R/paste.R
```diff
@@ -23,7 +25,7 @@
 #'
 #' dm_nycflights13() %>%
 #'   dm_paste(select = TRUE)
-dm_paste <- function(dm, select = FALSE, tab_width = 2) {
+dm_paste <- function(dm, select = FALSE, tab_width = 2, color = TRUE) {
```
krlmlr:
```suggestion
dm_paste <- function(dm, select = FALSE, tab_width = 2) {
  # FIXME: Expose as argument?
  color <- TRUE

```

tests/testthat/test-dm-paste.R
TSchiefer: weirdly the last example leads to an error when used in `verify_output()`, even though it works fine otherwise. It would be nice if you could have a look.

TSchiefer: Relevant commit is 374d7c6

TSchiefer:
Trying now `cli::cli_code(glue_collapse(c(code, code_pks, code_fks), sep = " %>%\n"))` for producing the output of `dm_paste()` since that works together with `verify_output()` (9d34652). Please revert if that's a bad idea.

krlmlr:
Thanks!

PR #271, opened 2020-02-06
better zoomed tracking

### LINKED ISSUE #266: Zooming and unzooming is expensive
``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter

bench::mark(dm_select(dm_nycflights13(), flights))
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 1 x 6
#>   expression                              min median `itr/sec` mem_alloc
#>   <bch:expr>                            <bch> <bch:>     <dbl> <bch:byt>
#> 1 dm_select(dm_nycflights13(), flights) 203ms  227ms      4.49     147MB
#> # … with 1 more variable: `gc/sec` <dbl>
```

<sup>Created on 2020-02-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Profiling shows that >90% of the time is used by zooming and unzooming. 

Collection of primary and especially foreign keys is expensive. Do we really need to collect all keys beforehand and store them redundantly? Would it be sufficient to keep track of the valid columns instead?

- The tracker starts with `set_names(names(data))`; this is a map from old names to new names
- If a column is removed with `select(-...)`, or mutated, we remove it from the tracker
- If a column is renamed, we rename it in our tracker
- We also track if an operation that repeats rows is applied to our data -- if yes, the primary key is invalidated
- When unzooming, we collect the information and only update the primary keys if anything changed

What am I missing?

TSchiefer:
should be quicker eventually:

- all columns are tracked of zoomed_table
- columns are removed from tracker under certain circumstances (mutate, deselect, etc.)
- if all columns are still tracked all keys will remain:
	This feature is only available for `dm_update_zoomed()`, since we need to access the key columns of the original table specifically when using `dm_insert_zoomed()` in order to replicate the keys for the new table

closes #266

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -146,7 +146,7 @@ new_dm <- function(tables = list()) {
   data <- unname(tables)
   table <- names2(tables)
   zoom <- new_zoom()
-  key_tracker_zoom <- new_key_tracker_zoom()
+  col_tracker_zoom <- new_col_tracker_zoom()
 
   pks <-
     tibble(
@@ -172,7 +172,7 @@ new_dm <- function(tables = list()) {
     left_join(fks, by = "table") %>%
     left_join(filters, by = "table") %>%
     left_join(zoom, by = "table") %>%
-    left_join(key_tracker_zoom, by = "table")
+    left_join(col_tracker_zoom, by = "table")
 
   new_dm3(def)
 }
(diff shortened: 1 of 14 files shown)


krlmlr:
Thanks!

PR #272, opened 2020-02-07
WIP: Repair winbuilder checks

### LINKED ISSUE #265: Ensure winbuilder checks pass
```r
devtools::check_win_devel(email = "...@....com")
```

Last run: https://win-builder.r-project.org/3yrJZt2vvzFO/00check.log

[00check.log](https://github.com/krlmlr/dm/files/4161850/00check.log)

TSchiefer:
- speed up examples for `dm_filter()` et al.
- solve problem with URLs

closes #265


TSchiefer:
`devtools::check_win_devel(email = "...")` won't work for me currently:

part of the output is:
```
Hmm ... looks like a package
   Converting Rd files to LaTeX ....
   Creating pdf output from LaTeX ...
   Error in texi2dvi(file = file, pdf = TRUE, clean = clean, quiet = quiet,  : 
     Running 'texi2dvi' on 'Rd2.tex' failed.
   LaTeX errors:
   ! Package textcomp Error: Symbol \textasciigrave not provided by
   (textcomp)                font family zi4 in TS1 encoding.
   (textcomp)                Default family used instead.
   
   See the textcomp package documentation for explanation.
   ! Emergency stop.
    ...                                              
                                                     
   l.479 ... works only when tables of `dm` are on DB
                                                     
   !  ==> Fatal error occurred, no output PDF file produced!
```
This seems to come from a comment in the examples in "dm.R":
```r
#' # this works only when tables of `dm` are on DB
```
If I remove the backticks it works, but there are a lot of them. Can we avoid this problem in a different way?

TSchiefer:
> If I remove the backticks it works, but there are a lot of them. Can we avoid this problem in a different way?

Even though e.g. {dplyr} also doesn't use backticks in its examples (cf. `?dplyr::count`), so maybe we can replace them if they cause trouble(?)

TSchiefer:
I did replace the backticks in the examples now and did the winbuild-check. You can revert the backticks-commit if you like.

@krlmlr URLs are still a problem, examples are solved: https://win-builder.r-project.org/knwn47lqf4hO/00check.log

[00check.log](https://github.com/krlmlr/dm/files/4170599/00check.log)

It would be nice if you could let me know how to solve this (or just solve it :) )

krlmlr:
The URL problems mean that some of the links are bad. https://krlmlr.github.io/dm/articles/dm-introduction-relational-data-model.html gives a 404, use https://CRAN.R-project.org/package=dm instead of whatever the README tries to do to link to the dm package on CRAN, search for `]()` to find the empty link.

Also the Package Title Needs to Be in Title Case.

TSchiefer:
2 iterations later...

https://win-builder.r-project.org/243aSUG62dCC/00check.log

[00check.log](https://github.com/krlmlr/dm/files/4171157/00check.log)

Almost there, looks like the only problem now is that the email address doesn't match the person (just kidding, guess it's the umlaut. Is there a way to repair this though? Set encoding to WINDOWS-1252??)

TSchiefer:
> Is there a way to repair this though?

I tried replacing "ü" with "\u00FC". This is reverted though before sending it off.

TSchiefer:
Just wondering if it's really necessary to reduce the `flights`-table in `dm_nycflights13()` by default (as done in 7d4dc92023d18e5d2b497ce2fa8c7d0790425634). It might have been enough to reduce the amount of examples for `dm_filter()` et al. and minimize the times `dm_nycflights13()` is being called.

Among others it messes with examples in vignettes, where we talk about the resulting numbers related to `nycflights` tables. 

But on the other hand it also has advantages, not sure what's best.

krlmlr:
Thanks! We might need to work on the LaTeX installation to fix the backtick problem for the future.

PR #273, opened 2020-02-07
abolish nse

TSchiefer:
closes #268

PR #274, opened 2020-02-11
use quo select table

### LINKED ISSUE #257: Use quo_select_table()
and add `quo_rename_table()`, instead of `vars_select()`, `eval_select()` and `vars_rename()`.

TSchiefer:
- instead of `dm_try_tables()` use `quo_select_table()` or `quo_rename_table()`
- also replaced every `tidyselect::vars_select()` with new `tidyselect::eval_select()`
- no merge conflicts with #273

closes #257


R/add-tbl.R
```diff
@@ -32,19 +32,23 @@ dm_add_tbl <- function(dm, ..., repair = "unique", quiet = FALSE) {
...
   dm_add_tbl_impl(dm, new_tables, names_list$new_names)
 }
 
-repair_table_names <- function(old_names, new_names, repair = "check_unique", quiet = FALSE) {
-  all_names <- tryCatch(
-    vctrs::vec_as_names(c(old_names, new_names), repair = repair, quiet = quiet),
+repair_names_vec <- function(names, repair, quiet) {
+  tryCatch(
+    vctrs::vec_as_names(names, repair = repair, quiet = quiet),
     error = function(e) {
```
krlmlr:
```suggestion
    vctrs_error_names_must_be_unique = function(e) {
```

and get rid of the check in the body.

R/select-tbl.R
```diff
@@ -11,18 +11,20 @@
...
 #' @param ... One or more table names of the tables of the [`dm`] object.
 #' `tidyselect` is supported, see [`dplyr::select()`] for details on the semantics.
+#' @inheritParams vctrs::vec_as_names
 #'
 #' @examples
 #' dm_nycflights13() %>%
 #'   dm_select_tbl(airports, fl = flights)
 #' @export
-dm_select_tbl <- function(dm, ...) {
+dm_select_tbl <- function(dm, ..., repair = "check_unique", quiet = FALSE) {
```
krlmlr: I'd rather not add new arguments to this exported function now.

R/select-tbl.R
```diff
@@ -36,11 +38,12 @@ dm_select_tbl <- function(dm, ...) {
 #' dm_nycflights13() %>%
 #'   dm_rename_tbl(ap = airports, fl = flights)
 #' @export
-dm_rename_tbl <- function(dm, ...) {
+dm_rename_tbl <- function(dm, ..., repair = "check_unique", quiet = FALSE) {
```
krlmlr: Same here.

R/select-tbl.R
TSchiefer: Alright, I was wondering. But do you agree with throwing an error if duplicated table names would be produced?

TSchiefer:
We need `eval_select()` and `eval_rename()` at least once more (apart from in `quo_select_table()` and `quo_rename_table()`) for choosing other things than tables, e.g. in `select.zoomed_dm()` et al. for the column tracking.

We could:
1. provide functions `quo_select_idx()` and `quo_rename_idx()`
1. just use `tidyselect::eval_select()` and `eval_rename()`

I think 2. might be more useful, since the new functions in 1. would do pretty much the same as the {tidyselect}-functions, so I am not sure why we should provide new ones. (having said that, maybe `quo_rename_idx()` could be useful, since `eval_rename()` doesn't include the columns that aren't touched).

krlmlr:
I added a thin layer around tidyselect to do the following:

1. Distinguish between tables and columns
2. Accept a vector of names instead of a named vector
3. Support returning names (which we need everywhere for column tracking)

I think this is well worth the effort, because it removes a lot of redundancy and friction in the code that uses tidyselect.

Could you please take a look?

R/select-tbl.R
krlmlr: Good catch. I adapted our tidyselect wrapper to throw errors early.

krlmlr:
I have no idea why tests are failing now on Travis. Running winbuilder checks to verify.

TSchiefer:
Only the R 3.3 test is failing, and the problem is apparently, that `vctrs::vec_as_names(..., repair = "unique")` in this setup does for some reason not produce messages (it's called in our function `repair_names_vec()`).

When the R 3.3 test is made, is then maybe also an older version of {vctrs} used? 

Also don't understand it, and the same thing started happening on master (or rahter the vignette update branch, which changed nothing apart from the vignettes)

krlmlr:
Thanks!

Issue #278, opened 2020-02-13
Double-check code coverage


krlmlr:
Did we forget any important parts of the codebase in our tests?

Review https://codecov.io/gh/krlmlr/dm/tree/master/R, sort by lines not covered.

PR #279, opened 2020-02-13
support alpha colors

TSchiefer:
- so far we are just ignoring the alpha channel
- after this PR we'd store everything as hex with alpha

(realised this while working on #278, when I was writing tests for color-functions, so that paid off already ;) )


krlmlr:
Thanks!

#261 now has conflicts.

PR #293, opened 2020-03-03
repair and speed up CRAN version tests

TSchiefer:
- repaired tests in one commit
- tried removing redundant tests
- tried removing redundant setup code

test now consistently under 1min; 
one slow-ish test will be additionally skipped_on_cran.


tests/testthat/helper-expectations.R
```diff
@@ -18,9 +18,9 @@ expect_dm_error <- function(expr, class) {
 }
 
 expect_name_repair_message <- function(expr) {
-  # FIXME: Name repair should get a message
+  # Name repair did not get a message during some time in {vctrs}
   # https://github.com/r-lib/vctrs/issues/849
-  if (packageVersion("vctrs") >= "0.2.3") {
+  if (packageVersion("vctrs") >= "0.2.3" && packageVersion("vctrs") < "0.2.99.9006") {
```
krlmlr:
```suggestion
  if (packageVersion("vctrs") < "0.2.99.9006") {
```

tests/testthat/test-db-interface.R
```diff
@@ -19,11 +19,15 @@ test_that("copy_dm_to() copies data frames from databases", {
 })
 
 test_that("copy_dm_to() copies between sources", {
+  # speed things up on cran
+  skip_on_cran()
+
+  all_combos <- crossing(tibble(a = 1:3), tibble(b = 1:3))
```
krlmlr: Good catch!

tests/testthat/test-dm.R
```diff
@@ -271,7 +269,6 @@ test_that("`pull_tbl()`-methods work", {
 
   expect_dm_error(
     new_dm3(dm_for_filter %>%
-      dm_zoom_to(t1) %>%
       dm_get_def() %>%
       mutate(zoom = list(t1)), zoomed = TRUE) %>%
```
krlmlr: How does this work? Do we need instead:

```suggestion
      mutate(zoom = list(t1)), zoomed = TRUE) %>%
      new_dm3() %>%
```

tests/testthat/test-learn.R
```diff
@@ -70,6 +70,7 @@ test_that("Learning from Postgres works?", {
 })
 
 test_that("Learning from SQLite works (#288)?", {
+  skip("learning from SQLite NYI")
```
krlmlr: Can we backport this fix? I think it's e8133c7.

```suggestion
```

tests/testthat/test-zzx-deprecated.R
```diff
@@ -1,110 +1,95 @@
+withr::local_options(c(lifecycle_verbosity = "quiet"))
```
krlmlr: It's slightly safer to have them inside every `test_that()` call, these calls shouldn't take too much time. Also, `rlang::local_options()` is slightly faster.

Something else: all of these tests can be skipped on CRAN.

krlmlr: Scratch that. Let's keep one `local_options()` call at the top of the test file, it's safe enough and easier to read.

tests/testthat/test-dm.R
TSchiefer: The idea is to have multiple tables at once in the `zoom`-column and your suggestion does that, and looks slightly more transparent at the same time.

TSchiefer: Sorry, misread, `zoomed = TRUE` is an argument for `new_dm3()`. 
I will do it like this:
```r
dm_get_def(dm_for_filter) %>%
      mutate(zoom = list(t1)) %>%
      new_dm3(zoomed = TRUE) %>%
      pull_tbl()
```

krlmlr:
Thanks!

PR #303, opened 2020-04-02
Added copy_to and compute documentation

jawond:
@krlmlr Changes are at the bottom of the document. It will need some example code.


krlmlr:
Can you please move this content to #309?

krlmlr:
I see we're still missing example code here. Can you please keep the FIXME so I can add code later if necessary?

krlmlr:
Should this remain here? Do we need to synchronize with `dm-db.Rmd`?

jawond:
> Should this remain here? Do we need to synchronize with `dm-db.Rmd`?

This has been moved to dm-db.Rmd in my local dm-db branch, which is part of the larger document consolidation.

krlmlr:
Now part of howto-dm-db.Rmd.

PR #313, opened 2020-04-11
Don't iterate over sources in tests

### LINKED ISSUE #312: Simplify tests
The current testing infrastructure will be incompatible with upcoming dplyr/dbplyr releases, and we're using unexported functionality. Need to adapt.

Suggesting the following test design, replacing #76:

- All tests test only one source. If a second source is needed, we use a dedicated SQLite database.
- The data source to be tested will be defined externally, prior to starting the tests. All dm and table objects remain accessible through their names, but `dm_..._src` will vanish.

This achieves a number of things:

- Simpler tests, remove iteration over sources, easier to rerun
- Faster tests
- Can run integration tests on all sources on CI
- Easy to add new sources and to test externally

krlmlr:
Similar changes should be applied throughout all tests. We will add infrastructure to test all sources later.

@TSchiefer: Can you prioritize please?

Closes #312.


TSchiefer:
Is the plan to have `collect()`-calls in each test (one for the expression to test and one for the expectation) in case the tester chooses to test on a DB?
Or will that be handled somehow automatically?

krlmlr:
`collect()` is a no-op for data frames, this should be handled automatically by `expect_equivalent_dm()`. Can you point me to an example?

TSchiefer:
Currently one test looks like this:
```r
  expect_identical(
    dm_zoom_to(dm_for_filter, t3) %>%
      mutate(new_col = row_number() * 3) %>%
      pull_tbl() %>%
      collect(),
    mutate(t3, new_col = row_number() * 3)
  )
```
I guess both `dm_for_filter` as well as `t3` would in the future be on the same source. So it seems that therefore it would be explicitly *wrong* to use `collect()` in the first part of the `expect...()` function and not in the 2nd part.

Is that the right way to see it? In this case we can probably remove most of the `collect()` statements that are currently present in the tests.

krlmlr:
Can you use `expect_equivalent_dm()` here?

TSchiefer:
Doesn't seem like it: the comparison needs to be with a table. (we could create a `dm` from that table and select only this table in the other `dm` and compare the two, but that doesn't seem right)

But just generally: will all objects be on the same source in the new test setup? (apart from those few SQLite-objects)

TSchiefer:
I was thinking that we can even use this (since we `collect()` anyway):

```r
expect_identical_tbl <- function(tbl_1, tbl_2) {
  tbl_1_lcl <- collect(tbl_1)
  tbl_2_lcl <- collect(tbl_2)
  expect_identical(tbl_1_lcl, tbl_2_lcl)
}
```
That ok?

krlmlr:
Sure, if it makes our tests easy to read. If we need to harmonize the order of records, we can throw in an `arrange_all()` too.

TSchiefer:
Good idea, but `arrange_all()` errors in the presence of list columns. Even though we don't have them now, we will get those in the future with compound keys. And `arrange_if()` will be superseded in the new {dplyr}, I think. Bit of a pickle, how would you do it?

krlmlr:
`arrange_if()` will stay around for quite some time.

TSchiefer:
What about this test for subsetting a `zoomed_dm`:
```r
expect_identical(
    dm_zoom_to(dm_for_filter, t2)$c,
    pull(t2, c)
  )
```
If `dm_for_filter` is on a DB, `dm_zoom_to(dm_for_filter, t2)$c` is `NULL`. 

I'd say we have one `dm` object that is locally available unconditionally.

krlmlr:
We need to use operations that work on both database and locally in all tests:

```r
dm_zoom_to(dm_for_filter, t2) %>% pull(c)
```

If you want to test behavior for specific backends, please add a `skip_if_local_src()` or `skip_if_remote_src()`, currently a no-op.

tests/testthat/test-add-tbl.R
```diff
@@ -71,28 +67,16 @@ test_that("dm_add_tbl() works", {
...
-    data_1_src,
-    function(data_1) {
-      expect_equal(
-        dm_add_tbl(dm(), test = data_1)$test %>%
-          collect(),
-        data_1 %>%
-          collect()
-      )
-    }
+  expect_equal(
```
krlmlr, line 76: Use `expect_equivalent_tbl()` here?

Seems worthwhile to go over all tests once more and use new expectations and idioms we implement along the way.

tests/testthat/test-check-cardinalities.R
```diff
@@ -48,24 +40,18 @@ test_that("check_cardinality_...() functions are checking the cardinality correc
...
-      expect_dm_error(
-        check_cardinality_0_n(
-          parent_table = ..1,
-          pk_column = a,
-          child_table = ..2,
-          fk_column = a
-        ),
-        class = "not_subset_of"
+  # FIXME: Regarding PR #313: the known output changes, depending on if it's on a DB or locally
+  expect_known_output(
```
krlmlr, line 44: Can you please switch to `verify_output()` everywhere?

tests/testthat/test-dm.R
```diff
@@ -76,58 +74,49 @@ test_that("'copy_to.dm()' works", {
...
-          dm_get_def()
-        test <- map_chr(map(def$data, sql_render), as.character)
-        all(map_lgl(test, ~ !grepl("WHERE", .)))
-      })
-    }
-  )
+  # FIXME: Regarding PR #313: should this be tested on object `dm_for_filter` and the test
+  # skipped if it is a local dm?
+  def <-
+    dm_for_filter_sqlite %>%
```
krlmlr, line 94: `skip_if_local_src()` and

```suggestion
    dm_for_filter %>%
```

tests/testthat/test-dm.R
```diff
@@ -344,6 +303,7 @@ test_that("subsetting for dm/zoomed_dm", {
 test_that("methods for dm/zoomed_dm work", {
   expect_identical(length(dm_for_filter), 6L)
   expect_identical(length(dm_zoom_to(dm_for_filter, t2)), 3L)
```
krlmlr, line 305:
```suggestion
  expect_identical(ncol(dm_zoom_to(dm_for_filter, t2)), 3L)
```

TSchiefer:
Sounds good, thanks.

This test and related ones we will need to skip in that case, since it's specifically about subsetting a `zoomed_dm` with the use of `$`

tests/testthat/test-dm.R
```diff
@@ -370,10 +330,14 @@ test_that("dm_get_src() works", {
     class = "is_not_dm"
   )
 
-  walk2(
-    dm_for_filter_src,
-    active_srcs_class,
-    ~ expect_true(inherits(dm_get_src(.x), .y))
+  expect_identical(
+    class(dm_get_src(dm_for_filter)),
+    c("src_local", "src")
```
krlmlr, line 335:
```suggestion
    class(my_test_src)
```

with an appropriate `my_test_src` .

Perhaps even `expect_dm_src(dm_for_filter, my_test_src)` ?

tests/testthat/test-dm.R
```diff
@@ -370,10 +330,14 @@ test_that("dm_get_src() works", {
...
+  )
+
+  expect_identical(
+    class(dm_get_src(dm_for_filter_sqlite)),
```
krlmlr, line 339: Why does this test test for SQLite?

tests/testthat/test-dm.R
```diff
@@ -388,13 +352,9 @@ test_that("dm_get_con() works", {
...
-  dm_for_filter_src_red <- dm_for_filter_src[!(names(dm_for_filter_src) == "df")]
-
-  walk2(
-    dm_for_filter_src_red,
-    active_con_class,
-    ~ expect_true(inherits(dm_get_con(.x), .y))
+  expect_identical(
+    class(dm_get_con(dm_for_filter_sqlite)),
+    `attr<-`("SQLiteConnection", "package", "RSQLite")
   )
```
krlmlr, line 358:
```suggestion
  skip_if_local()
  expect_identical(
    dm_get_con(dm_for_filter),
    my_test_src$con
  )
```

tests/testthat/test-examine-constraints.R
```diff
@@ -27,32 +27,45 @@ test_that("`dm_examine_constraints()` works", {
...
+      kind = c("PK", "FK"),
+      columns = new_keys("key"),
+      ref_table = c(NA, "iris_1"),
+      is_key = TRUE,
+      problem = ""
+    ) %>%
+      new_dm_examine_constraints()
+  )
+
+  # FIXME: maybe always include this test on a DB, despite PR #313?
```
krlmlr, line 43: Please only include tests for one source, unless the test specifically tests two sources (e.g. `copy_dm_to()`).

tests/testthat/test-examine-constraints.R
```diff
@@ -27,32 +27,45 @@ test_that("`dm_examine_constraints()` works", {
...
-        nyc_check
-      )
-    }
+  expect_identical(
+    dm_examine_constraints(dm_nycflights_small) %>%
+      mutate(problem = if_else(problem == "", "", "<reason>")),
+    nyc_check
+  )
+
+  # FIXME: maybe always include this test on a DB, despite PR #313?
```
krlmlr, line 64: Please only include tests for one source.

tests/testthat/test-filter-dm.R
```diff
@@ -97,42 +97,46 @@ test_that("we get filtered/unfiltered tables with respective funs", {
...
 })
 
+test_that("dm_filter() works on different srcs", {
+  expect_identical(
+    dm_filter(dm_for_filter, t1, a < 8, a > 3) %>% dm_apply_filters() %>% dm_get_tables(),
+    output_1
+  )
 
-
+  # FIXME: maybe always include this test on a DB, despite PR #313?
```
krlmlr, line 106: Why?

tests/testthat/test-foreign-keys.R
```diff
@@ -1,158 +1,151 @@
...
+  )
+
+  expect_true(
+    dm_add_pk(dm_test_obj, dm_table_4, c) %>%
+      dm_add_fk(dm_table_1, a, dm_table_4) %>%
+      dm_has_fk(dm_table_1, dm_table_4)
+  )
+
+  expect_true(
+    dm_add_pk(dm_test_obj_sqlite, dm_table_4, c) %>%
```
krlmlr, line 19: Why SQLite here?

tests/testthat/test-foreign-keys.R
```diff
@@ -168,16 +161,22 @@ test_that("dm_enum_fk_candidates() works as intended?", {
...
+    dm_test_obj %>%
+      dm_add_pk(dm_table_4, c) %>%
+      dm_enum_fk_candidates(dm_table_1, dm_table_4) %>%
+      mutate(why = if_else(why != "", "<reason>", "")) %>%
+      collect(),
+    tbl_fk_candidates_t1_t4
+  )
+
+  expect_identical(
+    dm_test_obj_sqlite %>%
```
krlmlr, line 174: ?

tests/testthat/test-filter-dm.R
TSchiefer: Idea was, that the things happening to the `dm` object (or maybe rather, how they are happening) depend to a certain extent on the `src` it is located. These differences are handled by {dbplyr}, so there is probably no real need to force the testing to be executed on a DB. I am gonna be more conservative from now on.

tests/testthat/test-check-cardinalities.R
TSchiefer: Sure, but that doesn't solve the problem at hand: e.g. for SQLite the line `# A tibble: 1 x 2` is missing with comparison to the local `src`.

tests/testthat/test-dm.R
TSchiefer: sure, but we are testing the method `length.zoomed_dm()` here. So maybe just `skip_if_remote_src()`?

tests/testthat/test-dm.R
TSchiefer: Ok, I'll use a dummy local `my_test_src` for now

tests/testthat/test-check-cardinalities.R
krlmlr: Good point. Can you please introduce `verify_output_src()` ?

tests/testthat/test-dm.R
krlmlr: Yes, and also add a FIXME to remind us that we need a test for `ncol()` if we don't have it yet.

tests/testthat/test-check-cardinalities.R
TSchiefer: I just saw that we introduced a workaround at some point for trimming the `tbl_sql` print-output like so:
```r
if (is_this_a_test()) {
  tbl_sql <- dbplyr::tbl_sql
  tbl_sum.tbl_sql <- function(x, ...) c()
  vctrs::s3_register("tibble::tbl_sum", "tbl_sql", tbl_sum.tbl_sql)
}
```
Adding two more lines here solves the issue (as long as we don't have tibbles with more than 10 rows):
```r
if (is_this_a_test()) {
  tbl_sql <- dbplyr::tbl_sql
  tbl_sum.tbl_sql <- function(x, ...) c()
  vctrs::s3_register("tibble::tbl_sum", "tbl_sql", tbl_sum.tbl_sql)
  tbl_sum.tbl_df <- function(x, ...) c()
  vctrs::s3_register("tibble::tbl_sum", "tbl_df", tbl_sum.tbl_df)
}
```
the head of the output then looks the same and we don't need `verify_output_src()`.

Caveat: if we execute a test in the console with `testthis:::test_this()`, afterwards neither `tbl_sql` nor `tbl_df` will look like we are used to (for `tbl_sql` already the case). (maybe we can somehow restore the normal `tbl_sum`-methods after tests are done?)

Is that ok for you?

TSchiefer:
I have a question on learning from `my_test_src`: for testing learning from a DB we need to write permanent tables to the DB. In the new setup, letting the tester choose the `src`, we probably do not want to leave those tables on their DB afterwards. I couldn't find a {dbplyr}-function for dropping tables, translating to the respective SQL-dialect. Should I write the respective SQL-code for deleting the test-tables for all sorts of `src`-objects or is there a better solution?

TSchiefer:
- The only test-file that hasn't been updated and still relies on the old setup is "test-learn.R", see question above.
- I think what remains to be done is implementing the user chosen `my_test_src` (out of curiosity: will this be done by setting an environment variable or so?)
- skipping of remote-only tests if `my_test_src` is local and vice versa works already due to functions `skip_if_remote_src(src)` and `skip_if_local_src(src)`
- test-all-time down to <43 sec :)

tests/testthat/test-check-cardinalities.R
```diff
@@ -48,23 +40,16 @@ test_that("check_cardinality_...() functions are checking the cardinality correc
...
-    ~ expect_known_output(
-      expect_dm_error(
-        check_cardinality_0_n(
-          parent_table = ..1,
-          pk_column = a,
-          child_table = ..2,
-          fk_column = a
-        ),
-        class = "not_subset_of"
+  verify_output(
```
krlmlr, line 43: I'd use either `verify_output()` or `expect_error()`, not both.

tests/testthat/test-dm.R
```diff
@@ -144,31 +133,21 @@ test_that("some methods/functions for `zoomed_dm` work", {
...
-    new_dm3(dm_get_def(dm_for_filter) %>%
-      mutate(data = if_else(table == "t1", list(dm_for_filter_src$postgres$t1), data))) %>%
-      validate_dm(),
-    "dm_invalid"
-  )
-})
-
 test_that("validator speaks up (sqlite)", {
-  skip_if_not("sqlite" %in% src_names)
+  # FIXME: PR #313: will this test fail, if chosen `src` is SQLite?
```
krlmlr, line 147: Is this test valid only on the database? I'll tweak.

tests/testthat/out/check-set-equality-1a-2a.txt
```diff
@@ -0,0 +1,5 @@
+> expect_dm_error(check_set_equality(data_1, a, data_2, a), class = "sets_not_equal")
```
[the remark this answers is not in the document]
TSchiefer: It is an error.
We provide the tibble in addition for information what went wrong.

tests/testthat/test-check-cardinalities.R
TSchiefer: Same reason as above, error plus info-tibble; but yeah, unclear if we need to check both

tests/testthat/test-dm.R
TSchiefer: we want to test, if the validator can detect it, when tables are on different `src`s.
That's why we need sqlite-version here.
FIXME here is cause I am wondering what happens if the user choses SQLite as test source; in this case all of the tables might be on the same `src`

krlmlr: Yes, it will be a different connection.

krlmlr:
Thanks!

PR #314, opened 2020-04-11
dm_paste() with table definition

### LINKED ISSUE #301: Prototype of a dm
Conceptually similar to `vec_ptype()` -- zero-row *data frames* with correct column names and keys.

### LINKED ISSUE #302: dm_paste() should optionally also include definition of all tables
so that it becomes fully self-contained and can recreate the prototype of a dm (#301).

`dm_paste()` also needs to be able to write to a file.

Output options should be a character vector, deprecate `select`.

krlmlr:
Closes #302.


R/paste.R
```diff
@@ -3,78 +3,214 @@
...
 #' @param tab_width Indentation width for code from the second line onwards
+#' @param options Formatting options. A character vector containing some of:
+#'   - `"tables"`: [tibble()] calls for empty table definitions
+#'     derived from [dm_ptype()], overrides `"select"`.
+#'   - `"select"`: [dm_select()] statements for columns that are part
+#'     of the dm.
+#'   - `"pks"`: [dm_add_pk()] statements for primary keys.
+#'   - `"keys"`: [dm_add_fk()] statements for foreign keys, implies `"pks"`.
+#'   - `"color"`: [dm_set_colors()] statements to set color.
+#'   - `"all"`: All options above except `"select"`
```
TSchiefer, line 18:
```suggestion
#'   - `"all"`: All options above except `"select"`
#' 
#'   Default `NULL` is equivalent to `c("fks", "color")`
```

R/paste.R
```diff
@@ -3,78 +3,214 @@
...
+#'   Deprecated, see `"select"` in the `options` argument.
+#' @param ... Must be empty.
 #' @param tab_width Indentation width for code from the second line onwards
+#' @param options Formatting options. A character vector containing some of:
+#'   - `"tables"`: [tibble()] calls for empty table definitions
+#'     derived from [dm_ptype()], overrides `"select"`.
+#'   - `"select"`: [dm_select()] statements for columns that are part
+#'     of the dm.
+#'   - `"pks"`: [dm_add_pk()] statements for primary keys.
+#'   - `"keys"`: [dm_add_fk()] statements for foreign keys, implies `"pks"`.
```
TSchiefer, line 16: Why not just have one option for setting keys? Do you have a use case when FKS should be explicitly avoided?
```suggestion
#'   - `"keys"`: [dm_add_pk()] and [dm_add_fk()] statements for adding keys.
```
If this suggestion is accepted, further changes later on in the code are required

R/paste.R
```diff
@@ -3,78 +3,214 @@
...
+    return(character())
   }
+  paste0('"', x, '"')
+}
+
+# Errors ------------------------------------------------------------------
+
+abort_unknown_option <- function(options, all_options) {
+  abort(error_txt_unknown_option(options, all_options), .subclass = dm_error_full("unknown_option"))
+}
```
TSchiefer, line 211: Move this and `error_txt_unknown_option()` to error-helpers.R?

krlmlr: No, `error-helpers.R` should disappear at some point, all error constructors should be moved to the files where they are used.

PR #319, opened 2020-04-14
Persistence

krlmlr:
This is a first draft for adding persistence operations.


R/persist-dm.R
```diff
@@ -0,0 +1,248 @@
...
+#'   dm_insert(flights_sqlite, flights_feb_sqlite)
+#'   print(dm_nrow(flights_sqlite))
+#' }
+NULL
+
+
+#' dm_insert
+#'
+#' `dm_insert()` adds new records.
+#' The primary keys must differ from existing records.
```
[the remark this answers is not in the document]
krlmlr: The user can do this with `dry_run = TRUE`, I updated the comment.

R/persist-dm.R
```diff
@@ -0,0 +1,248 @@
...
+  dm_persist(target_dm, dm, tbl_insert, top_down = TRUE, dry_run)
+}
+
+# dm_update
+#
+# `dm_update()` updates existing records.
+# Primary keys must match for all records to be updated.
+#
+# @rdname persist-dm
+# @export
```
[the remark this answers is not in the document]
krlmlr: Not yet.

R/persist-dm.R
```diff
@@ -0,0 +1,248 @@
...
+}
+
+# dm_delete
+#
+# `dm_delete()` removes matching records, based on the primary key.
+# The order in which the tables are processed is reversed.
+#
+# @rdname persist-dm
+# @export
+dm_delete <- function(target_dm, dm, ..., dry_run = FALSE) {
```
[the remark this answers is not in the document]
krlmlr: Good point. I was wondering if this should be the default for all operations. We could also do `dry_run = NULL` and give an explicit message unless the user sets `dry_run`.

R/persist-dm.R
```diff
@@ -0,0 +1,248 @@
...
+  dm_persist(target_dm, dm, tbl_truncate, top_down = FALSE, dry_run)
+}
+
+dm_persist <- function(target_dm, dm, operation, top_down, dry_run = FALSE) {
+  dm_check_persist(target_dm, dm)
+
+  dm_run_persist(target_dm, dm, operation, top_down, dry_run)
+}
+
+dm_check_persist <- function(target_dm, dm) {
```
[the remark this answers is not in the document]
krlmlr: The user can do `dry_run = TRUE` and call `dm_check_constraints()`. What's the benefit of the extra option?

R/persist-dm.R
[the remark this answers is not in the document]
TSchiefer: I like `dry_run = NULL`. This way, the user will always be forced to make a conscious decision.

R/persist-dm.R
```diff
@@ -0,0 +1,250 @@
...
+
+  if (dry_run) {
+    out
+  } else {
+    invisible(out)
+  }
+}
+
+dm_patch_tbl <- function(dm, ...) {
+  check_not_zoomed(dm)
```
TSchiefer, line 240: One more thing: So far this function is only called in `dm_run_persist()` which is called by `dm_persist()`.
In `dm_persist()` all sorts of checks are done, also the `check_not_zoomed()` one.
So unless you are planning to use this function also in other places, we could remove this line
```suggestion
```

TSchiefer:
Are the SQL-Queries for insert, update, upsert the same across different DBMS?

R/rows-d.R
```diff
@@ -0,0 +1,214 @@
...
+#' rows_insert
+#'
+#' `rows_insert()` adds new rows.
+#' This operation corresponds to `INSERT` in SQL.
+#' If `.key` is given, no two rows with the same values in the key columns
+#' are permitted.
+#' @param .sort Should the output be sorted by the key columns?
+#' @rdname rows
+#' @export
+rows_insert <- function(.data, ..., .key = NULL) {
```
hadley, line 37: I now think the basic interface should be `function(x, y, key = NULL)` to match the joins. I think it's unnecessary to allow free floating columns, and it makes the interface substantially simpler. 

That also means that we can be more clear that the output of these functions are consistent with the output of the joins: the return type is always the same type as x.

R/rows-d.R
```diff
@@ -0,0 +1,214 @@
...
+  UseMethod("rows_insert", .data)
+}
+
+#' @export
+#' @rdname rows
+rows_insert.data.frame <- function(.data, ..., .key = NULL, .sort = NULL) {
+  source <- dots_to_df(.data, ...)
```
hadley, line 44: I'd suggest these functions all start with 

```
y <- auto_copy(x, y, copy = copy)
keys <- check_keys(x, y, keys)
```

R/rows-d.R
```diff
@@ -0,0 +1,214 @@
...
+#' @name rows
+#' @example example/rows-d.R
+NULL
+
+
+#' rows_insert
+#'
+#' `rows_insert()` adds new rows.
+#' This operation corresponds to `INSERT` in SQL.
+#' If `.key` is given, no two rows with the same values in the key columns
```
hadley, line 32: For consistency with joins, should this be a character vector called `by`?  Or maybe we should add a `keys` parameter to the joins that uses tidyselect?

R/rows-d.R
```diff
@@ -0,0 +1,214 @@
...
+    out <- arrange(out, !!!data_key)
+  }
+  out
+}
+
+#' rows_update
+#'
+#' `rows_update()` updates existing rows.
+#' This operation corresponds to `UPDATE` in SQL.
+#' `.key` is mandatory and defaults to the first column in `...`.
```
hadley, line 60: I think key needs to have the same default in all these functions; otherwise it'll be too hard to remember the different cases

R/rows-d.R
```diff
@@ -0,0 +1,214 @@
...
+#' @export
+#' @rdname rows
+rows_insert.data.frame <- function(.data, ..., .key = NULL, .sort = NULL) {
+  source <- dots_to_df(.data, ...)
+  source_key <- df_key(source, {{ .key }}, default = !!integer())
+  data_key <- df_key(.data, !!names(source)[source_key])
+
+  out <- vctrs::vec_rbind(.data, source)
+  if (isTRUE(.sort)) {
+    # FIXME: Faster implementation, sort before binding
```
hadley, line 50: I think that's going to be hard to implement because you'd need to sort `x` and `y` individually then merge sort the two together? I doubt that'll be worth the effort.

hadley: But I think it's odd to mix vctrs and dplyr here — instead of arrange, maybe use `vec_order()`?

R/rows-d.R
```diff
@@ -0,0 +1,214 @@
...
+#' `INSERT`, `UPDATE` and `DELETE`.
+#' All operations expect existing and new data to be compatible.
+#'
+#' @param .data Target table object.
+#' @param .key Key columns, unquoted.
+#'   The default varies for the different operations:
+#'   - For `rows_insert()`: no keys.
+#'   - For `rows_update()`, `rows_patch()` and `rows_upsert()`: first column.
+#'   - For `rows_delete()`: all columns.
+#' @param ... New data.
```
hadley, line 17: Should persist be part of the signature of the generic?

krlmlr: `inplace = NULL`

R/rows-d.R
krlmlr: Let's discuss in dplyr.

PR #325, opened 2020-04-20
better error msg for `collect.zoomed_dm()`

### LINKED ISSUE #294: Define behavior of `collect.dm()` for `zoomed_dm`?
It seems that `collect.dm()` still won't work for `zoomed_dm` after e95ff5dbb270b83c38e5a73c54d534469e3b64eb, since it calls `dm_apply_filters()` which in turn calls `check_not_zoomed()`.

And even if it did, the resulting `zoomed_dm` would contain local tibbles whereas the zoomed table would potentially still be on the DB.

I think the underlying question is how we intend `collect()` to work on a `zoomed_dm`.

My suggestion would be to 

1. either not allow it at all for a `zoomed_dm` 
1. or to collect all tables in `def$data` plus the table in `def$zoomed`, in order to create a consistent `dm`-object. In the latter case we would need to think about forbidding filter conditions in general for `collect.dm()`, so the `dm_apply_filters()`-call would be unnecessary.

TSchiefer:
fixes #294


krlmlr:
Thanks!

PR #334, opened 2020-04-27
more changes to test infrastructure

TSchiefer:
as discussed.

- with current CRAN versions of {dbplyr} (1.4.3) and {dplyr} (0.8.5), one test fails; if we skip it, a warning is produced in another test etc.
- MSSQL could not yet be tested


krlmlr:
Thanks!

PR #335, opened 2020-04-27
Support compound keys

### LINKED ISSUE #3: Support for compound keys
Are there any plans/thoughts on supporting compound keys? I frequently have a table where the primary key is a combination of fields. It looks like putting a PK constraint on something like that in `dm` requires creating a new single column that is a mutation of the two non-unique components.

Thanks for the package!

krlmlr:
Closes #3.


krlmlr:
NEWS entry:

```
- `dm_get_all_fks()` includes `parent_pk_cols` column that describes the primary key columns of the parent table.
```

krlmlr:
NEWS entry:

```
- `dm_get_all_fks()` includes `parent_pk_cols` column that describes the primary key columns of the parent table.
- `dm_examine_constraints()` and other check functions count the number of rows that violate constraints for primary and foreign keys.
```

krlmlr:
NEWS entry:

```
- `dm_get_all_fks()` includes `parent_pk_cols` column that describes the primary key columns of the parent table.
- `dm_examine_constraints()` and other check functions count the number of rows that violate constraints for primary and foreign keys.
- `copy_dm_to(set_key_constraints = FALSE)` downgrades unique indexess to regular indexes.
- `rows_truncate()` implemented for data frames.
- `dm_nycflights13()` returns a data model with compound keys by default. Use `compound = FALSE` to return the data model from dm 0.1.13 or earlier.
- `dm_enum_fk_candidates()` enumerates column in the order they apper in the table.
```

PR #364, opened 2020-05-15
Better default for 'copy_dm_to()'

### LINKED ISSUE #80: Default settings for `cdm_copy_to()`
The default parameters setting for `cdm_copy_to()` should produce temporary tables with `unique_table_names`.

Idea is to have the default `unique_table_names = NULL`, which becomes `TRUE` when `temporary = TRUE` and `FALSE` otherwise.

Could either be fixed in #79 or in a future PR.

TSchiefer:
new interface: 
```r
# changed argument:
table_names = repair_table_names_for_db(src_tbls(dm), schema, temporary)
# new argument:
schema = NULL
# removed argument:
unique_table_names = FALSE
```
with
``` r
orig_table_names <- c("t1", "t2", "t3")
dm:::repair_table_names_for_db(table_names = orig_table_names, schema = NULL, temporary = TRUE)
#> <IDENT> t1_2020_05_15_10_03_41_1
#> <IDENT> t2_2020_05_15_10_03_41_1
#> <IDENT> t3_2020_05_15_10_03_41_1
dm:::repair_table_names_for_db(table_names = orig_table_names, schema = NULL, temporary = FALSE)
#> <IDENT> t1
#> <IDENT> t2
#> <IDENT> t3
dm:::repair_table_names_for_db(table_names = orig_table_names, schema = "test", temporary = TRUE)
#> Warning in dm:::repair_table_names_for_db(table_names = orig_table_names, :
#> Ignoring argument `schema` because `temporary = TRUE`.
#> <IDENT> t1_2020_05_15_10_03_41_2
#> <IDENT> t2_2020_05_15_10_03_41_2
#> <IDENT> t3_2020_05_15_10_03_41_2
dm:::repair_table_names_for_db(table_names = orig_table_names, schema = "test", temporary = FALSE)
#> <IDENT> test.t1
#> <IDENT> test.t2
#> <IDENT> test.t3
```

<sup>Created on 2020-05-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

WIP because:
- Tests missing
- documentation missing

closes #80


R/db-interface.R
```diff
@@ -42,11 +42,11 @@
 copy_dm_to <- function(dest, dm, ...,
                        types = NULL, overwrite = NULL,
                        indexes = NULL, unique_indexes = NULL,
-                       set_key_constraints = TRUE, unique_table_names = FALSE,
-                       table_names = NULL,
+                       set_key_constraints = TRUE,
+                       table_names = repair_table_names_for_db(src_tbls(dm), schema, temporary),
```
krlmlr, line 46: Defaulting to an unexported function looks odd. Can we keep the `NULL` default and use `repair_table_names_for_db()` internally?

TSchiefer:
- added further tests (there are two new FIXMEs; could you please have a look at them?)
- added documentation

also made ensureed in the tests, that we know if we are copying from DB to local or vice versa (95699a3)

resolving WIP status

krlmlr:
Thanks. I no longer think we should add a `schema` argument for now, it's getting more complex than necessary. Instead let's allow passing functions in `table_name`.

Agree to deprecate `unique_table_names` argument, now supported via `table_name = identity`.

I tweaked, can you please pick up from there?

R/db-interface.R
```diff
@@ -14,39 +14,72 @@
...
+#'
+#'   If left `NULL` (default), the names will be determined automatically depending on the `temporary` argument:
+#'
+#'   1. `temporary = TRUE` (default): unique table names based on the names of the tables in the `dm` are created.
+#'   1. `temporary = FALSE`: the table names in the `dm` are used as names for the tables on `dest`.
+#'
+#'   If a function or one-sided formula, `table_names` is converted to a function
+#'   using [rlang::as_function()].
+#'   This function is called with the table names of the `dm` object
+#'   as only argument, and is expected to return a character vector
```
jawond, line 33:
```suggestion
#'   as the only argument, and is expected to return a character vector
```

krlmlr:
Thanks!

PR #369, opened 2020-05-22
Add GitHub Actions workflow

pat-s:
## Notes

- To add support for `update-tic.yml`, create and add a GITHUB_PAT secret with workflow scopes via `tic::gha_add_secret(<secret>, "TIC_UPDATE")`

- cynkra/cynkratemplate is already added

- (TIC_ONLY_TESTS & TIC_DEV_VERSIONS) runner still errors. I assume something with the database setup is not working correctly yet. You might spot earlier than me what's wrong.


krlmlr:
Thanks.

- [ ] `TIC_ONLY_TESTS` with `TIC_DEV_VERSIONS` is permitted to fail. How does GHA support this?
- [ ] By default, only SQLite is tested -- no database needs to be installed for most jobs. For the database tests, I'd prefer one job per database, and install only that database for each job. (It's handled differently on Travis, we can do better here.) The `DM_TEST_SRC` environment variable specifies which database to test against, for the database tests set this and `TIC_ONLY_TESTS`
- [ ] Can you please add the MSSQL setup from the odbc or dbplyr tests?
- [ ] Why does GHA install from source on Windows?
- [ ] Do we need LaTeX?
- [ ] Does it work on R 3.3?
- [ ] Can we do the same that tidyverse does for installing packages on Linux?

pat-s:
> TIC_ONLY_TESTS with TIC_DEV_VERSIONS is permitted to fail. How does GHA support this?

Fails with four failed tests?

> Why does GHA install from source on Windows?

Usually binaries are used. No Windows builds are being run in the current setup.
Where did you see Windows builds?

For now I just mirrored the Travis setup with the difference of running R-devel checks on macOS instead of Linux.
Should every test also run on Windows? This would expand the build matrix substantially.

> Do we need LaTeX?

Some checks in R CMD Check (pdf manual) are only done when LaTeX is installed.
One could also just do it in certain runners.

> Does it work on R 3.3?

I guess so. I'll add it to the matrix.

> Can we do the same that tidyverse does for installing packages on Linux?

I'm not yet convinced from binaries on Linux. Does it really work? This should be discussed in {tic} then.

krlmlr:
Thanks!

PR #392, opened 2020-06-05
update documentation after #319

TSchiefer:
- agreed doc updates
- two types of FIXMEs

### DIFF
--- a/R/rows-db.R
+++ b/R/rows-db.R
@@ -163,6 +163,7 @@ check_db_superset <- function(x, y, by) {
 #' @rdname rows-db
 sql_rows_insert <- function(x, y, ...) {
   ellipsis::check_dots_used()
+  # FIXME: check here same src for x and y? if not -> error.
   UseMethod("sql_rows_insert")
 }
 
@@ -184,6 +185,7 @@ sql_rows_insert.tbl_sql <- function(x, y, ...) {
 #' @rdname rows-db
 sql_rows_update <- function(x, y, by, ...) {
   ellipsis::check_dots_used()
+  # FIXME: check here same src for x and y? if not -> error.
   UseMethod("sql_rows_update")
 }
 
--- a/R/rows-db.R
+++ b/R/rows-db.R
@@ -54,6 +54,7 @@ rows_insert.tbl_dbi <- function(x, y, by = NULL, ...,
     invisible(x)
   } else {
     # Checking mandatory by default, opt-out
+    # FIXME: contrary to doc currently also checks if `in_place = FALSE`
     if (is_null(check) || is_true(check)) {
       check_db_dupes(x, y, by)
     }
@@ -90,6 +91,7 @@ rows_update.tbl_dbi <- function(x, y, by = NULL, ...,
     invisible(x)
   } else {
     # Checking optional, can rely on primary key constraint
+    # FIXME: contrary to doc currently also checks if `in_place = FALSE`
     if (is_null(check) || is_true(check)) {
       check_db_superset(x, y, by)
     }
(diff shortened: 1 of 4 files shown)


R/rows-db.R
```diff
@@ -87,6 +93,7 @@ rows_update.tbl_dbi <- function(x, y, by = NULL, ...,
     invisible(x)
   } else {
     # Checking optional, can rely on primary key constraint
+    # FIXME: contrary to doc currently also checks if `in_place = FALSE`
```
krlmlr, line 96: Does the documentation need to change here? We want to offer checking in both cases, and enable by default only if `in_place = FALSE`.

R/rows-db.R
```diff
@@ -21,8 +24,10 @@
 #'   Set to `TRUE` to always check keys, or `FALSE` to never check.
 #'   The default is to check only if `in_place` is `TRUE` or `NULL`.
 #'
+#'   Currently these checks are no-ops and need yet to be implemented.
+#'
 #' @return A tbl object of the same structure as `x`.
-#'   If `in_place = TRUE`, [invisible] and identical to `x`.
+#'   If `in_place = TRUE`, a modified version of `x` is returned invisibly.
```
krlmlr, line 30: `x` is guaranteed to be unchanged, only the contents of the underlying data. Same for the dm case.

R/rows-db.R
TSchiefer: Do you want checks by default if `in_place = NULL`? If not we need to tweak.

Is your point, that if someone does `in_place = TRUE`, they know what they're doing and no additional checks are needed, is that right?

R/rows-db.R
TSchiefer: Could we please quickly compare our personal semantics here?

When we're talking about a `tbl object`, what do we mean? 
1. a reference to a table
2. the table itself (including everything)
3. the skeleton of a table (colnames, coltypes, data source)

I thought we were talking about 2. Does it depend on the data source what we mean by `tbl object`?

TSchiefer: I guess that locally a table including the data is returned and thus you need to say that `x` is modified (as it is done in rows.R). And on the other hand, the `x` referring to a remote table is just a reference to that table.
But then we can't really talk about the "same structure as `x`", when it's about a remote table, cause it would be rather meaningless, right?

Maybe as an alternative, we could write something like:

"A tbl object referring to a table with the same structure as `x`. If `in_place = TRUE`, [invisible] and referring to the now modified version of `x`."

R/rows-db.R
krlmlr: AFAICT, the implementation checks

- by default
- if `in_place = FALSE`
- if `in_place = TRUE, check = TRUE`

This is too convoluted, agree to simplify. We should check by default, with a message if it's likely that the user wants the checks off.

krlmlr: That part would be consistent with dplyr, and inconsistent with `add_pk()`. I guess we can't have both.

R/rows-db.R
krlmlr: This is different for mutable and immutable data sources. The former only point to the data, the latter contain the data.

```suggestion
#'   If `in_place = TRUE`, the underlying data is updated as a side effect and `x` is returned, invisibly.
```

R/rows-db.R
TSchiefer: Ok, just to make sure I got it right:

- `check = TRUE` by default (`NULL` is no longer allowed)
- inform in case: `in_place = TRUE`.

Additionally, should we also inform in case: `in_place = NULL`?

krlmlr: `check = NULL` remains the default, we inform if `check = NULL` if we think that `check = FALSE` is better.

We always inform with `in_place = NULL`?

krlmlr:
Thanks!

PR #393, opened 2020-06-05
printing of empty dm

### LINKED ISSUE #386: Don't print table source
if the dm doesn't have tables.

Reference: #385.

TSchiefer:
closes #386


TSchiefer:
yeah, why not. maybe `empty_dm()`; or `empty object of class dm`; can't think of anything better currently

krlmlr:
```
dm()
```

would be consistent with other classes.

krlmlr:
Thanks!

PR #398, opened 2020-06-08
GHA: Use a custom PAT to increase rate limit

pat-s:
As discussed in https://github.com/krlmlr/dm/issues/374#issuecomment-640413457.

Using a custom GitHub PAT (scopeless) as a secret instead of the default `GITHUB_TOKEN` secrets bumps the rate from 1000 to 5000: https://github.com/krlmlr/dm/runs/748994181?check_suite_focus=true#step:18:95


krlmlr:
Thanks! I guess we can still push using the default `$GITHUB_TOKEN`?

pat-s:
Ah. By default, `GITHUB_TOKEN` is a hidden secret, not being used as an env var unless explicitly set.

Doing so is not needed for push/pull access. 

It was only set as an env var to have a cheap way to workaround rate limits in {remotes}, which should work for most people with its default limit of 1000.

PR #399, opened 2020-06-08
GHA: Remove `ccache` parts from tic.yml

pat-s:
and update caching action version.

gfortran 8.2 (CRAN default) is meanwhile also installed upstream by r-lib/actions.


krlmlr:
Thanks!

PR #407, opened 2020-06-22
Update tic templates [ci-skip]

pat-s:
{tic} templates can be updated :rocket: :robot:


pat-s:
I'm still somewhat surprised but this automation seems to work quite well 😄

.github/workflows/tic.yml
```diff
@@ -23,55 +23,49 @@ jobs:
         config:
           # Custom configs
           # [Custom matrix env var]
-          - { os: macOS-latest, r: "devel", TIC_ONLY_TESTS: true, TIC_DEV_VERSIONS: true, id: "TIC_ONLY_TESTS & TIC_DEV_VERSIONS", cache: "dev" }
+          - { os: ubuntu-latest, r: 3.6, id: "default" }
```
krlmlr, line 26: Why is the order reversed here?

pat-s: this is due to the iterative approach internally which writes the last item in the loop (here `"custom matrix env var"`) right after the respective anchor (`config:` here).

krlmlr: Would sorting help get rid of the problem and also provide some well-defined order in the config here?

pat-s: A custom sort enforced by tic would probably not be desired by users?
Right now, the order is reversed but the the internal sequence between the config is kept (at least).

But yeah, thinking about it there would be another reverse going on during the next update via the logic I described above.
Definitely room for improvement.

PR #408, opened 2020-06-23
Add DB tests

pat-s:
Copied most from the {odbc} test suite.
SQLite tests seem to be working.

I also stripped the main `tic.yml` by one job and all DB installations since these now run in their own workflows.

@krlmlr 

- [x] Do I need some local credentials to get the mssql tests working?

- [x] Do I need some local credentials to get the postgres tests working?

- [ ] Are we testing MySQL? (is the MariaDB driver for MySQL?) -> Do we need the `install-mariadb.sh` script?


krlmlr:
You need a password, and perhaps also adapt the user name:

https://github.com/krlmlr/dm/pull/408/checks?check_run_id=799930438#step:8:40

The database tests should be failing without a database connection, not sure why I implemented `my_test_src()` the way I did. Can we replace `skip()` by `abort()`?

See `test_src_mssql()` and friends for details of the connection process. I don't mind special-casing here: "if GHA: connect with hard-coded password, else: connect the way we do now".

.github/workflows/db-test-mssql.yml
```diff
@@ -0,0 +1,107 @@
+on:
+  push:
+  pull_request:
+  # for now, CRON jobs only run on the default branch of the repo (i.e. usually on master)
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: "0 4 * * *"
+
+name: db-test-mssql
```
krlmlr, line 9: Let's go with shorter names, also for the file names.

```suggestion
name: db-mssql
```

krlmlr:
With 1b0aaa7 I suspect database tests start failing properly, until they work.

pat-s:
Connections are working now. The test errors are probably best corrected by you.

Still open: Tests on MySQL? I can see `test_src_maria()` but no `test-mysql` in the Makefile?

tests/testthat/helper-config-db.R
```diff
@@ -7,7 +7,11 @@ test_src_sqlite <- function() {
 }
 
 test_src_postgres <- function() {
-  con <- DBI::dbConnect(RPostgres::Postgres())
+  if (!is.null(Sys.getenv("CI"))) {
+    con <- DBI::dbConnect(odbc::odbc(), "PostgreSQL")
```
krlmlr, line 11: This should use {RPostgres} also on CI.

krlmlr:
Thanks!

PR #417, opened 2020-07-15
function for merging dm-objects

### LINKED ISSUE #404: add new function 'dm_merge()' (or another name)
I am working on a project where I have 2 `dm` objects, `dm_1` and `dm_2`. One table of `dm_1` has a column that is pointing to to a column in another table in `dm_2`, but currently I cannot easily bring those two `dms` together.

I suggest a function, e.g. `dm_merge(...)`, which takes several `dm` objects and produces one `dm` in which the original `dms` are coexisting (basically `bind_rows()` of the `def`-tibbles).

TSchiefer:
- Function `dm_merge()` merges one or more dm's into one
- name repair similar to `dm_add_tbl()`

closes #404


R/merge.R
```diff
@@ -0,0 +1,41 @@
...
+
+  walk(dms, check_dm)
+  walk(dms, check_not_zoomed)
+
+  table_names <- map(dms, src_tbls) %>% flatten_chr()
+  new_table_names <- repair_names_vec(table_names, repair, quiet)
+
+  dms_def <- map(dms, dm_get_def)
+  reduce(dms_def, bind_rows) %>%
+    mutate(table = new_table_names) %>%
```
krlmlr, line 29: What happens to the primary and foreign keys if the tables are renamed? Do we need to use `dm_rename_tbl()` instead?

R/merge.R
```diff
@@ -0,0 +1,41 @@
...
+#'
+#' @return `dm` containing the tables and key relations of all `dm`-objects.
+#' @export
+#'
+#' @examples
+#' dm_1 <- dm_nycflights13()
+#' dm_2 <- dm(mtcars, iris)
+#' dm_merge(dm_1, dm_2)
+dm_merge <- function(..., repair = "check_unique", quiet = FALSE) {
+  if (dots_n(...) == 0) abort_empty_ellipsis("dm_merge()")
```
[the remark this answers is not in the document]
TSchiefer: `dm_bind()` sounds good to me. And the result of the empty ellipsis being `dm()` would then be in line with the behavior of e.g. `bind_rows()`.

R/merge.R
TSchiefer: absolutely right, thank you

R/bind.R
```diff
@@ -0,0 +1,38 @@
...
+#'
+#' @return `dm` containing the tables and key relations of all `dm`-objects.
+#' @export
+#'
+#' @examples
+#' dm_1 <- dm_nycflights13()
+#' dm_2 <- dm(mtcars, iris)
+#' dm_bind(dm_1, dm_2)
+dm_bind <- function(..., repair = "check_unique", quiet = FALSE) {
+  if (dots_n(...) == 0) return(dm())
```
krlmlr, line 18: Do we need this special case?

TSchiefer: Short answer: I am not sure.

You are right, no error is thrown and the printed output of `dm_bind()` without that line looks like that of an empty dm.
Nevertheless there is a difference between `dm()` and the result of `dm_bind()`:
```r
unclass(dm_bind())
> $def
> data frame with 0 columns and 0 rows
dm()
> $def
> # A tibble: 0 x 9
> # … with 9 variables: table <chr>, data <list>, segment <chr>, display <chr>, pks <list<tbl_df[,1]>>, fks  <list<tbl_df[,2]>>, filters <list<tbl_df[,2]>>, zoom <list>,
> #   col_tracker_zoom <list>
```

I would probably keep the special case.

TSchiefer:
> Thanks. What happens if the input dms use different sources?

...an invalid `dm`-object is produced, argh! 
Just realized it actually makes it harder to design a general purpose function if you actually write it for your own use case.

R/bind.R
krlmlr: Sure -- without a dm we don't get the columns. Let's keep the special case and document it.

TSchiefer: ok, documented: dab9826

krlmlr:
Can you please merge master, fix the TIC_ONLY_IMPORTS tests and merge into master?

krlmlr:
(You need https://github.com/r-lib/roxygen2/pull/993.)

TSchiefer:
seems to be fine now 

there is one FIXME, where I want to check the output of an expression which requires {dbplyr} to be installed (0254d9b)

krlmlr:
We can do a conditional `verify_output()`, see e.g. `test-paste.R`.

krlmlr:
Thanks, good job!

PR #423, opened 2020-07-28
fix test

TSchiefer:
It seems to me that due to the commit https://github.com/krlmlr/dm/pull/419/commits/60882491aff6342f152bb9a4915fe1f2e4477565 `#` is also part of the remote name and therefore the test would fail on MSSQL if we don't add it to the pattern.


krlmlr:
Thanks!

PR #428, opened 2020-07-29
also learn schema name for MSSQL

### LINKED ISSUE #367: dm_from_src() MS SQL Server
Hi

I'm trying to import a database from our server, but get the error below.

`dw_dm <- dm_from_src(conn, learn_keys = TRUE)`

`Error: nanodbc/nanodbc.cpp:1617: 42000: [Microsoft][ODBC SQL Server Driver][SQL Server]Invalid object name 'AdditionalFundingRequestsFact'.  [Microsoft][ODBC SQL Server Driver][SQL Server]Statement(s) could not be prepared. 
<SQL> 'SELECT *
FROM AdditionalFundingRequestsFact AS "zzz15"
WHERE (0 = 1)'`

The table in question definitely exists (I can locate it and extract using `tbl() %>% collect()` just fine).
There are a number of different schema in the database, but it's the very first table in the database alphabetically.

Our server is running:
MSS version: 12.00.4416
R version: 4.0.0

TSchiefer:
now all tables can be accessed, independent of schema name.

In addition, IMHO an argument `schema` should also be provided for `dm_from_src()`.

closes #367

### DIFF
--- a/R/db-interface.R
+++ b/R/db-interface.R
@@ -142,8 +142,9 @@ copy_dm_to <- function(dest, dm, ...,
         table_names_out <- table_names
       }
       check_naming(names(table_names_out), src_names)
-
-      table_names_out <- unclass(DBI::dbQuoteIdentifier(dest_con, table_names_out[src_names]))
+      if (!inherits(table_names_out,"ident_q")) {
+        table_names_out <- unclass(DBI::dbQuoteIdentifier(dest_con, table_names_out[src_names]))
+      }
       # names(table_names_out) <- src_names
     }
 
--- a/R/db-interface.R
+++ b/R/db-interface.R
@@ -148,9 +148,7 @@ copy_dm_to <- function(dest, dm, ...,
         abort_copy_dm_to_table_names_duplicated(problem)
       }
 
-      if (!inherits(table_names_out,"ident_q")) {
-        table_names_out <- unclass(DBI::dbQuoteIdentifier(dest_con, table_names_out[src_names]))
-      }
+      table_names_out <- unclass(DBI::dbQuoteIdentifier(dest_con, table_names_out[src_names]))
       # names(table_names_out) <- src_names
     }
 
--- a/R/db-interface.R
+++ b/R/db-interface.R
@@ -106,9 +106,9 @@ copy_dm_to <- function(dest, dm, ...,
     abort_no_unique_indexes()
   }
 
-  if (!is_null(unique_table_names)) {
-    lifecycle::deprecate_soft(
-      "0.1.4", "copy_dm_to(unique_table_names = )",
+  if (!is.null(unique_table_names)) {
+    deprecate_soft(
+      "0.1.4", "dm::copy_dm_to(unique_table_names = )",
       details = "Use `table_names = identity` to use unchanged names for temporary tables."
     )
 
@@ -142,6 +142,12 @@ copy_dm_to <- function(dest, dm, ...,
         table_names_out <- table_names
       }
       check_naming(names(table_names_out), src_names)
+
+      if (anyDuplicated(table_names_out)) {
+        problem <- table_names_out[duplicated(table_names_out)][[1]]
+        abort_copy_dm_to_table_names_duplicated(problem)
+      }
+
       if (!inherits(table_names_out,"ident_q")) {
         table_names_out <- unclass(DBI::dbQuoteIdentifier(dest_con, table_names_out[src_names]))
       }
@@ -152,11 +158,10 @@ copy_dm_to <- function(dest, dm, ...,
     table_names_out <- map(table_names_out, dbplyr::ident_q)
   } else {
     # FIXME: Other data sources than local and database possible
-    if (!is.null(table_names)) {
-      lifecycle::deprecate_soft(
-        "0.1.6", "copy_dm_to(table_names = 'must be NULL if copying to a local source')"
-      )
-    }
+    deprecate_soft(
+      "0.1.6", "dm::copy_dm_to(dest = 'must refer to a remote data source')",
+      "dm::collect.dm()"
+    )
     table_names_out <- set_names(src_names)
   }
 
@@ -249,10 +254,21 @@ check_naming <- function(table_names, dm_table_names) {
 
 # Errors ------------------------------------------------------------------
 
-abort_copy_dm_to_table_names <- function(problems) {
+abort_copy_dm_to_table_names <- function() {
   abort(error_txt_copy_dm_to_table_names(), .subclass = dm_error_full("copy_dm_to_table_names"))
 }
 
 error_txt_copy_dm_to_table_names <- function() {
   "`table_names` must have names that are the same as the table names in `dm`."
+}
+
+abort_copy_dm_to_table_names_duplicated <- function(problem) {
+  abort(error_txt_copy_dm_to_table_names_duplicated(problem), .subclass = dm_error_full("copy_dm_to_table_names_duplicated"))
+}
+
+error_txt_copy_dm_to_table_names_duplicated <- function(problem) {
+  c(
+    "`table_names` must be unique.",
+    i = paste0("Duplicate: ", tick(problem))
+  )
 }
--- a/R/db-interface.R
+++ b/R/db-interface.R
@@ -245,6 +245,16 @@ dm_set_key_constraints <- function(dm) {
   invisible(dm)
 }
 
+get_db_table_names <- function(dm) {
+  if (!is_src_db(dm)) {
+    return(tibble(table_name = src_tbls(dm), remote_name = src_tbls(dm)))
+  }
+  tibble(
+    table_name = src_tbls(dm),
+    remote_name = map_chr(dm_get_tables_impl(dm), dbplyr::remote_name)
+  )
+}
+
 check_naming <- function(table_names, dm_table_names) {
   if (!identical(sort(table_names), sort(dm_table_names))) {
     abort_copy_dm_to_table_names()
--- a/R/dm-from-src.R
+++ b/R/dm-from-src.R
@@ -23,6 +23,8 @@
 #'   Currently supports `schema` (default: `"public"`)
 #'   and `table_type` (default: `"BASE TABLE"`) for Postgres databases.
 #'
+#'   For MSSQL databases `schema` (default: `"dbo"`) is supported.
+#'
 #' @return A `dm` object.
 #'
 #' @export
--- a/R/dm-from-src.R
+++ b/R/dm-from-src.R
@@ -10,14 +10,14 @@
 #' @param table_names
 #'   A character vector of the names of the tables to include.
 #' @param learn_keys
-#'   \lifecycle{experimental}
+#'   `r lifecycle::badge("experimental")`
 #'
 #'   Set to `TRUE` to query the definition of primary and
 #'   foreign keys from the database.
 #'   Currently works only for Postgres and SQL Server databases.
 #'   The default attempts to query and issues an informative message.
 #' @param ...
-#'   \lifecycle{experimental}
+#'   `r lifecycle::badge("experimental")`
 #'
 #'   Additional parameters for the schema learning query.
 #'   Currently supports `schema` (default: `"public"`)
--- a/R/dm-from-src.R
+++ b/R/dm-from-src.R
@@ -38,7 +38,7 @@
 #' dm_from_src(con)
 #'
 #' DBI::dbDisconnect(con)
-dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL, schema = NULL,
+dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL,
                         ...) {
   if (is_null(src)) {
     # FIXME: Check empty arguments and ellipsis
--- a/R/dm-from-src.R
+++ b/R/dm-from-src.R
@@ -49,8 +49,7 @@ dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL,
   con <- con_from_src_or_con(src)
 
   if (is.null(learn_keys) || isTRUE(learn_keys)) {
-    if (is.null(schema)) schema <- db_standard_schema(src)
-    dm_learned <- dm_learn_from_db(src, schema = schema, ...)
+    dm_learned <- dm_learn_from_db(src, ...)
 
     if (is.null(dm_learned)) {
       if (isTRUE(learn_keys)) {
--- a/R/dm-from-src.R
+++ b/R/dm-from-src.R
@@ -38,7 +38,7 @@
 #' dm_from_src(con)
 #'
 #' DBI::dbDisconnect(con)
-dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL,
+dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL, schema = NULL,
                         ...) {
   if (is_null(src)) {
     # FIXME: Check empty arguments and ellipsis
@@ -49,7 +49,8 @@ dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL,
   con <- con_from_src_or_con(src)
 
   if (is.null(learn_keys) || isTRUE(learn_keys)) {
-    dm_learned <- dm_learn_from_db(src, ...)
+    if (is.null(schema)) schema <- db_standard_schema(src)
+    dm_learned <- dm_learn_from_db(src, schema = schema, ...)
 
     if (is.null(dm_learned)) {
       if (isTRUE(learn_keys)) {
--- a/R/learn.R
+++ b/R/learn.R
@@ -62,7 +62,7 @@ dm_learn_from_db <- function(dest, ...) {
   table_names <-
     overview %>%
     arrange(table) %>%
-    select(schema, table) %>%
+    distinct(schema, table) %>%
     transmute(
       name = table,
       value = schema_if(schema, DBI::dbQuoteIdentifier(con, table))
--- a/R/learn.R
+++ b/R/learn.R
@@ -122,11 +122,11 @@ mssql_learn_query <- function(con, schema = "dbo") { # taken directly from {data
       and ind_col.column_id = cols.column_id
     left outer join sys.systypes [types] on
       types.xusertype = cols.system_type_id
-  where tabs.schema_id = schema_id(%s)
+  where tabs.schema_id = schema_id('%s')
   order by
     tabs.create_date,
     cols.column_id",
-    dbQuoteString(con, schema)
+    schema
   )
 }
 
--- a/R/learn.R
+++ b/R/learn.R
@@ -126,7 +126,7 @@ mssql_learn_query <- function(con, schema = "dbo") { # taken directly from {data
   order by
     tabs.create_date,
     cols.column_id",
-    schema
+  gsub("(^|[^'])(['])($|[^'])", "\\1\\2\\2\\3", schema)
   )
 }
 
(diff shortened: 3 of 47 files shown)


TSchiefer:
Regarding ed3e7bb: for some reason we didn't realize earlier that `table_names` contained repetitive entries for each column in the respective tables. This didn't influence the end result though. (this has nothing to do with the intent of this PR, but I thought I'd add it, since it's at least somehow related)

TSchiefer:
WIP cause I still have to ensure that the schema name is quoted correctly for MSSQL

R/dm-from-src.R
```diff
@@ -38,7 +38,7 @@
 #' dm_from_src(con)
 #'
 #' DBI::dbDisconnect(con)
-dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL,
+dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL, schema = NULL,
```
krlmlr, line 41: I'd rather leave out the formal argument for now, and pass it through the ellipsis the same way as with Postgres. This will make the pull request much smaller.

```suggestion
dm_from_src <- function(src = NULL, table_names = NULL, learn_keys = NULL,
```

R/learn.R
```diff
@@ -119,9 +121,12 @@ mssql_learn_query <- function() { # taken directly from {datamodelr}
       and ind_col.column_id = cols.column_id
     left outer join sys.systypes [types] on
       types.xusertype = cols.system_type_id
+  where schema_name(tabs.schema_id) = %s
```
krlmlr, line 124: Perhaps:

```suggestion
  where tabs.schema_id = schema_id(%s)
```

In the worst case it doesn't change anything, in the best case the query is a bit faster.

R/learn.R
```diff
@@ -47,22 +47,23 @@ dm_learn_from_db <- function(dest, ...) {
...
   }
 
   table_names <-
     overview %>%
     arrange(table) %>%
-    select(schema, table) %>%
+    distinct(schema, table) %>%
     transmute(
       name = table,
       value = schema_if(schema, DBI::dbQuoteIdentifier(con, table))
```
krlmlr, line 69: Suggesting to pass `con` to `schema_if()` and do the quoting there.

R/learn.R
```diff
@@ -119,9 +121,12 @@ mssql_learn_query <- function() { # taken directly from {datamodelr}
       and ind_col.column_id = cols.column_id
     left outer join sys.systypes [types] on
       types.xusertype = cols.system_type_id
+  where schema_name(tabs.schema_id) = %s
   order by
     tabs.create_date,
-    cols.column_id"
+    cols.column_id",
+  dbQuoteString(con, schema)
```
krlmlr, line 128:
```suggestion
    dbQuoteString(con, schema)
```

tests/testthat/test-dm-from-src.R
```diff
@@ -8,7 +8,8 @@ test_that("table identifiers are quoted", {
     map_chr(dbplyr::remote_name)
 
   con <- dm_get_con(dm)
-  expect_equal(unname(remote_names), unclass(DBI::dbQuoteIdentifier(con, names(dm))))
+  # `gsub()`, cause schema names are part of the remote_names (also standard schemas "dbo" for MSSQL and "public" for Postgres).
+  expect_equal(gsub("^.*\\.", "", unname(remote_names)), unclass(DBI::dbQuoteIdentifier(con, names(dm))))
```
krlmlr, line 12: I wonder how this test could even succeed before?

TSchiefer: It does fail locally on Postgres for me. Anyway...

TSchiefer:
regarding: af980c2

As I see it, in the current master `copy_dm_to()` is not ready to deal with schemas.
My problem when setting up a test with a `dm` in a schema was, that when I used something like:
```r
copy_dm_to(src_mssql, dm_for_disambiguate(), temporary = FALSE, table_names = function(x) {in_schema("testthat_for_dm", x)})
```
the tables would not end up in the schema "testthat_for_dm" but in "dbo", with names names like "testthat_for_dm.name_1" etc.

The problem is that the whole name was quoted as one object, but in order to deal with schemas, the dot between schema and table name can't be quoted.

I did not really fix this problem here, just made some sort of hack that assumes, that if the user already uses `ident_q`-class names, that they thought about this problematic and chose names that correctly address schemas.

I will open an issue addressing this though.

Meanwhile, if you don't like it, you can take it out, but then we need to also disable the tests until the issue is resolved (or you tell me how to use schemas in the current implementation of `copy_dm_to()`)

R/learn.R
```diff
@@ -119,9 +122,12 @@ mssql_learn_query <- function() { # taken directly from {datamodelr}
       and ind_col.column_id = cols.column_id
     left outer join sys.systypes [types] on
       types.xusertype = cols.system_type_id
+  where tabs.schema_id = schema_id('%s')
```
krlmlr, line 125:
```suggestion
  where tabs.schema_id = schema_id(%s)
```

R/learn.R
```diff
@@ -119,9 +122,12 @@ mssql_learn_query <- function() { # taken directly from {datamodelr}
...
   order by
     tabs.create_date,
-    cols.column_id"
+    cols.column_id",
+    schema
```
krlmlr, line 129:
```suggestion
    DBI::dbQuoteIdentifier(con, schema)
```

Otherwise the SQL is bad and prone to injection if `schema` contains `'`.

tests/testthat/test-learn.R
```diff
@@ -1,49 +1,87 @@
...
-  def_original <-
-    dm_get_def(dm_for_filter()) %>%
-    select(-data)
-
-  expect_identical(
-    def_learned_renamed_reclassed,
-    def_original
+  # in case there happen to be other tables in schema "dbo"
+  dm_for_filter_mssql_learned <- dm_select_tbl(
+    dm_for_filter_mssql_learned_all,
```
krlmlr, line 19: When using the pipe, please start with the first argument. Mixing piped and non-piped code is confusing.

```suggestion
  dm_for_filter_mssql_learned <-
    dm_for_filter_mssql_learned_all %>%
    dm_select_tbl(
```

tests/testthat/test-learn.R
```diff
@@ -1,49 +1,87 @@
...
 
+  skip_if_src_not("mssql")
   # dm_learn_from_mssql() --------------------------------------------------
-  src_mssql <- skip_if_error(src_test("mssql"))
+  src_mssql <- my_test_src()
 
   # create an object on the MSSQL-DB that can be learned
   if (!any(src_tbls(src_mssql) %>%
     grepl("^tf_1_", .))) {
     copy_dm_to(src_mssql, dm_for_filter(), temporary = FALSE, table_names = unique_db_table_name)
```
krlmlr, line 12: - Ensure qualified identifiers are not quoted
- Use copied dm to query actual table names for a more robust test

```suggestion
    dm_for_filter_copied <- copy_dm_to(src_mssql, dm_for_filter(), temporary = FALSE, table_names = ~ DBI::SQL(unique_db_table_name(.x)))
```

R/db-interface.R
```diff
@@ -148,7 +148,9 @@ copy_dm_to <- function(dest, dm, ...,
         abort_copy_dm_to_table_names_duplicated(problem)
       }
 
-      table_names_out <- unclass(DBI::dbQuoteIdentifier(dest_con, table_names_out[src_names]))
+      if (!inherits(table_names_out,"ident_q")) {
```
krlmlr, line 151: It already works if we pass in an already quoted object, e.g. created by `DBI::SQL()`.

tests/testthat/test-learn.R
```diff
@@ -1,49 +1,87 @@
...
+    tf_1 = starts_with("tf_1"), tf_2 = starts_with("tf_2"), tf_3 = starts_with("tf_3"),
+    tf_4 = starts_with("tf_4"), tf_5 = starts_with("tf_5"), tf_6 = starts_with("tf_6"))
+
+  expect_equivalent_dm(
+    dm_for_filter_mssql_learned,
+    dm_for_filter()
+  )
+
+  walk(
+    dm_get_tables_impl(dm_for_filter_mssql_learned)[c(2, 1, 5, 6, 4, 3)],
```
krlmlr, line 31: Should we index by name here instead?

tests/testthat/test-learn.R
```diff
@@ -1,49 +1,87 @@
...
-
+test_that("Standard learning from MSSQL (schema 'dbo') works?", {
 
+  skip_if_src_not("mssql")
   # dm_learn_from_mssql() --------------------------------------------------
-  src_mssql <- skip_if_error(src_test("mssql"))
+  src_mssql <- my_test_src()
 
   # create an object on the MSSQL-DB that can be learned
   if (!any(src_tbls(src_mssql) %>%
```
krlmlr, line 10: We should always `copy_dm_to()`.

tests/testthat/test-learn.R
```diff
@@ -1,49 +1,87 @@
...
+  ) %>% dm_select_tbl(
+    tf_1 = starts_with("tf_1"), tf_2 = starts_with("tf_2"), tf_3 = starts_with("tf_3"),
+    tf_4 = starts_with("tf_4"), tf_5 = starts_with("tf_5"), tf_6 = starts_with("tf_6"))
+
+  expect_equivalent_dm(
+    dm_for_filter_mssql_learned,
+    dm_for_filter()
+  )
+
+  walk(
```
krlmlr, line 30: This should be code in `on.exit()`, so that it is run also when the test aborts.

tests/testthat/test-learn.R
```diff
@@ -1,49 +1,87 @@
...
+})
+
+
+test_that("Learning from specific schema on MSSQL works?", {
+
+  skip_if_src_not("mssql")
+  src_mssql <- my_test_src()
+  con_mssql <- src_mssql$con
+
+  # this schema name should be special enough to avoid any conflicts
```
krlmlr, line 43: Can we use a randomized schema name?

R/learn.R
TSchiefer: I tried this and your next suggestion in all 3 combinations, but it won't work, not even with normal schema names. 
Probably the reason is that here we do not want to address the schema in a direct way, but we want to refer to the schema name representation in the result of the query. If we use `'%s'`, it is even possible to address schemas like "SELECT" or other keywords.

> Otherwise the SQL is bad and prone to injection if schema contains '.

Since I am not able to create a schema containing `'`, I unfortunately can't test this corner case.

R/learn.R
TSchiefer:
```suggestion
    gsub("'", "''", schema)
```
This way also schemas containing apostrophes are addressed correctly

R/db-interface.R
krlmlr: This means we no longer need this change?

R/learn.R
krlmlr: This is what `dbQuoteIdentifier()` is supposed to do.

R/learn.R
krlmlr: Maybe:

```suggestion
    unclass(DBI::dbQuoteIdentifier(con, schema))
```

? Let's not reimplement `dbQuoteIdentifier()` here.

TSchiefer: this doesn't work: what we want for a schema called `SELECT` is `'SELECT'`, and with `unclass(DBI::dbQuoteIdentifier(con, schema))` we get `\"SELECT\"`.

for a schema with name `123fn'321` we get `\"123fn'321\"` from `unclass(DBI::dbQuoteIdentifier(con, "123fn'321"))`, but what we want is `'123fn''321'`.

what do you suggest?

R/db-interface.R
TSchiefer: ok, let's properly implement in #432 and until then do as you suggested

tests/testthat/test-learn.R
TSchiefer: Good catch

TSchiefer:
After 0e9d1b4 tests for MSSQL are passing when I let them run in the console with `testthis:::test_this()`.

But I get a strange behavior locally when I let the whole test-suite run with `devtools::test()`: I get warnings, but no errors for the tests for `dm_learn_from_src()`. The warnings indicate, that the keys could not be queried, and it tries to include all tables seen by `src_tbls(src)`, which of course doesn't work, cause there are some tables which are located in schemas. So I don't understand two things:

1. why doesn't the query work that focuses on a certain schema (or 'dbo')
1. if 1. happens, why is there only a warning and not a failing test? (the proper `dm` hasn't been learned)

It seems to work for the tic-run: `db-mssql / r-release` though.

R/learn.R
krlmlr: I mean `dbQuoteString()` -- you're right that `dbQuoteIdentifier()` is incorrect here.

R/learn.R
```diff
@@ -119,9 +122,12 @@ mssql_learn_query <- function() { # taken directly from {datamodelr}
       and ind_col.column_id = cols.column_id
     left outer join sys.systypes [types] on
       types.xusertype = cols.system_type_id
+  where tabs.schema_id = schema_id('%s')
   order by
     tabs.create_date,
-    cols.column_id"
+    cols.column_id",
+  gsub("(^|[^'])(['])($|[^'])", "\\1\\2\\2\\3", schema)
```
krlmlr, line 129:
```suggestion
  DBI::dbQuoteString(con, schema)
```

TSchiefer: Ok, thanks, that looks way better :) Only thing is, that it also doubles `''`, so if the user tries to use their SQL knowledge, it will fail. But I think that's a minor issue.

tests/testthat/test-learn.R
TSchiefer: FYI: I used `withr::defer()` in d07251a and the equivalent commit for the other test because I found this in the web: https://testthat.r-lib.org/articles/test-fixtures.html

tests/testthat/test-learn.R
TSchiefer: would this be ok with you? 
715d93a and 1e835ab

TSchiefer:
> Looks good. What does it take to unify the tests for MSSQL and Postgres? The CREATE SCHEMA and DROP SCHEMA syntax seems standardized.

seems to work now with 706c432

TSchiefer:
maybe https://krlmlr.github.io/dm/articles/howto-dm-db.html#building-a-dm-from-a-subset-of-tables could be updated to include the possibility to learn from a schema on MSSQL and Postgres DBs? Here or in a future PR

tests/testthat/test-learn.R
```diff
@@ -1,74 +1,86 @@
...
-
-  expect_identical(
-    def_learned_renamed_reclassed,
-    def_original
+  dm_for_filter_copied <- copy_dm_to(src_db, dm_for_filter(), temporary = FALSE, table_names = ~ DBI::SQL(unique_db_table_name(.x)))
+  order_of_deletion <- c("tf_2", "tf_1", "tf_5", "tf_6", "tf_4", "tf_3")
+  remote_tbl_names <- map_chr(
+    order_of_deletion,
+    ~dbplyr::remote_name(dm_for_filter_copied[[.x]])
+  ) %>% set_names(order_of_deletion)
```
krlmlr, line 25:
```suggestion
    set_names(order_of_deletion),
    ~ dbplyr::remote_name(dm_for_filter_copied[[.x]])
  )
```

tests/testthat/test-learn.R
```diff
@@ -1,74 +1,86 @@
...
+  dm_for_filter_db_learned_all <- dm_from_src(src_db)
+
+  # in case there happen to be other tables in schema "dbo" or "public"
+  dm_for_filter_db_learned <-
+    dm_for_filter_db_learned_all %>%
+    dm_select_tbl(!!!remote_tbl_names)
+
+  expect_equivalent_dm(
+    dm_for_filter_db_learned,
+    dm_for_filter()[order_of_deletion]
```
krlmlr, line 43: Do we need to reorder?

```suggestion
    dm_for_filter()
```

tests/testthat/test-learn.R
```diff
@@ -1,74 +1,86 @@
...
+    table_names = ~ DBI::SQL(dbplyr::in_schema(schema_name_q, .x))
   )
+  order_of_deletion <- c("iris_3", "iris_2", "iris_1")
+  remote_tbl_names <- set_names(paste0(schema_name_q, order_of_deletion), order_of_deletion)
+
+  withr::defer(
+    {
+      walk(
+        remote_tbl_names,
+        ~ dbExecute(con_db, paste0("DROP TABLE ", .x))
```
krlmlr, line 71:
```suggestion
        ~ try(dbExecute(con_db, paste0("DROP TABLE ", .x)))
```

tests/testthat/test-learn.R
```diff
@@ -1,74 +1,86 @@
...
+      )
+      dbExecute(con_db, paste0("DROP SCHEMA ", schema_name_q))
```
krlmlr, line 73:
```suggestion
      try(dbExecute(con_db, paste0("DROP SCHEMA ", schema_name_q)))
```

tests/testthat/test-learn.R
```diff
@@ -1,74 +1,86 @@
...
+  order_of_deletion <- c("tf_2", "tf_1", "tf_5", "tf_6", "tf_4", "tf_3")
+  remote_tbl_names <- map_chr(
+    order_of_deletion,
+    ~dbplyr::remote_name(dm_for_filter_copied[[.x]])
+  ) %>% set_names(order_of_deletion)
+
+  withr::defer(
+    walk(
+      dm_get_tables_impl(dm_for_filter_copied)[order_of_deletion],
+      ~ dbExecute(src_db$con, paste0("DROP TABLE ", dbplyr::remote_name(.x)))
```
krlmlr, line 30:
```suggestion
      ~ try(dbExecute(src_db$con, paste0("DROP TABLE ", dbplyr::remote_name(.x))))
```

tests/testthat/test-learn.R
```diff
@@ -1,74 +1,86 @@
...
+  )
+
+  dm_for_filter_db_learned <- dm_from_src(src_db, schema = schema_name) %>%
+    dm_select_tbl(!!!order_of_deletion)
 
   expect_equivalent_dm(
-    dm_postgres_learned_from_con_renamed,
-    dm_for_filter()
+    dm_for_filter_db_learned,
+    dm_for_disambiguate()[order_of_deletion]
```
krlmlr, line 82: Why is `dm_for_filter_...` compared with `dm_for_disambiguate()` ? Also:

```suggestion
    dm_for_disambiguate()
```

tests/testthat/test-learn.R
TSchiefer: either reorder, or change the implementation of `expect_equivalent_dm()`; it fails currently, if the order of the tables in the `dm`-objects differ.

tests/testthat/test-learn.R
TSchiefer: good catch :) I must have just copied & pasted (in the end it was a `dm_for_disambiguate` in disguise).
Don't know about reordering, see above.

TSchiefer:
Think I addressed your points from the review apart from the reordering. Please feel free to solve this however you want or let me know.

TSchiefer:
> I wonder if we should define a random seed for each test file?

I am not sure I understand how you'd like to implement that. Would the schema name then not be the same each time and therefore defy the reason for the function `random_schema()`?

krlmlr:
Thanks!

PR #436, opened 2020-08-18
raw tests for compound keys

### LINKED ISSUE #343: Understand limitations of current implementation
when compound keys exist in the data model.

On top of #335, or as part of it.

- [x] Implement `compound = FALSE` argument to `dm_nycflights13()`.
- [x] Walk through all functions in https://krlmlr.github.io/dm/dev/reference/. For each function:
    - find an example that uses `dm_nycflights13(compound = TRUE)`
    - add a `verify_output()` test to the bottom of the relevant `tests/testthat/test-*.R` file
        - naked, do not wrap into `test_that("...", {})`
    - do not attempt to fix

TSchiefer:
- arg `compound = FALSE` for dm_nycflights13()
- tests for all functions using dm_nycflights13(compound = TRUE)

closes #343


krlmlr:
Thanks. Checking is not yet supported for `rows_update()` and `rows_insert()`.

What do you think about the following order of priorities? The basic tests look like a minimum viable product that could be merged into the main branch, before we take on the database and checking stuff?

## Basic

1. test-primary-keys.R (`dm_get_all_pks()` doesn't properly return composite keys)
1. test-foreign-keys.R (`dm_get_all_fks()` doesn't properly return composite keys)
1. test-zoom.R (`dm_update_zoomed()` works, but `dm_insert_zoomed()` doesn't when the zoomed table had a composite foreign key, it silently drops the key)
1. test-paste.R (`dm_paste()` needs to be tweaked)
1. test-examine-constraints.R (currently doesn't work, cause it doesn't realize 2 columns are involved here)
1. test-rows-dm.R (composite keys are not yet supported)

## Database

1. test-dm-from-src-R (both `copy_dm_to()` and `dm_from_src()` (also specifically `dm_learn_from_db()`) cannot deal with compound keys yet)
1. test-db-interface.R (for DB tests `copy_dm_to()` is used anyway, so there might not be a need for another, specific test here?)
1. test-learn.R (cf. test-dm-from-src.R)

## Checking

1. test-check-cardinalities.R (implementation needs to be changed to support compound keys)
1. test-key-helpers.R (should `check_subset()` and `check_set_equality()` also work for multiple columns?)

## Obsolete

1. test-dplyr.R (looks good in general, but `slice.zoomed_dm()` with `.keep_pk = FALSE` cannot deal with compound keys ATM; apart from that all methods work, including joins)
    - Get rid of `.keep_pk`
1. test-nest.R (`nest_join_zoomed_dm()` cannot deal with composite keys yet)
    - Rethink nesting (?)

krlmlr:
All FIXMEs related to compound keys should now have the form "FIXME: COMPOUND:" .

krlmlr:
Thanks!

PR #444, opened 2020-09-14
dm_get_all_column_types helper function for drawing

samssann:
Using `dm_draw(..., col_attr = c("column", "type"))` now produces a graph which includes the class/type of each column. Before this PR, all column types were labeled as "integer".

image

Sometimes a column can have more than one class. For now, a simple method is used where the first class is chosen: `first_class <- function(x) class(x)[1]`


krlmlr:
Thanks, I love it! I tweaked a bit (so that it also works for dm objects on a database and that types are only queried when needed). Would you like to tweak the test (that now fails) and also contribute a visual test in `test-draw-dm.R` ?

samssann:
Thank you. I will try to get update the tests later this week.

krlmlr:
Are you still interested in updating the tests? Please merge master to start a new test run.

krlmlr:
Thanks!

PR #472, opened 2021-01-08
argument 'dbname' for 'dm_from_src()' for MSSQL

TSchiefer:
Access different databases on the DB-server.

- works exclusively for MSSQL
- warns, if argument is used for other sources
- works for both learning with keys and without

Testing this feature would only be possible if we can create and drop a test-DB. Can we do this?

Some tests for more specialized functions are implemented.


krlmlr (commented):
Thanks, looks good!

R/learn.R
```diff
@@ -52,6 +52,8 @@ dm_learn_from_db <- function(dest, ...) {
     return()
   }
 
+  dbname <- list(...)$dbname
```
krlmlr, line 55: We could pull out the `dbname` argument explicitly, need to pass it with the ellipse in `db_learn_query()` .

R/db-helpers.R
```diff
@@ -167,43 +167,56 @@ repair_table_names_for_db <- function(table_names, temporary, con) {
...
+    if (is_null(dbname)) {
+      dbname <- ""
+      dbname_sql <- ""
+    } else {
+      dbname_sql <- paste0(DBI::dbQuoteIdentifier(con, dbname), ".")
+    }
     names_table <- DBI::dbGetQuery(
       con,
-      "SELECT table_schema as schema_name, table_name as table_name from information_schema.tables"
+      glue::glue("SELECT tabs.name AS table_name, schemas.name AS schema_name
```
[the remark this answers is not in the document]
TSchiefer: makes problems when `dbname_sql` is `""`

R/learn.R
TSchiefer: Sorry for the late reply.
I don't really understand what you mean by "pulling out".

There are currently 3 different arguments that can be given in the ellipsis: `dbname`, `schema` and `table_type`.

How (mabye `list(...)$dbname` isn't so nice?) and when should I access them?

krlmlr: `dm_learn_from_db <- function(dest, dbname, ...)` and pass it explicitly to other functions.

TSchiefer:
I addressed your comments & also tests are now in place:
- learning from another DB from default schema
- learning from another DB from specific schema

Seems ready for merging to me.

R/dm-from-src.R
```diff
@@ -20,10 +20,12 @@
...
 #'   Additional parameters for the schema learning query.
-#'   Currently supports `schema` (default: `"public"`)
-#'   and `table_type` (default: `"BASE TABLE"`) for Postgres databases.
 #'
-#'   For MSSQL databases `schema` (default: `"dbo"`) is supported.
+#'   - `schema`: supported for MSSQL (default: `"dbo"`) and Postgres (default: `"public"`).
+#'   Learn the tables in a specific schema.
+#'   - `dbname`: supported for MSSQL. Access different databases on the connected MSSQL-server;
+#'   default: database addressed by `src`.
+#'   - `table_type`: supported for Postgres (default: `"BASE TABLE"`). Specify the table type.
```
krlmlr, line 28: What are other possible options here?

R/dm-from-src.R
```diff
@@ -20,10 +20,12 @@
...
 #'
 #'   Additional parameters for the schema learning query.
-#'   Currently supports `schema` (default: `"public"`)
-#'   and `table_type` (default: `"BASE TABLE"`) for Postgres databases.
 #'
-#'   For MSSQL databases `schema` (default: `"dbo"`) is supported.
+#'   - `schema`: supported for MSSQL (default: `"dbo"`) and Postgres (default: `"public"`).
+#'   Learn the tables in a specific schema.
+#'   - `dbname`: supported for MSSQL. Access different databases on the connected MSSQL-server;
+#'   default: database addressed by `src`.
```
krlmlr, line 27:
```suggestion
#'   default: current database.
```

krlmlr (commented):
Thanks, let's do another round. Can you please merge the main branch before pushing, to get rid of the codecov annotations?

R/db-helpers.R
```diff
@@ -167,43 +167,56 @@ repair_table_names_for_db <- function(table_names, temporary, con) {
...
+      # for Postgres:
+      dbname <- NULL
+    }
+    if (!is_postgres(src)) {
+      # any DBMS different from MSSQL or Postgres
+      if (!is_null(schema)) {
+        warn("Argument `schema` ignored: currently only supported for MSSQL and Postgres")
+      }
+      # in any case return `src_tbls(src)` if not MSSQL or Postgres
       return(src_tbls(src))
```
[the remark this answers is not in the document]
TSchiefer: Alright, I can do that, it won't change the behavior. 
Only thing is, that then we need to have the same warning about a non-NULL parameter `dbname` twice in the code: once for `if (!is_mssql(src) && !is_postgres(src))` and once for `if (!is_postgres(src))`. I wanted to avoid that.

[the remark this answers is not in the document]
krlmlr: We can extract a function that issues a warning if the argument is non-`NULL` to avoid repetition:

```r
warn_if_non_null <- function(arg) {
  arg_expr <- enexpr(arg)
  if (is.null(arg)) return()
  ...
}
```

R/learn.R
```diff
@@ -78,55 +78,79 @@ dm_learn_from_db <- function(dest, ...) {
   legacy_new_dm(tables, data_model)
 }
 
-schema_if <- function(schema, table, con) {
+schema_if <- function(schema, table, con, dbname = NULL) {
   table_sql <- DBI::dbQuoteIdentifier(con, table)
-  if_else(is.na(schema), table_sql, paste0(DBI::dbQuoteIdentifier(con, schema), ".", table_sql))
+  if (is_null(dbname) || dbname == "") {
+    if_else(
```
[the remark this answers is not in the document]
TSchiefer: It is used in the function `get_src_tbl_names()` in a `mutate()`-call for a table with columns `table_name` and `schema_name`, so it comes as a vector. But we should only support one unique value for `schema`. I should probably explicitly check for that, what do you think?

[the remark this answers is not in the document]
krlmlr: Checking uniqueness should be the responsibility of the caller, here we should accept only strings (and maybe check that it is a string).

R/dm-from-src.R
TSchiefer: we are currently only supporting those 3. Anything else in the ellipsis will be ignored.

krlmlr: What options other than `"BASE TABLE"` exist?

TSchiefer:
I now extracted a `warn`-function (lives in error-helpers.R) and 2 `check`-functions (live in test-dm.R; is this correct?) of general purpose.

If you think these functions make sense I will add tests for them.

Also I made the logic in `get_src_tbl_names()` more straightforward.

unclear why both codecov-tests fail.

krlmlr:
I have enabled coverage checks for the database runs, this does boost our code coverage stats and also indicates that not all of the new code is covered by the tests. You can look at the details to see what lines in the patch were not covered, and add tests as necessary.

TSchiefer:
Nice, thanks.

When I click on details I get to a login page: https://about.codecov.io/
When I log in with my github account, I only see the DBI repos and can't access details to the {dm}-runs.

R/db-helpers.R
```diff
@@ -167,43 +167,55 @@ repair_table_names_for_db <- function(table_names, temporary, con) {
...
+  if (!is_mssql(src) && !is_postgres(src)) {
+    warn_if_not_null(schema)
+    warn_if_not_null(dbname, only_on = "MSSQL")
     return(src_tbls(src))
   }
 
-  # src is now either Postgres or MSSQL, schema is not `NULL`
+  con <- src$con
+
   if (is_mssql(src)) {
```
krlmlr, line 179: Can you please extract functions for the code in the `if {}`, and also be explicit about the database in the check (if MSSQL, if Postgres)?

R/learn.R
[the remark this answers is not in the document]
krlmlr: Let's keep this as is.

krlmlr (commented):
Thanks, looks good! If you add tests for the new helpers, coverage should go up.

TSchiefer:
I'm oddly proud to have achieved 100% codecov of the diff ;)

krlmlr:
Thanks!

PR #478, opened 2021-02-03
Update financial-db-con.R

pachamaltese:
the certificate expired, no way to renew easily, so I switched db-edu -> for databases


krlmlr:
Thanks!

PR #482, opened 2021-02-16
Methods for dealing with schemas (for Postgres and MSSQL)

### LINKED ISSUE #23: Schema support
Required for copying to and learning from a database.

Options:

1. Rely on "default schema" (e.g. Oracle) or "search path" (e.g. in Postgres) -- many but not all DBMS know this concept
2. Pass in an optional schema qualifier in copy/learn operations

The first option is simpler, but it looks like we need to take care of the schema identifier ourselves.

Let's add a `schema` argument which takes a `DBI::Id` object.

Ideally, `cdm_learn_from_db()` would be part of the `dm()` constructor -- later.

TSchiefer:
- `sql_schema_list(dest, include_default = TRUE, ...)`: lists all schemas on the DB and their owners in a tibble
- `sql_schema_exists(dest, schema, ...)`: checks if a schema exists
~~- `sql_schema_table_list(dest, schema, ...)`: lists the tables in a schema; errors if it doesn't exists~~
- `sql_schema_create(dest, schema, ...)`: creates a schema on the DB; errors if such a schema already exists
- `sql_schema_drop(dest, schema, ...)`: drops a schema on the DB; errors if no such a schema exists

All methods for MSSQL have an additional argument `dbname`.

Only methods for MSSQL and Postgres (and SQLite: errors) are implemented in the PR.

closes #23


TSchiefer:
Is it maybe too early to use R 4.0.4 for the tests?

krlmlr:
What changes in R 4.0.4?

TSchiefer:
Nvm, sorry, yesterday's tests failed when trying to install R 4.0.4, so even before doing anything. Now it seems to work.

I am trying to fix the pkgdown issue and then I'll let you know, ok?

R/error-helpers.R
```diff
@@ -582,3 +582,56 @@ warn_if_not_null <- function(arg, arg_name = deparse(substitute(arg)), only_on =
...
+
+abort_schema_exists <- function(schema, dbname = NULL) {
+  abort(error_txt_schema_exists(schema, dbname),
+    .subclass = dm_error_full("schema_exists")
+  )
+}
+
+error_txt_schema_exists <- function(schema, dbname) {
+  if (!is_null(dbname)) {
+    msg_suffix <- paste0(" on database ", tick(dbname))
```
krlmlr, line 596: Should this be a function?

TSchiefer:
Hopefully now it's ready for review.

You mentioned in #23 methods `sql_db_*()`. I could in this PR also add methods for MSSQL:
- `sql_db_create()`
- `sql_db_drop()`
- `sql_db_use()`

Also, I was wondering if it's better to have explicit error messages for e.g. SQLite that state that it doesn't support schemas, not sure.

R/schema.R
```diff
@@ -0,0 +1,301 @@
...
+  if (!is_null(schema)) {
+    check_param_class(schema, "character")
+    check_param_length(schema)
+  }
+  if (!is_null(schema) && !sql_schema_exists(dest, schema, ...)) {abort_no_schema_exists(schema, ...)}
+  UseMethod("sql_schema_table_list")
+}
+
+#' @export
+`sql_schema_table_list.src_Microsoft SQL Server` <- function(dest, schema = NULL, dbname = NULL, ...) {
```
krlmlr, line 206: Do we need the `src_` methods? Either way, we should implement methods without `src_` and if necessary forward the `src_` methods.

Also, listing tables (in a schema) should maybe be part of dbplyr?

R/schema.R
```diff
@@ -0,0 +1,301 @@
...
+#'
+#' @return `NULL` invisibly.
+#'
+#' @family schema handling functions
+#' @export
+sql_schema_drop <- function(dest, schema, ...) {
+  check_param_class(schema, "character")
+  check_param_length(schema)
+  if (nrow(sql_schema_table_list(dest, schema, ...)) > 0) {
+    abort_schema_not_empty(schema, ...)
```
krlmlr, line 269: I wouldn't mind getting the error message straight from the database here. Some servers offer a `CASCADE` option that cleans all contents, we could match that to a `force` argument.

krlmlr (commented):
Thanks, looks good. When this is ready we could think about contributing to dbplyr too.

krlmlr:
Let's postpone `sql_db_*()` . Explicit SQLite methods sound good.

R/schema.R
TSchiefer: yes, I was also thinking of special scenarios: maybe some DBMS allow for schemas without tables but with views, which would lead to an error when trying to drop the schema. Gonna remove this error and add the force argument.

R/schema.R
TSchiefer: > [...] we should implement methods without src_ and if necessary forward the src_ methods.

I did it this way with the exception of `sql_schema_table_list()`, where I implemented the `src_`-methods, since we had a handy function for this purpose readily available. I think it's just slightly more comfortable for users not to have to worry, if it's a `src` or a `con` object.

Shall we make a PR for listing the tables in dbplyr? Is the output table something you consider useful? (tibble with 2 columns: table_name and remote_name). In dbplyr, should it be rather a `src`-method than a `con`-method?

TSchiefer:
Update: 
- One thing that I had forgotten about before is the quoting of identifiers. Now it's taken care of and it seems fine to me now, but I need to add a few specific tests in that respect. 
- Also tests are missing for the `force` parameter for Postgres.
- tests for `sql_schema_list()` are actually also still missing

R/schema.R
TSchiefer: So I will not export `sql_schema_table_list()` and just use 2 internal functions for now, `sql_schema_table_list_mssql()` and `sql_schema_table_list_postgres()`

TSchiefer:
@krlmlr After a few times breaking and repairing to me it looks ok now.

Remains the question, if we should implement the methods for any additional DBMS. I read that for MariaDB it is kind of similar to SQLite, that there is no schema support, but you need to create a new database if you wanna have a separation of context for your tables. How to handle this? Like for SQLite?
What about other DBMS?

tests/testthat/test-schema.R
```diff
@@ -0,0 +1,234 @@
...
+    table_name == "test_schema_2"
+  ) %>%
+    pull(remote_name)
+  expect_identical(
+    tbl(src_db, remote_table_2) %>% collect(),
+    tibble(b = letters[1:5])
+  )
+})
+
+test_that("schema handling on MSSQL works for different DBs", {
```
krlmlr, line 92:
```suggestion
test_that("schema handling on Postgres works for different DBs", {
```

tests/testthat/test-schema.R
```diff
@@ -0,0 +1,234 @@
...
+
+  expect_dm_error(sql_schema_exists(con_db, 1), "parameter_not_correct_class")
+  expect_dm_error(sql_schema_exists(con_db, letters[1:2]), "parameter_not_correct_length")
+  expect_dm_error(sql_schema_exists(src_db, 1), "parameter_not_correct_class")
+  expect_dm_error(sql_schema_exists(src_db, letters[1:2]), "parameter_not_correct_length")
+
+  withr::defer({
+    try(dbExecute(con_db, "DROP TABLE test_schema_1"))
+    try(dbExecute(con_db, SQL("DROP TABLE \"1-dm_schema_TEST\".\"test_schema_2\"")))
+    try(dbExecute(con_db, SQL("DROP SCHEMA \"1-dm_schema_TEST\"")))
```
krlmlr, line 20:
```suggestion
    try(dbExecute(con_db, SQL('DROP TABLE "1-dm_schema_TEST"."test_schema_2"')))
    try(dbExecute(con_db, SQL('DROP SCHEMA "1-dm_schema_TEST"')))
```

tests/testthat/test-learn.R
```diff
@@ -157,19 +155,25 @@ test_that("'schema_if()' works", {
...
   )
 
   # schema and table set
   expect_length(
-    pluck(strsplit(expect_s4_class(
-      schema_if(schema = "schema", table = "table", con = con_db),
-      "SQL"), "\\."),
-      1),
+    pluck(
+      strsplit(expect_s4_class(
```
krlmlr, line 171: This looks a tad too complex. Maybe extract a variable?

krlmlr (commented):
Thanks, looks good! Can you please apply/tweak and merge?

krlmlr:
Remember to use a bullet in the merge commit message, like:

```
- New `sql_schema_...()` for handling with schemas, implemented for PostgreSQL and SQL Server (#23).
```

PR #484, opened 2021-02-26
argument 'schema' for 'copy_dm_to()'

### LINKED ISSUE #432: 'copy_dm_to()' to schema
cf. https://github.com/krlmlr/dm/pull/428#issuecomment-666305219 and af980c2 (this failed for both MSSQL and Postgres)

For me it seems the right way to provide an additional argument `schema`, that will make it clear, that this should be quoted independently of the `table_name`. 
But I am open to other ideas.

TSchiefer:
- schema argument for `copy_dm_to()` (more accessible way than setting `table_names` appropriately)
- if `schema != NULL`: error, if `temporary = TRUE`; error, if `table_names != NULL`
- tests are in place

closes #432


krlmlr:
Thanks!

PR #485, opened 2021-03-03
read/write dm as csv/zip(csv)/xlsx

### LINKED ISSUE #276: Store `dm` as xlsx or collection of csv files (zip)
...and read it again.

- xlsx: one table per sheet plus one extra sheet with the metadata
- csv: one table per file plus one extra file with the metadata(?)

CC @christophsax

TSchiefer:
- `dm_write_csv(dm, csv_directory)`: write `dm` as collection of `csv`-files
- `dm_read_csv(csv_directory)`: read `dm` from directory created using `dm_write_csv()`
- `dm_write_zip(dm, zip_file_path = "dm.zip", overwrite = FALSE)`: same as csv, but zipped.
- `dm_read_zip(zip_file_path)`
- `dm_write_xlsx(dm, xlsx_file_path = "dm.xlsx", overwrite = FALSE)`
- `dm_read_xlsx(xlsx_file_path)`

I am prepared for a longer wait until the conclusion of this PR, since there might be a few things to discuss.

For example: 
- should we already prepare for compound keys.
- will it work for all (important) column classes? what will be the default if it doesn't work?
- could there be some way to support writing remote `dm` to a file/directory?

closes #276


krlmlr:
Nice! To the questions:

1. Yes, absolutely.
2. Failure, perhaps with extensibility option later.
3. Looks very niche -- if it fit into RAM, we can `collect()`, if not then what's the purpose to save as CSV?

We should also integrate with {shard}.

R/read-write.R
```diff
@@ -0,0 +1,428 @@
...
+#' @inheritParams dm_add_pk
+#' @param csv_directory For `dm_write_csv()`: The path to a non-existent or empty
+#' directory to write the `csv`-files defining the `dm` to.
+#'
+#' For `dm_read_csv()`: The path to the directory containing the `dm` as `csv-files`.
+#' @details `dm_write_csv()`: write a `dm` to a collection of `csv`-files to a non-existent or empty directory.
+#'
+#' @return `read`-family: A `dm` object.
+#'
+#' `write`-family: The path to the file/directory containing the stored `dm`, invisibly.
```
krlmlr, line 17: Should return the input dm?

R/read-write.R
```diff
@@ -0,0 +1,428 @@
...
+      mutate(data, !!column := get(paste0("as.", class))(!!sym(column)))
+    }, .init = data)
+    }) %>%
+    set_names(table_names) %>%
+    enframe(name = "table", value = "data")
+
+  make_dm(def_base, table_names, table_tibble_reclassed, pk_info, fk_info)
+}
+
+make_dm <- function(def_base, table_names, table_tibble, pk_info, fk_info) {
```
krlmlr, line 268: This looks similar to the older `new_dm*()` functions, any overlap?

R/read-write.R
```diff
@@ -0,0 +1,428 @@
...
+      "POSIXct" %in% col_class_tibble$class |
+      "POSIXlt" %in% col_class_tibble$class) {
+    tbl_dt_list <- filter(
+      col_class_tibble,
+      class == "datetime" |
+        class == "POSIXct" |
+        class == "POSIXlt"
+    )
+    message(
+      glue::glue(
```
krlmlr, line 340: The conversion to UTC is lossless, we can convert ourselves (and print the timestamp as `'....+00:00'` or `'...Z'`.

R/read-write.R
```diff
@@ -0,0 +1,428 @@
...
+    left_join(fk_tibble, by = c("table" = "child_table")) %>%
+    mutate(fks = map(fks, ~ if (is_null(.x)) {new_fk()} else {.x})) %>%
+    mutate(fks = vctrs::as_list_of(fks)) %>%
+    left_join(filter_tibble, by = "table") %>%
+    left_join(zoom, by = "table") %>%
+    left_join(col_tracker_zoom, by = "table") %>%
+    new_dm3()
+}
+
+prepare_tables <- function(dm) {
```
krlmlr, line 313: The function name doesn't seem specific enough.

R/read-write.R
```diff
@@ -0,0 +1,428 @@
...
+#' @inheritParams dm_write_csv
+#' @param zip_file_path
+#' For `dm_write_zip`: The file path to the `zip`-file to write; defaults to `dm.zip`.
+#'
+#' For `dm_read_zip`: The file path to the `zip`-file to read the `dm` from.
+#' @param overwrite Logical, default: `FALSE`. In case the file already exists, should it be overwritten?
+#' @details `dm_write_zip()`: write a `dm` to a `zip`-file containing a collection of `csv`-files.
+#' @rdname dm-read-write
+#' @export
+dm_write_zip <- function(dm, zip_file_path = "dm.zip", overwrite = FALSE) {
```
krlmlr, line 117: Not sure I like the default file name.

R/read-write.R
TSchiefer: I wouldn't mind that, and probably that makes more sense.
I somehow liked that it's pipeable (e.g. `dm_write_csv(dm_for_filter(), csv_directory = test_path) %>% dm_read_csv()`), but you're right, this will be rarely a use-case :)

I will change it to return the `dm`

krlmlr: Yeah, this is also consistent with what `readr::write_csv()` does.

It's more useful to return the dm because writing to a file is a side effect, and the path can't really be computed on.

R/read-write.R
TSchiefer: I had a bet with myself, that you would wonder about that. 
I did wonder, too, what's the best way to recreate the `dm`, but just came up with two options:
1. the one that I chose
2. using `dm()` on the list of tables (after repairing the classes) and then using the metainfo to set the key-constraints on the raw `dm`.

On a different note: I am quite sure, that the current implementation can't work with compound keys. E.g., how could the foreign key column know, which primary key column it should refer to? 
Maybe I should have a look into the compound key PR first.

R/read-write.R
TSchiefer: Why? or to ask differently, do you have another name in mind? I guess the same is valid for `dm_write_xlsx(dm, xlsx_file_path = "dm.xlsx")`?

R/read-write.R
krlmlr: If we still allow only one primary key per table? I think it might just work.

The implementation is fine, let's make sure there's no overlap with existing code.

R/read-write.R
krlmlr: Do we even need a default?

TSchiefer:
> We should also integrate with {shard}.

Excuse my ignorance, but is what you mean [this](https://www.rdocumentation.org/packages/disk.frame/versions/0.3.7/topics/shard)?
If yes, maybe in another PR?

R/read-write.R
TSchiefer: we don't really, removing.

R/read-write.R
TSchiefer: ok, nice.

I didn't see overlap with existing code.

TSchiefer:
@krlmlr think I took all of your comments into account.

maybe a new issue for {shard}-integration?

krlmlr:
This looks good in general. I'll resolve conflicts and test-drive locally.

R/error-helpers.R
```diff
@@ -646,3 +646,76 @@ abort_one_of_schema_table_names <- function() {
     .subclass = dm_error_full("one_of_schema_table_names")
   )
 }
+
+
+# write or read dm to/from files ------------------------------------------
+
+abort_dir_not_empty <- function() {
+  abort(
+    "Please choose a non-existent or empty directory for the csv-files.",
```
krlmlr, line 655: Should we drop the dash here and elsewhere?

```suggestion
    "Please choose a non-existent or empty directory for the csv files.",
```

R/read-write.R
```diff
@@ -0,0 +1,465 @@
...
+  if (dir.exists(csv_directory) && length(list.files(csv_directory)) > 0) {
+    abort_dir_not_empty()
+  }
+  tryCatch(
+    dm_write_csv_impl(dm, csv_directory, zip = FALSE),
+    error = function(e) {
+      # remove directory in case it was created but there was an error
+      try(unlink(csv_directory, recursive = TRUE))
+      # keep error message and class though
+      abort(conditionMessage(e), .subclass = class(e))
```
krlmlr, line 50: The following will retain all other attributes:

```suggestion
      stop(e)
```

R/read-write.R
```diff
@@ -0,0 +1,465 @@
...
+    # empty subset is necessary because
+    # https://www.tidyverse.org/blog/2018/12/readr-1-3-1/#tibble-subclass
+    mutate(data = map(data, `[`))
+
+  make_dm(def_base, table_tibble, pk_info, fk_info)
+}
+
+#' @inheritParams dm_write_csv
+#' @param zip_file_path
+#' For `dm_write_zip`: The file path to the `zip`-file to write.
```
krlmlr, line 122: Here and elsewhere:

```suggestion
#' For `dm_write_zip()`: The file path to the `zip`-file to write.
```

R/read-write.R
```diff
@@ -0,0 +1,465 @@
...
+
+  csv_directory <- file.path(tempdir(), basename(tempfile(pattern = "dm_zip_")))
+  dm_write_csv_impl(dm, csv_directory, zip = TRUE)
+
+  withr::defer(try(unlink(csv_directory, recursive = TRUE)))
+
+  csv_files <- list.files(csv_directory)
+  # compress the file ("-j" junks the path to the file)
+
+  zip(
```
krlmlr, line 146: I suspect {zip} might work better here: https://cran.r-project.org/web/packages/zip/index.html.

R/read-write.R
```diff
@@ -0,0 +1,465 @@
...
+#'
+#' For `dm_read_xlsx()`: The file path to the `xlsx`-file to read the `dm` from.
+#' @details `dm_write_xlsx()`: write a `dm` to an `xlsx`-file containing several tables defining the `dm`.
+#' @rdname dm-read-write
+#' @export
+dm_write_xlsx <- function(dm, xlsx_file_path, overwrite = FALSE) {
+
+  if (file.exists(xlsx_file_path)) {
+    if (overwrite) {
+      message(glue::glue("Overwriting file {tick(xlsx_file_path)}."))
```
krlmlr, line 179: How can we mute this message?

R/read-write.R
```diff
@@ -0,0 +1,465 @@
...
+  "logical",
+  "integer",
+  "numeric",
+  "character",
+  "Date",
+  "POSIXlt",
+  "POSIXct"
+)
+
+convert_all_times_to_utc <- function(table_list, col_class_table) {
```
krlmlr, line 447: Is this within the scope of the function? Will it affect the roundtrip, or only the snapshot tests, if we omit UTC conversion?

R/read-write.R
```diff
@@ -0,0 +1,465 @@
...
+  if (any(col_class_table$class %in% c("POSIXlt", "POSIXct"))) {
+    to_convert <- filter(col_class_table, class %in% c("POSIXlt", "POSIXct"))
+    message(
+      c("Converting the datetime values for the following column(s) to timezone `UTC`:\n",
+        glue::glue("{paste0(to_convert$table, '$', to_convert$column, collapse = '\n')}"))
+    )
+    table_list <- reduce2(to_convert$table, to_convert$column, function(tables, table, column) {
```
krlmlr, line 454: `mutate(across())` ?

R/read-write.R
```diff
@@ -0,0 +1,465 @@
...
+  "integer",
+  "numeric",
+  "character",
+  "Date",
+  "POSIXlt",
+  "POSIXct"
+)
+
+convert_all_times_to_utc <- function(table_list, col_class_table) {
+  if (any(col_class_table$class %in% c("POSIXlt", "POSIXct"))) {
```
krlmlr, line 448: It looks like this won't pick up subclasses of `"POSIXlt"` or `"POSIXct"`? Not sure if this is relevant though.

krlmlr (commented):
We now have compound keys, need to adapt.

I'd prefer using existing methods for enumerating primary and foreign keys.

krlmlr:
We also want `check_suggested()` from #572.

R/read-write.R
TSchiefer: Is it necessary to be able to mute this? i.e., how often will users recreate/overwrite files? (honestly not sure)
We could add a `quiet` argument

R/read-write.R
TSchiefer: as it is implemented, just the following types are supported: `character`, `Date`, `integer`, `logical`, `numeric`, `POSIXct`, `POSIXlt`.
If if turns out that there are useful further classes that should be supported, I would suggest these should be added in future PRs.

R/read-write.R
TSchiefer: maybe if it works it's more platform agnostic, but in the first test (which works with `utils::zip()`), I get the error:
```r
Error in zip_internal(zipfile, files, recurse, compression_level, append = FALSE,  : 
  zip error: `Cannot add file `/var/folders/x3/ndmkxk1j2wn0mx2pw9v9httm0000gn/T//RtmpmKOSNA/dm_zip_1149248ed84b/___coltypes_file_dm.csv` to archive `___test_path/dm.zip`` in file `zip.c:348`
```
Not sure if it's worth the effort to try finding the source of this problem.

R/read-write.R
TSchiefer: maybe, but my thought was that it's good to inform users which columns are converted. And since then we know those columns, we can make those changes as well explicitly.

R/read-write.R
TSchiefer: Several problems:
1. unfortunately it's not possible to read timezones via `readr::read_csv()`. Therefore, it seems much safer to convert to UTC and to inform when writing the csv files.
1. for xlsx: actually `writexl::write_xlsx()` does the conversion to UTC quietly itself anyway. In this case we would not need to perform the conversion, but just inform the user.

I think it's not harmful to leave it as is, but I am open to suggestions.

TSchiefer:
Compound works now

TSchiefer:
> We also want `check_suggested()` from #572.

for the functions from {readr}, {readxl}, {writexl}? good idea. Even though there is some implementation of it in main, shall we wait for the merge?

R/read-write.R
TSchiefer: > 1. unfortunately it's not possible to read timezones via readr::read_csv(). Therefore, it seems much safer to convert to UTC and to inform when writing the csv file

I was too quick to claim that:
https://readr.tidyverse.org/articles/locales.html
It is actually possible to steer the timezone with the argument `locale` in `readr::read_csv()`. Not sure if making use of this possibility improves the transparency of our functions (mainly if it doesn't work with xlsx)

krlmlr:
This looks good, I'd like to play with it before merging. Is this blocking another project? If we had to choose between csv, zip and xlsx, what would be the preference?

TSchiefer:
> This looks good, I'd like to play with it before merging. Is this blocking another project? If we had to choose between csv, zip and xlsx, what would be the preference?

Sure, take your time. It's not blocking anything - at least not for me.
xlsx is nice, since it's just one file (as opposed to csv) and one can still easily get an overview of the contents (as opposed to zip). Obvious disadvantage is the requirement of MS Excel.

krlmlr:
This is mostly new code, could also be moved elsewhere.

krlmlr:
Closing for now.

PR #495, opened 2021-04-24
WIP: Add code to time each chunk

krlmlr:
so that we later can look for slow code.


krlmlr:
Not actionable, revisiting later.

PR #506, opened 2021-04-26
WIP: Use tidyselect for cardinality check

krlmlr:
Part of #3.


krlmlr:
We're good here now, already forwarding to `select()` .

Issue #342, opened 2020-05-01
Define data model for INFORMATION_SCHEMA


krlmlr:
The `INFORMATION_SCHEMA` is a standardized way for databases to provide information on tables, columns and keys. See https://dataedo.com/kb/databases/all/information_schema for details.

To better understand it, we should build a dm for it. This requires compound keys and needs to be done on top of #335, or as part of it.

- [ ] Build data model for `INFORMATION_SCHEMA` for MariaDB, as found in relational.fit
- [ ] Define a simpler "meta dm" derived from `INFORMATION_SCHEMA` that captures all necessary information (table definition, pk, fk)
    - what about color?
- [ ] Make compatible with Postgres
- [ ] Make compatible with SQL Server

krlmlr:
Diagram for `INFORMATION_SCHEMA`: http://mysql.xonu.de/INFORMATION_SCHEMA_5_1/

<details>
<summary>Details</summary>

![INFORMATION_SCHEMA_5_1](https://user-images.githubusercontent.com/1741643/116497829-8e483100-a8a8-11eb-9fad-f1f54a855e70.gif)

</details>

PR #517, opened 2021-04-29
ENH: New internal `dm_meta()` for learning a data model from the database, for now for SQL Server only (#342)

krlmlr:
Part of #342.


krlmlr:
I'll do another cleanup iteration and ping for review.

krlmlr:
Follow-up tasks, after this is merged:

- Record ptype of `dm_meta()` in a snapshot test
- Fix learning from Postgres
- Implement learning from MySQL
- Implement learning from SQLite (#352)
- Document and export `dm_meta()` (should this be a generic over the connection?)

krlmlr:
@TSchiefer: This is as good as it gets in this iteration. Let's file follow-up tickets. What do we need to change before merging?

R/global.R
```diff
@@ -105,6 +105,37 @@ utils::globalVariables(c(
...
+  "column_name",
+  "con",
+  "constraint_catalog",
+  "constraint_column_id",
+  "constraint_column_usage",
+  "constraint_name",
+  "constraint_schema",
+  "constraint_type",
+  "dbname",
+  "FIXME",
```
TSchiefer, line 122: Is this intentional?

R/learn.R
```diff
@@ -41,6 +41,130 @@ dm_learn_from_db <- function(dest, dbname = NULL, ...) {
...
+  dm_learn_from_db_meta(con, catalog = dbname, ...)
+}
+
+dm_learn_from_db_meta <- function(con, catalog = NULL, schema = NULL, name_format = "{table}") {
+  info <- dm_meta(con, catalog = catalog, schema = schema)
+
+  df_info <-
+    info %>%
+    dm_select_tbl(-schemata) %>%
+    collect()
```
TSchiefer, line 57: `collect()` is already done at the end of `dm_meta()`

R/meta.R
```diff
@@ -0,0 +1,145 @@
+dm_meta <- function(con, catalog = NA, schema = NULL) {
+  if (is_mssql(con)) {
+    if (is.null(catalog)) {
+      # FIXME: Classed error message?
+      abort("SQL server only supports learning from one database.")
```
TSchiefer, line 5: classed error msg in follow-up PR?

R/learn.R
```diff
@@ -41,6 +41,130 @@ dm_learn_from_db <- function(dest, dbname = NULL, ...) {
     return()
   }
 
+  if (!is_mssql(con)) {
```
TSchiefer, line 44: AFAICT, `dm_meta()` works well for PG too. We could learn from PG using the function as well in this PR.

TSchiefer (approved):
This is great.
I didn't go through every detail, but the behavior is nice.

Shall we also already use `dm_meta()` for learning from PG in this PR?
I tried it for both learning from a specific schema and without providing a schema name. It works great when providing a schema name, but I think we could filter out some or all of the non-user tables when not learning from a specific schema(?)

TSchiefer:
I forgot: should we also add a direct test for `dm_meta()` (and maybe even for some of its helper functions?)? In a future PR?

PR #567, opened 2021-05-25
Issue 566 testthat wrappers

### LINKED ISSUE #566: Fix compatibility with dev testthat
`expect_<condition>()` now returns the expectation in testthat edition 3 (development version), it previously returned the first argument. We need to rework quite a few tests.

moodymudskipper:
closes #566

We build wrappers by replacing the tunneled argument `{{object}}` with `obj <- rlang::eval_tidy({{object}})`, this way we don't execute object 2 times, as we would if we'd just have `object` as the last call of these functions.

We add a simple test for each of these wrappers.

We replace some instances of `expect_message` and `expect_warning` by their counterpart, so relevant tests are now passed.

### DIFF
--- a/tests/testthat/test-testthat-wrappers.R
+++ b/tests/testthat/test-testthat-wrappers.R
@@ -1,15 +1,30 @@
 
 testthat::test_that("testthat wrappers return the object", {
   expect_equal(
-    expect_message_obj({message("abc");"foo"}, "a"),
+    expect_message_obj(
+      {
+        message("abc")
+        "foo"
+      },
+      "a"),
     "foo"
   )
   expect_equal(
-    expect_warning_obj({warning("abc");"foo"}, "a"),
+    expect_warning_obj(
+      {
+        warning("abc")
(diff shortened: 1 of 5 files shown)


tests/testthat/test-testthat-wrappers.R
```diff
@@ -0,0 +1,15 @@
+
+testthat::test_that("testthat wrappers return the object", {
+  expect_equal(
+    expect_message_obj({message("abc");"foo"}, "a"),
```
krlmlr, line 4: These tests are implicit, not sure we need them.

We use styler to format the code, the result would look like this:

```suggestion
    expect_message_obj(
      {
        message("abc")
        "foo"
      },
      "a"
    ),
```

In particular:

- Each statement gets a separate line
- Multiline blocks are always indented

krlmlr (approved):
Thanks, this should do it for now.

krlmlr:
Thanks!

PR #568, opened 2021-05-25
Issue 262 pb dm examine constraints

### LINKED ISSUE #262: Add progress bar for dm_examine_constraints()
because this can take a while on a larger data model.

moodymudskipper:
closes #262

We add a dependency on {progress} package

We do some minor refactoring to loop only once in `check_pk_constraints`.

We add a progress bar for `check_pk_constraints` and `check_pk_constraints`.

It would be possible to have only one progress bar but we'd need deeper refactoring.

The following example prints as before because {progress} doesn't kick in for short operations, but if I insert artifical waiting times we get the following:

image


DESCRIPTION
```diff
@@ -49,6 +49,7 @@ Imports:
     magrittr,
     memoise,
     methods,
+    progress,
```
krlmlr, line 52: I wonder if we should add a `progress = NA` argument to the functions that implement a progress bar. This would allow moving the package to "Suggests". Behavior:

- If `NA`, hide in non-interactive mode, show in interactive mode, give a message if package not installed
- If `FALSE`, hide
- If `TRUE`, fail if package is not installed

What do you think?

krlmlr (commented):
Thanks, looks good!

R/examine-constraints.R
```diff
@@ -126,9 +133,17 @@ check_fk_constraints <- function(dm) {
   fks_tibble <-
     mutate(fks, t1 = cts, t2 = pts) %>%
     select(t1, t1_name = child_table, colname = child_fk_cols, t2, t2_name = parent_table, pk = parent_key_cols)
+
+  pb <- progress::progress_bar$new(
+    format = "  checking fk constraints [:bar] :percent in :elapsed",
+    total = nrow(fks_tibble), clear = FALSE, width = 60)
+
   fks_tibble %>%
```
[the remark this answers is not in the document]
moodymudskipper: I don't understand what you refer to

moodymudskipper:
I've implemented a `new_ticker` function, it works as you described.
It has a `progress = NA` argument to accommodate your last comment.
I've added a `progress = NA` argument to all functions that call directly or indirectly `new_ticker()`
An exception is the deprecated functions, that I thought we wouldn't want to alter, so I didn't touch their arguments and they use `progress = FALSE` so nothing changes.
{progress} was moved from Imports to Suggests
the progress argument was documented in `?dm_examine_constraints`

R/examine-constraints.R
```diff
@@ -100,13 +103,16 @@ check_pk_constraints <- function(dm) {
     ))
   }
   table_names <- pull(pks, table)
+  columns     <- pull(pks, pk_col)
```
krlmlr, line 106:
```suggestion
  table_names <- pks$table
  columns     <- pks$pk_col
```

R/examine-constraints.R
```diff
@@ -119,16 +125,19 @@ check_pk_constraints <- function(dm) {
...
-check_fk_constraints <- function(dm) {
+check_fk_constraints <- function(dm, progress = NA) {
   fks <- dm_get_all_fks_impl(dm)
   pts <- pull(fks, parent_table) %>% map(tbl_impl, dm = dm)
   cts <- pull(fks, child_table) %>% map(tbl_impl, dm = dm)
   fks_tibble <-
     mutate(fks, t1 = cts, t2 = pts) %>%
     select(t1, t1_name = child_table, colname = child_fk_cols, t2, t2_name = parent_table, pk = parent_key_cols)
+
+  ticker <- new_ticker("checking pk constraints", nrow(fks_tibble))
```
krlmlr, line 136: Same above.

```suggestion
  ticker <- new_ticker("checking pk constraints", nrow(fks_tibble), progress = progress)
```

R/zzx-deprecated.R
```diff
@@ -403,7 +403,7 @@ cdm_learn_from_db <- function(dest) {
 #' @export
 cdm_check_constraints <- function(dm) {
   deprecate_soft("0.1.0", "dm::cdm_check_constraints()", "dm::dm_examine_constraints()")
-  dm_examine_constraints_impl(dm = dm)
+  dm_examine_constraints_impl(dm = dm, progress = FALSE)
```
krlmlr, line 406: Do we still need this?

R/new_ticker.R
```diff
@@ -0,0 +1,33 @@
+new_ticker <- function(label, n, progress = NA) {
+  # if progress bar should'nt be shown, return identity
+  # message or fail depending on `progress`
+
+  if (is.na(progress)) {
+    progress <- interactive()
+    if (progress && !requireNamespace("progress")) {
+      message("The 'progress' package must be installed in order to display progress bars.")
```
krlmlr, line 8: This package uses rlang idioms where approriate.

```suggestion
    progress <- is_interactive()
    if (progress && !is_installed("progress")) {
      inform("The 'progress' package must be installed in order to display progress bars.")
```

krlmlr (commented):
Thanks, let's do one more iteration.

R/new_ticker.R
```diff
@@ -0,0 +1,33 @@
...
+      return(identity)
+    }
+  }
+
+  if (!progress) return(identity)
+
+  if (!requireNamespace("progress")) {
+    stop("The 'progress' package must be installed in order to display progress bars.")
```
krlmlr, line 16: Ultimately, this should be `abort()`. I wonder if we should implement

```r
check_suggested <- function(package, use, message) {
  if (!use) {
    return(FALSE)
  } else if (is.na(use)) {
    if (is_installed(package)) {
  ...
}
```

so that this pattern can be applied consistently and we can move more packages to "Suggests" if needed?

moodymudskipper: good idea, it's also more readable this way, I've implemented it

R/zzx-deprecated.R
moodymudskipper: Is it a question for me ? progress is NA by default, so by setting to FALSE I make sure old behavior is unchanged. Or do you mean to hard deprecate the function ?

R/new_ticker.R
moodymudskipper: Once we merge this we should use `check_suggested()` in `bdm_render_graph()` to replace :

```
  if (!rlang::is_installed("DiagrammeR")) {
    stop("DiagrammeR package needed for this function to work. Please install it.",
      call. = FALSE
    )
  }
```

R/zzx-deprecated.R
krlmlr: I'd expect that tests run in noninteractive mode, so this would not be necessary. But we can leave it as is.

krlmlr:
Thanks!

R/examine-constraints.R
```diff
@@ -119,16 +125,19 @@ check_pk_constraints <- function(dm) {
...
-check_fk_constraints <- function(dm) {
+check_fk_constraints <- function(dm, progress = NA) {
   fks <- dm_get_all_fks_impl(dm)
   pts <- pull(fks, parent_table) %>% map(tbl_impl, dm = dm)
   cts <- pull(fks, child_table) %>% map(tbl_impl, dm = dm)
   fks_tibble <-
     mutate(fks, t1 = cts, t2 = pts) %>%
     select(t1, t1_name = child_table, colname = child_fk_cols, t2, t2_name = parent_table, pk = parent_key_cols)
+
+  ticker <- new_ticker("checking pk constraints", nrow(fks_tibble), progress = progress)
```
krlmlr, line 136: This needs to be "fk constraints"? Should we spell out "primary key" and "foreign key"?

PR #570, opened 2021-05-26
Build expression instead of paste() + parse() in check_fk

moodymudskipper:


moodymudskipper:
On a closer look I think this expression might be overkill, we build an expression `is.na(value1) | is.na(value2) | ...` to be used in a filter call, but I think we can just do : 

    t1_join %>%
      na.omit() # or `%>% filter(complete.cases(.))`

instead of 

    t1_join %>%
      filter(!(!!any_value_na_expr))

since t1_join contains only `value*` columns and a `n` column that is never NA.

krlmlr:
Thanks! The pattern occurs in one more spot IIRC, do we need/want to extract a function?

PR #571, opened 2021-05-27
Issue 4 dm repair constraints

moodymudskipper:
This is a draft of `dm_repair_constraints`, discussed in #4. 

It checks only foreign keys so far.

*  `"add"` adds rows to the parent table with missing primary keys, it seems to work well.
*  `"delete"` should remove rows from the child table when foreign keys don't match primary keys, this doesn't work yet, as discussed in #4.

```
nycflights <- dm_nycflights13()

repaired_add <-
  nycflights %>%
  dm_repair_constraints("add")
#> Not persisting, use `in_place = FALSE` to turn off this message.

repaired_delete <-
  nycflights %>%
  dm_repair_constraints("delete")
#> Not persisting, use `in_place = FALSE` to turn off this message.

nrow(nycflights$planes)
#> [1] 945
nrow(nycflights$flights)
#> [1] 1761

# we have more planes, and the same amount of flights
nrow(repaired_add$planes)
#> [1] 1111
nrow(repaired_add$flights)
#> [1] 1761


# we (should) have the same amount of planes, and fewer flights
nrow(repaired_delete$planes)
#> [1] 945
nrow(repaired_delete$flights) # this is not right
#> [1] 1761
```

`check_fk()` now returns a list of a repair_plan" and a label, which are processed downstream


R/examine-constraints.R
```diff
@@ -33,13 +33,13 @@ dm_examine_constraints <- function(dm, progress = NA) {
   dm %>%
     dm_examine_constraints_impl(progress = progress) %>%
     rename(columns = column) %>%
-    mutate(columns = new_keys(columns)) %>%
+    mutate(columns = new_keys(columns), repair_plan = NULL) %>%
     new_dm_examine_constraints()
 }
 
-dm_examine_constraints_impl <- function(dm, progress = NA) {
+dm_examine_constraints_impl <- function(dm, progress = NA, fk_repair = "none") {
```
krlmlr, line 40:
```suggestion
dm_examine_constraints_impl <- function(dm, progress = NA, fk_repair = NULL, sample = TRUE) {
```

Could we use two arguments:

- `fk_repair` defines what kind of repair we consider: adding dimensions or pruning facts. Are there other options?
- `sample` defines if we're satisfied with a sample (current implementation) or if we want all instances (for cleanup)

R/examine-constraints.R
```diff
@@ -135,11 +135,20 @@ check_fk_constraints <- function(dm, progress = NA) {
 
   ticker <- new_ticker("checking pk constraints", nrow(fks_tibble), progress = progress)
 
+  problems <-
+    pmap(fks_tibble, ticker(check_fk), fk_repair = fk_repair) %>%
+    transpose()
```
krlmlr, line 140:
```suggestion
    pmap(fks_tibble, ticker(check_fk), fk_repair = fk_repair, sample = sample)
```

Can we derive the label from the repair plan here?

R/foreign-keys.R
```diff
@@ -520,34 +520,63 @@ check_fk <- function(t1, t1_name, colname, t2, t2_name, pk) {
...
     error = identity
   )
 
   # return error message if error occurred (possibly types didn't match etc.)
   if (is_condition(res_tbl)) {
-    return(conditionMessage(res_tbl))
+    return(list(repair_plan = NULL, label = conditionMessage(res_tbl)))
+  }
+
+  if(fk_repair == "none") {
```
krlmlr, line 532:
```suggestion
  if(sample) {
```

R/repair_constraints.R
```diff
@@ -0,0 +1,83 @@
...
+
+  # TODO : Emit message only once (also fix in the current implementation?)
+  if (is_null(in_place)) {
+    message("Not persisting, use `in_place = FALSE` to turn off this message.")
+    in_place <- FALSE
+  }
+
+  out <- dm
+  if (!is.null(plan$insert)) {
+    # TODO: find a way to remove "Matching, by =" message due to `dm_rows_run` and `rows_insert.tbl_dbi`
```
krlmlr, line 69: We need to pass `by = ...` to `rows_insert()` .

krlmlr (commented):
Thanks, I like a lot where this is going!

krlmlr:
To clarify: I think now `check_fk()` only should return the repair plan, and we derive the label from that repair plan elsewhere.

R/examine-constraints.R
moodymudskipper: At the moment we might, but our repair plan for pruning doesn't work, if for example we end up with a repair plan containing the subset of the fact table to be removed, it wouldn't be possible anymore. Unless we make the repair plan a richer object that would contain the ingredients to build the label, I don't know if it's more elegant than the current.

moodymudskipper:
I propose to make it work first, and then simplify this part.
One detail that I missed is that several fact tables  might refer to the same missing primary key in the dimension table, in that case the current solution would add duplicate primary keys.

Another issue will be that to consider once we figure out the deletion is that the whole thing might need to be recursive, because if we delete rows from a table that has both foreign and primary keys, we'll need to prune its child tables too. This recursivity in turn might induce a progress bar hell.

We also need a set of minimal test cases  to investigate corner cases.

R/examine-constraints.R
krlmlr: I think `dm_rows_delete()` should use all columns if no key column is set. Also, I think we should change it to take the PK information from the "other" dm (that contains the rows to be deleted). We can handle that in a separate PR that we merge prior to this PR.

krlmlr:
Right, this turns out to be more complicated than anticipated. Should we tackle the blocking problems first? Could you please file new issues in the current milestone so that we can discuss separately?

R/foreign-keys.R
```diff
@@ -502,7 +502,7 @@ check_fk <- function(t1, t1_name, colname, t2, t2_name, pk) {
 
   t1_join <-
     t1 %>%
-    count(!!!t1_vals) %>%
+    count(!!!t1_vals, name = "*n*") %>%
```
krlmlr, line 505: See `unique_prefix()` for an existing alternative approach to names that are guaranteed to be unique.

R/repair_constraints.R
```diff
@@ -0,0 +1,129 @@
...
+  out <- dm
+
+  if (repair == "insert") {
+    # TODO: find a way to remove "Matching, by =" message due to `dm_rows_run` and `rows_insert.tbl_dbi`
+    dm2 <- new_dm(lst(!!sym(repair_tbl_name) := repair_tbl_obj))
+    suppressMessages(out <- dm_rows_insert(dm, dm2, in_place = in_place))
+    return(out)
+  }
+
+  if (repair == "delete") {
```
krlmlr, line 112: I wonder if we can still use `dm_rows_delete()` in this case: specify the **parent rows** in the dm to delete, even if they don't exist; the delete still should propagate to the child tables.

I'd like to avoid introducing a new operation just for this special case.

krlmlr:
Moving forward, I think we might need to split this PR in half: the first part would be about the extension to `dm_examine_constraints()`, the second part would handle the rest. Would you like to squash this PR into two separate commits, to retain authorship?

moodymudskipper:
I squashed everything then edited and recommited into https://github.com/cynkra/dm/pull/571/commits/2efc77fc59937680e8610be722b1a1c6e88860a3 and https://github.com/cynkra/dm/pull/571/commits/a65893e183a854e09491fcc1f4beaadb59efbde5 , it's a first for me so I'm not sure if I did it right ?

krlmlr:
Let's keep this open to ensure we don't forget the two commits you've prepared. We could start a new PR from the first commit, and work on the second commit later.

moodymudskipper:
I think I messed up the squash, the new_repair_plan function ended up in the other commit and is required here...

krlmlr:
Relevant for #4. Keeping as reference.

PR #572, opened 2021-05-27
use `check_suggested()`

moodymudskipper:
`check_suggested()` is used instead of `if/stop` and `stopifnot` patterns


krlmlr (commented):
Thanks.

- Can you please format as

    ```r
    check_suggested("brio",
      use = ...,
      ...
    )
    ```

  so that it's easier to grep for the suggested-but-required packages?
- Should we take the opportunity to emit customized messages in these checks, like "`dm_draw()` needs the DiagrammeR package. Do you need `install.packages("DiagrammeR")`?" . (The "Do you need" part could be canonical.)

moodymudskipper:
> * Should we take the opportunity to emit customized messages in these checks, like "`dm_draw()` needs the DiagrammeR package. Do you need `install.packages("DiagrammeR")`?" . (The "Do you need" part could be canonical.)

Is there any new logic to code here or do you only want more detailed and consistent messages ?

krlmlr:
It looks like the caller should pass in the top-level function name (`"dm_draw()"`, `"dm_financial()"`) and the package name that's missing. The goal is to achieve consistent error messages.

moodymudskipper:
In `.onLoad()` we check if {DiagrammeR} and {DiagrammeRsvg} are installed and they're both suggested packages.

`.onLoad()` is a mandatory step to use {dm} isn't it ? Then shouldn't these packages be imported rather than suggested ?

moodymudskipper:
Apart from the above, other errors are now standardized, I also documented `check_suggested` with noRmd tag.

krlmlr (commented):
Thanks. Could you please add a test for `check_suggested()`? A snapshot test will do, we can test for a bogus package name e.g. with a hyphen in it.

krlmlr:
We also need to resolve conflicts.

krlmlr (approved):
Thanks, looks good!

Issue #574, opened 2021-05-29
scope of dm_repair_constraints and repair plans


moodymudskipper:
Related to #4 and #571

`dm_repair_constraints()` should be able to : 

* Deal with orphaned foreign keys, either by removing rows containing those, or adding primary keys
* Disambiguate duplicate primary keys, by removing duplicates, rules and approach TBD 
* Something else ? maybe circular references ? remove childless primary keys ?

A repair plan is currently a list containing dms called `insert`, `delete` or `update` . `update` is not used and it's not clear if it will be. It's not clear to me if a repair plan containing more than one of those makes sense.

### insert

`check_pk(..., fk_repair = "add")` 's output features an repair plan containing a "insert" item ("add" might be renamed to "insert" for consistency),  this item is a one table data model where the table is named as the dimension table to update and the columns are the primary keys to insert. It uses `dm_rows_insert`

A naive design (as is now) of `dm_repair_constraints` might create duplicate primary keys if several fact tables feature the same extra keys, thus we should make this reparations recursive OR disambiguate the duplicates as a last step. The former is probably cleaner.

### delete

`check_pk(..., fk_repair = "delete")` 's output should feature a repair plan containing a "delete" item. This item is a one table data model where the table is named as the fact table to remove from, it contains the subset to remove. It uses `dm_rows_delete`

If the table has primary keys itself, we might create orphan foreign keys in other tables by removing them, which means more rows to delete, so here also the algo should be recursive.

Needs *(edit to insert issue reference)*.

## Deal with pk duplicates primary keys

If rows containing duplicate primary keys aren't identical a choice has to be made.

* The simplest approach is to keep the first only. 
* A second approach would be to take the most complete (for instance if the original rows contain a missing value, filled by the duplicate), and keep the first 
* A third approach is to remove all ambiguous rows, and relevant values in linked tables
* Is there more ?

If we choose a flexible approach, a `pk_repair` arg  to `check_pk_constraints` or another function might set those rules.

krlmlr:
add -> insert

Do we really need recursion for "insert"? In the case of multiple facts using the same dimensions, could we collect the duplicates from all facts and use `union()` ?

For "delete", propagating deletion sounds hard. Should we instead support recursive delete in `dm_rows_delete()`? Other than that, I'm not sure we need recursion -- if multiple dimensions use the same fact, we can again use `union(anti_join(...), union(anti_join(...), ...))` .

[...]

PR #579, opened 2021-06-01
Use RHS to decide the keys of a dm_rows_*() operation

krlmlr:
For `dm_repair_constraints()` in #571, #574, #4.


krlmlr:
Let's make this an option, `keys_from = c("lhs", "rhs")` .

krlmlr:
Relevant for #4, keeping as reference.

PR #596, opened 2021-07-05
new_dm2()

krlmlr:
for `dm_meta()`.

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -93,7 +93,7 @@ new_dm <- function(tables = list()) {
   new_dm2(tables)
 }
 
-new_dm2 <- function(tables,
+new_dm2 <- function(tables = list(),
                     pks = structure(list(), names = character()),
                     fks = structure(list(), names = character()),
                     validate = TRUE) {
--- a/R/dm.R
+++ b/R/dm.R
@@ -138,9 +138,9 @@ new_dm3 <- function(def, zoomed = FALSE, validate = TRUE) {
   out <- structure(list(def = def), class = class, version = 1L)
 
   # Enable for strict tests:
-  if (validate) {
-    validate_dm(out)
-  }
+  # if (validate) {
+  #   validate_dm(out)
+  # }
 
   out
 }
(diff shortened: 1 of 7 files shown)


R/zzx-deprecated.R
```diff
@@ -590,19 +590,45 @@ cdm_insert_zoomed_tbl <- function(dm, new_tbl_name = NULL, repair = "unique", qu
...
     ) %>%
-    new_dm3(zoomed = TRUE)
+    new_dm3(zoomed = TRUE, validate = FALSE)
 
   dm_wo_outgoing_fks %>%
     dm_insert_zoomed_outgoing_fks(new_tbl_name_chr, names_list$old_new_names[old_tbl_name], col_tracker_zoomed(dm)) %>%
     dm_clean_zoomed()
 }
 
+dm_select_tbl_zoomed_impl <- function(dm, selected) {
```
TSchiefer, line 606: Good catch that these two functions were needed to avoid producing an invalid dm at some point in `cdm_insert_zoomed_tbl()` (even though it somehow worked in the end).

R/dm.R
```diff
@@ -116,13 +119,15 @@ new_dm <- function(tables = list()) {
...
+    mutate(fks = as_list_of(map(fks, `%||%`, new_fk()), .ptype = new_fk())) %>%
+    mutate(filters = list_of(new_filter())) %>%
     left_join(zoom, by = "table") %>%
     left_join(col_tracker_zoom, by = "table")
 
-  new_dm3(def, validate = FALSE)
+  new_dm3(def, validate = validate)
 }
 
 new_dm3 <- function(def, zoomed = FALSE, validate = TRUE) {
```
TSchiefer, line 133: Why don't we enable validation now?

TSchiefer (approved):
Nice intermediate function to create dms, makes sense. Shall we also implement specific tests for `new_dm2()`?

R/dm.R
krlmlr: It's slow. #597 adds a workflow that enables this. It currently fails, will succeed after we merge this PR.

PR #597, opened 2021-07-06
Add validation workflow

### LINKED ISSUE #202: Enable optional validation
for all functions that return a `dm`.

Rationale: Validation is expensive, but we want to fail as early as possible when testing.

krlmlr:
The workflow is a tweaked copy of the db-sqlite workflow.

Closes #202.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -1,0 +1,124 @@
+on:
+  push:
+    branches:
+      - main
+      - master
+      - "r-*"
+  pull_request:
+    branches:
+      - main
+      - master
+      - "r-*"
+  schedule:
+    - cron:  '20 2 * * *'
+
+name: validate
+
+jobs:
+  db-tests:
+    runs-on: ${{ matrix.config.os }}${{ matrix.config.os-version }}
+
+    name: ${{ matrix.config.os }}${{ matrix.config.os-version }} (${{ matrix.config.r }}) ${{ matrix.config.desc }}
+
+    # Begin custom: services
+    # End custom: services
+
+    strategy:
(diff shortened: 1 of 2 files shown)


krlmlr:
The codecov change is likely a false positive -- different workflows cover different parts of the code, perhaps the upload fails for some?

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,124 @@
...
+      # End custom: before install
+
+      - name: Install dependencies
+        run: |
+          remotes::install_deps(dependencies = TRUE, type = .Platform$pkgType)
+        shell: Rscript {0}
+
+      - name: Instrument R code for validation
+        run: |
+          sed -i -r '/INSTRUMENT: validate/ s/^( +)# /\1/g' R/*
```
TSchiefer, line 96: 1. Are you adding this extra workflow so we can see how long the validation takes by comparison? 
1. Do we know in how many cases the validation is then executed in the following tests?
1. Are you using the general regex in the sed command because there might be more places where we want to use `validate_dm()` in the future?

TSchiefer (commented):
TBH I am not sure I understand what's the purpose of doing it like this. A short explanation would be appreciated.

.github/workflows/validate.yaml
krlmlr: 1. No
2. It's executed every time `new_dm3()` is called, except if it's called from `validate_dm()` -- to avoid infinite recursion
3. I don't worry about replacing multiple instances. However, we should check that the `sed` actually changes at least one file.

TSchiefer (approved):
LGTM

PR #607, opened 2021-08-17
Support `returning` in `rows_*.tbl_dbi()`

### LINKED ISSUE #593: rows_insert() et al. with option to return modified rows
SQL Server: `OUTPUT` (https://docs.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql?view=sql-server-ver15)

Postgres: `RETURNING` (https://www.postgresql.org/docs/current/sql-insert.html)

SQLite: `RETURNING` (https://www.sqlite.org/lang_returning.html)

Other DBMS are likely to support this too.

This is important to avoid an extra roundtrip after inserting rows with an `IDENTITY` column.

mgirlich:
Closes #593.

* No tests yet. For SQLite tests are blocked by https://github.com/r-dbi/RSQLite/issues/380 and hindered by https://github.com/r-dbi/RSQLite/issues/381.
* You might prefer a different code style. For convenience I used the `dbplyr` functions `ident()` and `escape()`
* The generic `dplyr::rows_update()` doesn't have an argument `returning` but now `rows_update.tbl_dbi()` has it. To use the tidyselect interface I `enquo()`. This doesn't work with the `check_dots_used()` used in the generic.


krlmlr:
Thanks. Should we discuss conditional updates and mutating joins separately?

krlmlr (commented):
Thanks, this is a good start.

I think type stability for the return values is really important here. Can we return the inserted rows in an attribute to the returned table?

krlmlr:
Please also run `devtools::document()` to avoid check errors.

mgirlich:
> I think type stability for the return values is really important here. Can we return the inserted rows in an attribute to the returned table?

You mean something like

```r
returned_rows <- dbExecute(con, sql, immediate = TRUE)
attr(x, "returned_rows") <- returned_rows
invisible(x)
```

and then have a helper function e.g. `get_returned_rows(x)` to get the returned rows?

I can do that and I think type stability is important. On the other hand it also seems a bit counterintuitive to me. The `x` argument can only be a table that wasn't modified (otherwise `remote_table()` doesn't work and also doesn't really make sense). And using  `returning` seems to be a very good indicator that one wants to use the result.

In case you have a pipeline of insert/update/upsert operations in mind, i.e.

```r
x %>% 
  dm::rows_insert(y, returning = "id") %>% 
  dm::rows_insert(y2, returning = "id") %>% 
  dm::rows_update(z, returning = "id")
```

 it might make sense that the user could choose the name of the attribute so that all results from returning can still be accessed.

R/rows-db.R
```diff
@@ -25,10 +25,18 @@
...
-#' @return A tbl object of the same structure as `x`.
-#'   If `in_place = TRUE`, the underlying data is updated as a side effect,
-#'   and `x` is returned, invisibly.
+#' @return Either a tbl object of the same structure as `x` or a tibble with
+#'   the columns as in `returning`.
+#'   If `in_place = TRUE`, the underlying data is updated as a side effect and
+#'   if
+#'     * `returning = NULL`, then `x` is returned, invisibly.
+#'     * else a tibble with the inserted data and the specified columns is
+#'       returned.
```
krlmlr, line 39: This is no longer true?

R/rows-db.R
```diff
@@ -144,9 +160,8 @@ rows_delete.tbl_dbi <- function(x, y, by = NULL, ...,
     }
 
     con <- dbplyr::remote_con(x)
-    sql <- sql_rows_delete(x, y, by)
-    dbExecute(con, sql, immediate = TRUE)
-    invisible(x)
+    sql <- sql_rows_delete(x, y, by, returning = returning)
+    get_or_execute(x, con, sql, returning)
```
krlmlr, line 164:
```suggestion
    rows_get_or_execute(x, con, sql, returning)
```

?

R/rows-db.R
```diff
@@ -63,9 +74,9 @@ rows_insert.tbl_dbi <- function(x, y, by = NULL, ...,
     }
 
     con <- dbplyr::remote_con(x)
-    sql <- sql_rows_insert(x, y)
-    dbExecute(con, sql, immediate = TRUE)
-    invisible(x)
+    sql <- sql_rows_insert(x, y, returning = returning)
```
krlmlr, line 77: Do we really need to pass `returning` to the `sql_()` methods?

```suggestion
    sql <- paste0(
      sql_rows_insert(x, y, returning = returning)
      if (!is.null(returning)) paste0("\n", sql_returning(...))
    )
```

krlmlr (commented):
Thanks. Can we simplify further?

krlmlr:
Thanks, the attribute looks good.

Regarding stacking of "returning" data, I think it's dangerous to use the pipe for functions that have side effects:

```r
x %>% 
  dm::rows_insert(y, returning = "id", in_place = TRUE) %>% 
  dm::rows_insert(y2, returning = "id", in_place = TRUE) %>% 
  dm::rows_update(z, returning = "id", in_place = TRUE)
```

If the first call succeeds but the second call fails, you might be tempted to re-run the pipe, which might now also fail the first call. It's safer to un-pipe this and consume return values implicitly. I don't think we need to act here.

R/rows-db.R
mgirlich: If `returning` is not an argument of the `sql_()` methods then I would rather have something like

```r
sql_rows_insert(x, y, returning = returning) %>%
  sql_add_returning(x, returning)
```

and export `sql_add_returning()` (this should be exported to complete the other sql functions). To me this feels like a more user-friendly interface. What do you think?

krlmlr: I think we need an `sql_returning()` generic anyway -- to special-case both the clause and the way column expressions are constructed (to work around that RSQLite bug). I don't mind `sql_add_returning()` as an equivalent to the `paste0()` call in my proposal.

R/rows-db.R
```diff
@@ -391,5 +412,98 @@ sql_rows_delete.tbl_sql <- function(x, y, by, ...) {
...
+#'
+#' @export
+#' @rdname rows-db
+sql_add_returning <- function(sql, x, returning_cols) {
+  if (is_empty(returning_cols)) {
+    return(sql)
+  }
+
+  returning_clause <- sql_returning(x, returning_cols)
+  paste0(sql, if (!is_empty(returning_clause)) paste0("\n", returning_clause))
```
krlmlr, line 455: I think `sql_returning()` should either return a string or throw an error.

```suggestion
  paste0(sql, "\n", returning_clause)
```

R/rows-db.R
```diff
@@ -391,5 +412,98 @@ sql_rows_delete.tbl_sql <- function(x, y, by, ...) {
...
+#'
+#' Extract the RETURNING rows produced by `rows_insert()`, `rows_update()`,
+#' `rows_upsert()`, or `rows_delete().`
+#'
+#' @param x A lazy tbl.
+#'
+#' @return A tibble.
+#'
+#' @export
+get_returned_rows <- function(x) {
```
krlmlr, line 440: Do we need `has_returned_rows()` ?

R/rows-db.R
```diff
@@ -391,5 +412,98 @@ sql_rows_delete.tbl_sql <- function(x, y, by, ...) {
...
+    dbExecute(con, sql, immediate = TRUE)
+  } else {
+    returned_rows <- dbGetQuery(con, sql, immediate = TRUE)
+    attr(x, "returned_rows") <- as_tibble(returned_rows)
+  }
+
+  invisible(x)
+}
+
+#' Extract the RETURNING rows
```
krlmlr, line 430: Should we mark this (and the `returning` argument) as "experimental"?

krlmlr (commented):
Thanks, a few final nitpicks.

For tests of this feature, we can use persistent tables, referring to the blocking issue. If you could add them, I'll start the CI run and merge.

mgirlich:
> For tests of this feature, we can use persistent tables, referring to the blocking issue. If you could add them, I'll start the CI run and merge.

I think it should be sufficient to rely on the dev version of RSQLite as the bug is fixed in there.

krlmlr:
RSQLite 2.2.8 is on CRAN now.

krlmlr (dismissed):
Thanks, this is huge! Looking forward to using it.

krlmlr:
I love SQL Server...

krlmlr (approved):

krlmlr:
Thanks!

PR #610, opened 2021-08-18
Add `rows_patch.tbl_dbi()`

mgirlich:
It seems that `rows_patch.tbl_dbi()` was forgotten to implement 😉 

This also shows a potential strategy for mutating joins: pass a function (or a named list of functions) that combines the values of the value in `x` and in `y`.


krlmlr:
Thanks. I've learned the hard way to stay away from metaprogramming if it's easy to avoid. In this particular case, an `sql_patch_rows()` generic seems like the way to go, even if there's mild duplication at the implementation level of the individual methods.

I also wonder if the `MERGE` SQL statement, where supported, allows for more elegant code than an array of `COALESCE()` in the generated SQL.

krlmlr:
Would you like to take another stab? Could you please resolve conflicts?

mgirlich:
> I also wonder if the MERGE SQL statement, where supported, allows for more elegant code than an array of COALESCE() in the generated SQL.

I don't really understand how you want to use the [MERGE SQL](https://en.wikipedia.org/wiki/Merge_(SQL)) for this. `rows_patch()` is used to overwrite missing values. MERGE SQL is used to upsert. I don't really see how this should help in this case. Can you explain what you had in mind?

> In this particular case, an sql_patch_rows() generic seems like the way to go

Done.

R/rows-db.R
```diff
@@ -408,6 +468,160 @@ sql_rows_update_prep <- function(x, y, by) {
...
+    sql_returning_cols(x, returning_cols)
+  )
+
+  glue::as_glue(sql)
+}
+
+#' @export
+sql_rows_patch.tbl_duckdb_connection <- sql_rows_patch.tbl_SQLiteConnection
+
+sql_rows_patch_prep <- function(x, y, by) {
```
krlmlr, line 581: Should we integrate this functionality into `sql_rows_update_prep()` or perhaps extract common parts?

man/dplyr_table_manipulation.Rd
```diff
@@ -98,7 +98,7 @@ The default returns the last column (on the assumption that's the
 column you've created most recently).
 
 This argument is taken by expression and supports
-\link[rlang:nse-force]{quasiquotation} (you can unquote column
+\link[rlang:bang-bang]{quasiquotation} (you can unquote column
```
krlmlr, line 101: This seems to trigger test failures.

krlmlr (commented):
Thanks, looks good! Do you think full duckdb support is feasible?

krlmlr:
I misunderstood `MERGE` -- you're right it looks more useful for `rows_upsert()` (and perhaps also `rows_insert()` and `rows_delete()` ?).

R/rows-db.R
mgirlich: Probably, but I would do that in a separate PR and focus on implementing `rows_patch()` and `rows_upsert()` for now. What do you think?

mgirlich:
> Thanks, looks good! Do you think full duckdb support is feasible?

It might already work, I just haven't tested it yet. I will have a look at it.

krlmlr (approved):
Thanks!

PR #613, opened 2021-08-31
Establish compatibility with rlang > 0.4.11

krlmlr:
.


TSchiefer (approved):
Thought at first the new rlang version would break our code, but yeah, the info symbol looks nice (and is/will be probably also standard in other pkgs(?))

PR #615, opened 2021-09-13
Use `rows_check_key*()` like in `dplyr`

mgirlich:
The `rows_*.data.frame()` functions use `rows_check_key()` and `rows_check_key_df()` for nicer error messages. Currently, 

```r
dplyr:::rows_patch.data.frame <- function (x, y, by = NULL, ..., copy = FALSE, in_place = FALSE) {
    key <- rows_check_key(by, x, y)
    y <- auto_copy(x, y, copy = copy)
    rows_df_in_place(in_place)
    rows_check_key_df(x, key, df_name = "x")
    rows_check_key_df(y, key, df_name = "y")
    idx <- vctrs::vec_match(y[key], x[key])
    bad <- which(is.na(idx))
    if (has_length(bad)) {
        abort("Attempting to patch missing rows.")
    }
    new_data <- map2(x[idx, names(y)], y, coalesce)
    x[idx, names(y)] <- new_data
    x
}

dplyr:::rows_check_key_df <- function (df, by, df_name) {
    y_miss <- setdiff(by, colnames(df))
    if (length(y_miss) > 0) {
        abort(glue("All `by` columns must exist in `{df_name}`."))
    }
    if (vctrs::vec_duplicate_any(df[by])) {
        abort(glue("`{df_name}` key values are not unique."))
    }
}

dplyr:::rows_check_key <- function (by, x, y) {
    if (is.null(by)) {
        by <- names(y)[[1]]
        inform(glue("Matching, by = \"{by}\""), class = c("dplyr_message_matching_by", 
            "dplyr_message"))
    }
    if (!is.character(by) || length(by) == 0) {
        abort("`by` must be a character vector.")
    }
    if (any(names2(by) != "")) {
        abort("`by` must be unnamed.")
    }
    bad <- setdiff(colnames(y), colnames(x))
    if (has_length(bad)) {
        abort("All columns in `y` must exist in `x`.")
    }
    by
}
```

### DIFF
--- a/tests/testthat/_snaps/rows-dm.md
+++ b/tests/testthat/_snaps/rows-dm.md
@@ -29,6 +29,9 @@
       out <- dm_rows_insert(flights_sqlite, flights_hour10_sqlite)
     Message <simpleMessage>
       Not persisting, use `in_place = FALSE` to turn off this message.
+    Message <dplyr_message_matching_by>
+      Matching, by = "origin"
+      Matching, by = "year"
     Code
       print(dm_nrow(flights_sqlite))
     Output
@@ -36,6 +39,10 @@
             15       86        0      945        0 
     Code
       dm_rows_insert(flights_sqlite, flights_hour10_sqlite, in_place = TRUE)
+    Message <dplyr_message_matching_by>
+      Matching, by = "origin"
+      Matching, by = "year"
+    Code
       print(dm_nrow(flights_sqlite))
     Output
       airlines airports  flights   planes  weather 
@@ -47,6 +54,10 @@
       hour == 11) %>% dm_update_zoomed()
       flights_hour11_sqlite <- copy_dm_to(sqlite, flights_hour11)
       flights_new <- dm_rows_insert(flights_sqlite, flights_hour11_sqlite, in_place = FALSE)
+    Message <dplyr_message_matching_by>
+      Matching, by = "origin"
+      Matching, by = "year"
+    Code
       print(dm_nrow(flights_new))
     Output
       airlines airports  flights   planes  weather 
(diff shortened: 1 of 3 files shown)


krlmlr (commented):
Thanks, looks good. Do you need to update other snapshot tests?

mgirlich:
In `dm_rows_insert(flights_sqlite, flights_hour10_sqlite)` you now get a message

```
Matching, by = "origin"
Matching, by = "year"
```

I'm not sure whether you want this or not. You might want to have a look whether/how you could avoid them.

krlmlr:
Thanks. I'm not sure we want to issue these messages for `dm_rows_*()` operations -- keys are available in the data model after all. What should `dm_rows_insert()` do for tables that have no primary key -- should we pass `by = c()` and issue no message in this case? Should dplyr be adapted too?

tests/testthat/_snaps/rows-dm.md
```diff
@@ -29,13 +29,20 @@
       out <- dm_rows_insert(flights_sqlite, flights_hour10_sqlite)
     Message <simpleMessage>
       Not persisting, use `in_place = FALSE` to turn off this message.
+    Message <dplyr_message_matching_by>
```
krlmlr, line 32: We shouldn't see those.

mgirlich: I used the same class for consistency even though it doesn't really fit the conventions regarding packages. What exactly do you want:

a) messages without a class?
b) messages with a different class, e.g. `c("dm_message_matching_by", "dm_message")`?

mgirlich:
> What should dm_rows_insert() do for tables that have no primary key -- should we pass by = c() and issue no message in this case?

Actually, to me the primary key is not the decisive factor. See the use case for `returning` and `rows_insert()`: often you have a primary key column `id` and a unique constraint on some identifiers. If your local data don't have an `id` column then unique constraint(s) is/are the relevant part for the insert.

> Should dplyr be adapted too?

I don't really understand what you mean here. In `dplyr` you get a message if `by` is `NULL` and no message when you specify `by`. What would you want to adapt here?

tests/testthat/_snaps/rows-dm.md
krlmlr: No message.

krlmlr:
I think we need to adapt `dm_rows_run()` -- the `require_keys` argument is a hack, we need to operate more fine-grained depending on if a table has primary keys or not. If there are no primary keys, we should pass `by = c()`, and this should be supported for both data frame and database backends.

krlmlr:
This seems obsolete with dbplyr 2.2.0. Thanks for your work on that!

PR #616, opened 2021-09-14
Add `rows_upsert.tbl_dbi()`

mgirlich:
Together with `rows_patch()` this would complete the support for the `rows_*()` functions 😄 

* I only added SQLite, MariaDB and Postgres for now.
* I used the ON CONFLICT clause. This only works when there also is a unique constraint but - at least for Postgres - has some [advantages](https://wiki.postgresql.org/wiki/UPSERT#SQL_MERGE_syntax) but also the disadvantage that the primary key sequence is increased.


R/rows-db.R
```diff
@@ -408,6 +460,117 @@ sql_rows_update_prep <- function(x, y, by) {
...
+    "WHERE true\n",
+    "ON DUPLICATE KEY UPDATE\n",
+    update_clause, "\n",
+    sql_returning_cols(x, returning_cols)
+  )
+
+  glue::as_glue(sql)
+}
+
+sql_rows_upsert_prep <- function(x, y, by) {
```
krlmlr, line 535: Can we integrate into `sql_rows_update_prep()` or extract common parts?

tests/testthat/test-rows-db.R
```diff
@@ -107,6 +107,57 @@ test_that("update with returning argument (#607)", {
...
+    data %>% arrange(select)
+    rows_upsert(data, test_db_src_frame(select = 2:3), in_place = TRUE)
+    data %>% arrange(select)
+    rows_upsert(data, test_db_src_frame(select = 0L, where = "a"), by = "where", in_place = TRUE)
+    data %>% arrange(select)
+  })
+})
+
+test_that("upsert with returning argument (#607)", {
+  skip_if_src("duckdb", "mariadb")
```
krlmlr, line 140: Is this a typo?

```suggestion
  skip_if_src("duckdb", "mssql")
```

krlmlr (commented):
Thanks, looks good.

The test failure looks weird, does it succeed locally?

What do we need unique indices for? Can you summarize the advantages of `ON CONFLICT` -- or why we shouldn't use `MERGE` here?

mgirlich:
> What do we need unique indices for? Can you summarize the advantages of `ON CONFLICT` -- or why we shouldn't use `MERGE` here?

## Upsert Support
According to Wikipedia the following Databases support MERGE:
* Oracle Database
* DB2
* Teradata
* EXASOL
* Firebird (with some restrictions according to Wikipedia?)
* CUBRID
* H2
* HSQLDB
* MS SQL
* Vectorwise
* Apache Derby

Alternative in MySQL, SQLite, PostgreSQL
* MySQL
  * INSERT ... ON DUPLICATE KEY UPDATE -> join on PRIMARY KEY or UNIQUE constraints
  * REPLACE INTO -> attempt insert, if it fails delete the row (if existing) and insert the new one
  * Support for ignoring duplicate keys (insert only new rows)
* SQLite
  * INSERT OR REPLACE INTO -> similiar to MySQL
* PostgreSQL
  * INSERT INTO ... ON CONFLICT [ conflict_target ] conflict_action -> similar to MySQL

Other Databases
* Apache Phoenix supports UPSERT VALUES[12] and UPSERT SELECT[13] syntax.
* Spark SQL supports UPDATE SET * and INSERT * clauses in actions.[14]
* Apache Impala supports UPSERT INTO ... SELECT.[15]

## Merge
There are a couple of reason why Postgres doesn't support MERGE. One of them is issues with concurrency:

> the SQL standard does not require that MERGE be atomic in the sense of atomically providing either an INSERT or UPDATE, and other products do not provide any such guarantees with their MERGE statement.

The biggest issue with INSERT INTO ... ON CONFLICT is that the typical ID field (i.e. an auto incrementing sequence) is increased by one for every row that one tries to insert, i.e. in particular also for existing rows that end up being updated. If you have frequent big upsert where most rows already exists this can blow up your ID quite a big and suddenly it has to be a bigint. Then the fun starts thanks to the terrible support of big integers in R...

## Alternative for MySQL, SQLite, PostgreSQL
If there is no unique constraint one can build an upsert query like follows

```sql
WITH updated AS
( 
    UPDATE table
        SET ...
    FROM new_values
    WHERE ...
    RETURNING ...
)
INSERT INTO table
SELECT ...
FROM new_values
WHERE NOT EXISTS (
    SELECT 1 
    FROM updated
    WHERE ... -- if there is a primary key one can use it. Otherwise basically the same condition as in the CTE
)
```

Apart from a longer syntax this has concurrency issues (there are many SO threads on this).

mgirlich:
> The test failure looks weird, does it succeed locally?

It works when I add `Sys.setenv(DM_TEST_SRC = "postgres")` to `helper-src.R` but produces the same error without it. I'll have a look.

mgirlich:
There seems to be another issue for SQLite with temporary tables and RETURNING

``` r
test_sqlite <- function(temporary) {
  con <- RSQLite::dbConnect(RSQLite::SQLite(), ":memory:")
  
  if (temporary) {
    DBI::dbExecute(con, 'CREATE TEMPORARY TABLE test_frame_2021_09_14_11_22_08_18668_10 ("select" integer, "where" varchar(20), "exists" REAL);')
  } else {
    DBI::dbExecute(con, 'CREATE TABLE test_frame_2021_09_14_11_22_08_18668_10 ("select" integer, "where" varchar(20), "exists" REAL);')
  }
  DBI::dbExecute(con, "INSERT INTO test_frame_2021_09_14_11_22_08_18668_10 VALUES (1, 'a', 0.5);")
  DBI::dbExecute(con, "INSERT INTO test_frame_2021_09_14_11_22_08_18668_10 VALUES (2, 'b', 1.5);")
  DBI::dbExecute(con, "INSERT INTO test_frame_2021_09_14_11_22_08_18668_10 VALUES (3, NULL, 2.5);")
  DBI::dbExecute(con, 'CREATE UNIQUE INDEX unique_select_id ON test_frame_2021_09_14_11_22_08_18668_10 ("select");')
  
  DBI::dbExecute(con, 'CREATE TEMPORARY TABLE dbplyr_031 ("select" integer, "where" varchar(20));')
  DBI::dbExecute(con, "INSERT INTO dbplyr_031 VALUES (2, 'x');")
  DBI::dbExecute(con, "INSERT INTO dbplyr_031 VALUES (3, 'y');")
  DBI::dbExecute(con, "INSERT INTO dbplyr_031 VALUES (4, 'z');")
  
  result <- DBI::dbGetQuery(
    con,
    'WITH "...y"("select", "where") AS (
    SELECT *
    FROM "dbplyr_031"
    )
    INSERT INTO test_frame_2021_09_14_11_22_08_18668_10 ("select", "where")
    SELECT * FROM "...y"
    WHERE true
    ON CONFLICT ("select")
    DO UPDATE
    SET "where" = excluded."where"
    RETURNING *;'
  )
  
  print(result)
  
  DBI::dbRemoveTable(con, "test_frame_2021_09_14_11_22_08_18668_10")
}

test_sqlite(FALSE)
#>   select where exists
#> 1      2     x    1.5
#> 2      3     y    2.5
#> 3      4     z     NA
test_sqlite(TRUE)
#>   select where exists
#> 1      4     z     NA
```

<sup>Created on 2021-09-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
I don't really know why it doesn't work for MS SQL. Could someone please get rid of MS SQL?...

It works fine on [sqliteonline](https://sqliteonline.com/)

Microsoft SQL Server 2019 (RTM-CU9) (KB5000642) - 15.0.4102.2 (X64) 
	Jan 25 2021 20:16:12 
	Copyright (C) 2019 Microsoft Corporation
	Express Edition (64-bit) on Linux (CentOS Linux 8) <X64>

but not on [sqlfiddle](http://sqlfiddle.com/)

Microsoft SQL Server 2017 (RTM-CU2) (KB4052574) - 14.0.3008.27 (X64)
	Nov 16 2017 10:00:49
	Copyright (C) 2017 Microsoft Corporation
	Express Edition (64-bit) on Linux (Ubuntu 16.04.3 LTS)

So probably it only works with MS SQL 2019.

krlmlr:
I feel your pain with MSSQL. We keep it as a special case for entertainment.

The tests complain about a semicolon, can we pass it via DBI?

Is there an upstream bug report for SQLite, is it perhaps fixed already?

mgirlich:
To me the syntax looks the same for [2017](https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-linux-2017) and [2019](https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-linux-ver15).

mgirlich:
> Is there an upstream bug report for SQLite, is it perhaps fixed already?

I haven't created one and in a quick search I haven't seen one.

krlmlr:
Can we repro using `sqlite3` ?

I don't mind installing MSSQL 2019 on GitHub Actions, perhaps in parallel to MSSQL 2017. There is https://github.com/ankane/setup-sqlserver that can help with that, or perhaps install manually.

mgirlich:
> Can we repro using `sqlite3` ?

See my [comment above](https://github.com/cynkra/dm/pull/616#issuecomment-920624883)

> I don't mind installing MSSQL 2019 on GitHub Actions, perhaps in parallel to MSSQL 2017. There is https://github.com/ankane/setup-sqlserver that can help with that, or perhaps install manually.

You can if you want to. I don't care so much about MS SQL itself. I mainly implemented it because it was one of the databases with support for MERGE.

krlmlr:
Yeah, let's skip the test for now with a comment.

mgirlich:
@krlmlr Do you know why it fails for MS SQL?

Error (test-dplyr-src.R:34:3): 'copy_to.dm()' works
Error: Error: promise already under evaluation: recursive default argument reference or earlier problems?
Backtrace:
 1. dm:::test_src_frame(!!!mtcars)
 3. dbplyr:::copy_to.src_sql(...)
 5. dbplyr:::db_copy_to.DBIConnection(...)
 7. dbplyr:::create_indexes(con, table, unique_indexes, unique = TRUE)

mgirlich:
The code coverage check fails because

* currently no tests are run for Maria DB. Do you want to add them to the workflow?
* upsert tests are skipped for MS SQL -> I remove them from the coverage with `#nocov`
* no database in the workflows use `sql_rows_upsert.tbl_sql()` -> I remove them from the coverage with `#nocov`

krlmlr:
I resolved the conflict by keeping both changes. At some point the new `_prep()` function should be integrated with `sql_rows_prep()` from #624?

krlmlr (approved):
Thanks! Looks like we no longer need `sql_rows_update_prep()` -- I'll clean up later.

PR #617, opened 2021-09-16
Support DuckDB in `rows_insert()`, `rows_update()` and `rows_delete()`

mgirlich:
Supporting DuckDB in these three functions seems to be pretty easy. Unfortunately, DuckDB doesn't seem to support RETURNING. At least I didn't find in the [documentation](https://duckdb.org/docs/sql/statements/insert) and got nothing back when I tried it out at first.

I guess that we could also remove `sql_rows_update.tbl_SQLiteConnection()` and let SQLite simply use `sql_rows_update.tbl_sql()`. What do you think?


R/rows-db.R
```diff
@@ -511,6 +534,13 @@ sql_returning_cols.tbl_dbi <- function(x, returning_cols, ...) {
   paste0("RETURNING ", returning_cols)
 }
 
+#' @export
+sql_returning_cols.tbl_duckdb_connection <- function(x, returning_cols, ...) {
+  if (!is_empty(returning_cols)) {
+    abort("DuckDB does not support the `returning` argument.")
+  }
```
krlmlr, line 541: Emptiness is handled by the generic already?

```suggestion
  abort("DuckDB does not support the `returning` argument.")
```

R/rows-db.R
```diff
@@ -423,13 +445,14 @@ sql_rows_delete.tbl_sql <- function(x, y, by, ..., returning_cols = NULL) {
...
-    "WITH ", p$y_name, "(", p$y_columns_qq, ") AS (\n",
-    dbplyr::sql_render(y),
-    "\n)\n",
-    #
     "DELETE FROM ", p$name, "\n",
     sql_output_cols(x, returning_cols, delete = TRUE),
-    "WHERE EXISTS (SELECT * FROM ", p$y_name, " WHERE ", p$compare_qual_qq, ")",
+    "WHERE EXISTS (\n",
+    "  SELECT * FROM (\n",
+    "    ", dbplyr::sql_render(y), "\n",
```
krlmlr, line 452: What's the motivation for this change? How/why did it work before?

krlmlr (commented):
Thanks! Yeah, the fewer specialized methods we have the better.

krlmlr:
Is there an existing DuckDB issue regarding support for `RETURNING` ?

CC @hannesmuehleisen.

R/rows-db.R
mgirlich: This just replaces the CTE with a subquery to make it work with DuckDB (CTEs don't work with INSERT, UPDATE, DELETE...)

krlmlr (approved):
Thanks!

mgirlich:
> Is there an existing DuckDB issue regarding support for `RETURNING` ?
> 
> CC @hannesmuehleisen.

I didn't find an issue yet and therefore created one: https://github.com/duckdb/duckdb/issues/2291.

PR #621, opened 2021-09-17
Add `na_matches` argument to `rows_*()`

### LINKED ISSUE #619: `rows_*()`: add `na_matches` argument
It probably makes sense to allow the user to control whether two `NULL`s should be considered equal or not. Maybe this should be possible on a per column basis, i.e. `na_matches` could either be a scalar or a named character vector.

mgirlich:
Closes #619.

* add `na_matches`. The default is `"never"` as in `dbplyr` (though in `dplyr` the default is `"na"`).
* clean up `sql_rows_update_prep()` and `sql_rows_patch_prep()`
* remove some unnecessary specialisations

Notes:
* this won't really work for `rows_upsert()` for databases that need a unique constraint (Postgres, SQLite, and MariaDB) as `NULL != NULL` in SQL...
* I haven't added tests yet.


mgirlich:
For DuckDB `in_place = FALSE` doesn't work with `na_matches = "na"` due to a [bug in `dbplyr`](https://github.com/tidyverse/dbplyr/pull/705).

krlmlr:
Thanks. This looks good by and large, I prefer smaller PRs. Can you please move the consolidation of `sql_rows_prep()` into a separate PR? I suspect this will be the basis for the other changes (removal of CTE specializations, `na_matches`) which can then be reviewed separately without conflict.

mgirlich:
I created a new PR #624 for `sql_rows_prep()`, so I close this one.

PR #624, opened 2021-09-20
Refactor `sql_rows_prep()`

mgirlich:
Combine `sql_rows_update_prep()` and `sql_rows_patch_prep()` into a single function.
Note that Maria DB used the non-existing field `target_columns_qual_qq`. Maybe you want to add MariaDB to the workflows?


krlmlr (approved):
Thanks!

Issue #618, opened 2021-09-16
Refactor `rows_db.R`


mgirlich:
* [x] Refactor `sql_rows_*_prep()`
* [ ] Remove SQLite workaround for incorrect column names after RETURNING -> #632
* [ ] Improve readability of tests?

mgirlich:
Closing as `rows_*()` is now implemented in dbplyr.

PR #632, opened 2021-09-28
Remove workaround for SQLite returning issue

mgirlich:
Part of #618


krlmlr:
This requires the latest and greatest RSQLite, unfortunately it is not enforced or checked in the `DESCRIPTION` .

Should we add logic to `.onLoad()` to alert the user? We would need to handle both cases: RSQLite loaded before, or after our package, perhaps similarly to `vctrs::s3_register()` .

mgirlich:
Ah, I thought it would be enough to change the version of SQLite in the `DESCRIPTION`. But this doesn't help as it is only a suggested package? Makes sense to inform the user. I think the newest SQLite is only required for the `returning` argument. 
Do you want to explicitly error in there for older versions?

mgirlich:
I added a check when loading `RSQLite`. Do you want another version check in returning (though that would bring back `sql_returning_cols.tbl_SQLiteConnection()`)?

Issue #599, opened 2021-07-19
Support passage of `pool` object to `dm_from_src`


brancengregory:

krlmlr:
Thanks. If it doesn't work currently, could you please provide a reprex, perhaps with the RSQLite or duckdb packages?

jaredlander:
This isn't exactly the reprex you asked for, but it's what I have on hand.

```r
conn <- pool::dbPool(
    odbc::odbc(),
    Driver = "postgresql",
    Server = 'server_address',
    Port = 'port',
    Database = 'database_name',
    UID = 'user',
    PWD = 'password',
    timeout = 10
)

thedb_raw <- dm_from_src(conn, learn_keys=FALSE)

> Error in UseMethod("sql_escape_ident") : 
  no applicable method for 'sql_escape_ident' applied to an object of class "c('Pool', 'R6')"
```

krlmlr:
Thanks. AFAICT you need to "check out" a connection from a pool object to be able to work with it. Are you suggesting to enhance dm so that it will auto-checkout as needed?

jaredlander:
Well that's something I didn't know about.

This works.

```r
thedb_raw <- dm_from_src(conn %>% pool::poolCheckout(), learn_keys=FALSE)
```

But for an end user, having `{dm}` check it out might make a lot of sense. Especially if it is used in `{shiny}`, I'm not quite sure where/when it should be checked out and returned.

krlmlr:
Thanks, makes sense. We might be able to check out and return during initialization of the dm object, so that each table access does its own checkout if needed.

jaredlander:
If you return at creation, do you need to checkout again each time you perform work on the object?
 And will mutations be written to the database?

krlmlr:
To support Pool objects in dm we'd need to make them work transparently in dbplyr first.

dm objects are immutable, except for the `dm_rows_*()` family of functions. What mutations are you referring to?

jaredlander:
I'm not at the computer, but if we wanted to do something like `dm_insert_zoomed()` after we return the pool connection, would that work?

krlmlr:
`dm_insert_zoomed()` only modifies the metadata stored on the client-side, it never executed DDL on the server.

jaredlander:
I thought there was a way to write back to the database (still learning the package, this is my first time using it extensively)

krlmlr:
[...]

PR #633, opened 2021-09-28
support 'Pool' objects in sql_schema_list

brancengregory:
For #599.

I am not familiar enough with how to write a test for this, sorry


krlmlr:
Thanks, good catch. Let's not worry about the tests for now, I think we can add a "pool" test source to the test suite at some point, so that all methods will be covered.

One thing: Can you please run `devtools::document()`, commit the resulting change to `NAMESPACE` to this branch, and push?

brancengregory:
@krlmlr, I ran `devtools::document()` and there was actually no change. I believe `NAMESPACE` was already edited in my initial commit. Thanks

krlmlr (dismissed):
Thanks!

PR #634, opened 2021-09-29
CRAN release v0.2.4

krlmlr:
- [x] Resolve all tasks in `cran-comments.md`
- [x] Run `fledge::release()`
- [ ] When the package is accepted on CRAN, run `fledge::post_release()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,56 +2,27 @@
 
 # dm 0.2.4
 
-- Same as previous version.
-
-
-# dm 0.2.3.9007
-
-- `sql_schema_list()` supports `pool::Pool` objects (#633, @brancengregory).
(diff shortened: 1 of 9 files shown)


krlmlr:
@mgirlich: Winbuilder shows failures related to `returning`, https://win-builder.r-project.org/2CE07rUYK6s8/00check.log. Any idea? Should we skip these checks on CRAN?

mgirlich:
This seems to be due to a breaking change in [testthat](https://github.com/r-lib/testthat/blob/master/NEWS.md#breaking-changes):

We currently use

```r
expect_equal(expect_warning(f(), "warning"), "value")
```

but with the new version we must use

```r
expect_warning(expect_equal(f(), "value"), "warning")
```

PR #637, opened 2021-10-01
Add `duplicates` argument to `rows_insert()`

### LINKED ISSUE #608: Add `rows_insert_missing()` to insert rows not yet in db table
It would be great to have a function to only insert new rows to a db table (basically like the SQL `ON CONFLICT IGNORE`).
Even nicer would be an option to return not only the newly inserted rows but all uploaded rows. For example this is useful when one needs to add an `id` column from the database to local data. But this doesn't work directly with `RETURNING` so this is a bit tricky.

This can be done in two ways:

* using `WHERE NOT EXISTS`
* using a unique constraint and something like `ON CONFLICT IGNORE`

It would be great to support both.

mgirlich:
Closes #608.
Instead of a new verb I added the argument `duplicates` as DavisVaughan proposed in a [dplyr PR](https://github.com/tidyverse/dplyr/pull/5588#issuecomment-722522654). While I was working on this I also realised that `rows_upsert()` is basically just `duplicates = "update"` (and one could also imagine `duplicates = "patch"`).
Still, I prefer the expressiveness of `rows_upsert()` over `rows_insert(duplicates = "update")`. So, maybe a new verb might be better than a new argument.


krlmlr:
Looks good, thanks. Can you please take a look at the conflicts?

mgirlich:
Conflicts are gone 😄

krlmlr:
What was the consensus for `rows_insert()` in dplyr?

krlmlr:
Moving over to the dbplyr PR.

PR #638, opened 2021-10-01
Support `returning` for `in_place = FALSE`

### LINKED ISSUE #611: Support `returning` for `in_place = FALSE`
It's basically possible with one restriction for insert operations (`rows_insert()` and `rows_upsert()`): only columns in `y` can be returned. This is relevant for columns in `x` that have default values or are a sequence (the typical `id` column).
But if one is interested in the `id` column than `in_place = FALSE` doesn't make much sense any way so this should be fine if it is documented.
@krlmlr what do you think about the restrictions?

mgirlich:
Closes #611

### DIFF
--- a/R/rows-db.R
+++ b/R/rows-db.R
@@ -102,7 +102,7 @@ rows_insert.tbl_dbi <- function(x, y, by = NULL, ...,
 
     if (!is_empty(returning_cols)) {
       # Need to `union_all()` with `x` so that all columns of `x` exist in the result
-      returned_rows <- union_all(y, x %>% filter(FALSE)) %>%
+      returned_rows <- union_all(y, x %>% filter(0 == 1)) %>%
         select(returning_cols) %>%
         collect()
       returned_x <- set_returned_rows(returned_x, returned_rows)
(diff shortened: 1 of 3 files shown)


R/rows-db.R
```diff
@@ -148,19 +157,26 @@ rows_update.tbl_dbi <- function(x, y, by = NULL, ...,
...
-    unchanged <- anti_join(x, y, by = by)
-    updated <-
-      x %>%
+    updated <- x %>%
       select(!!!existing_columns) %>%
       inner_join(y, by = by)
 
-    union_all(unchanged, updated)
+    if (is_empty(new_columns)) {
+      returned_x <- x
```
krlmlr, line 166: I tend to name these variables `out` .

krlmlr (approved):
Thanks, awesome! Let's tackle the name of `returning_x` another time.

R/rows-db.R
```diff
@@ -98,7 +97,18 @@ rows_insert.tbl_dbi <- function(x, y, by = NULL, ...,
     if (is_null(check) || is_true(check)) {
       check_db_dupes(x, y, by)
     }
-    union_all(x, y)
+
+    returned_x <- union_all(x, y)
+
+    if (!is_empty(returning_cols)) {
+      # Need to `union_all()` with `x` so that all columns of `x` exist in the result
+      returned_rows <- union_all(y, x %>% filter(0 == 1)) %>%
```
krlmlr, line 105: `0L == 1L` or `0L != 0L` generates slightly shorter SQL, shouldn't matter.

PR #645, opened 2021-10-12
Fix sql_rows_delete() for SQL Server

krlmlr:
Add test first.


krlmlr:
Error observed in test, now gone: https://github.com/cynkra/dm/runs/3869924364#step:13:220

PR #677, opened 2021-10-18
add "on_delete"-support for dm_paste()

### LINKED ISSUE #673: Add on_delete to dm_paste()
It's not covered there yet. Should only be added if different from the `"no_action"` default.

TSchiefer:
in case foreign keys have a non default `on_delete` option, `dm_paste()` picks this up now.

closes #673


krlmlr:
Can you use https://github.com/mxschmitt/action-tmate to troubleshoot the mssql failures?

PR #679, opened 2021-10-18
smaller problems stemming from latest PRs

TSchiefer:
- unnecessary parameter inheritance for `db_schema_...`-functions
- `db_schema_...`-function tests often still used `src` and didn't expect deprecation warning


krlmlr:
Now the db-mssql test starts failing. Let's keep an eye on it.

PR #692, opened 2021-11-01
662 always disambiguate col names

### LINKED ISSUE #662: dm_disambiguate_cols() and compound keys
We need to flesh out the criterion under which we omit a column from disambiguation, here: https://github.com/cynkra/dm/blob/65db4b0b3784cadc902b03c680d1ae5cac0c8be3/R/disambiguate.R#L53-L56

``` r
library(dm)
library(tidyverse)

child <- tibble(x = 1, y = 2)
parent <- tibble(y = 2, x = 3)

dm <-
  dm(child, parent) %>%
  dm_add_pk(child, c(x, y)) %>%
  dm_add_fk(child, y, parent, y)

dm %>%
  dm_disambiguate_cols()
#> Tables: `child`, `parent`
#> Columns: 4
#> Primary keys: 1
#> Foreign keys: 1

dm %>%
  dm_squash_to_tbl(child)
#> # A tibble: 1 × 3
#>     x.x     y   x.y
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
```

<sup>Created on 2021-10-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

moodymudskipper:
closes #662 

We add an argument `exclude_pk = TRUE` to `get_table_colnames()` and we set it to `TRUE` in `dm_disambiguate_cols_impl()`.

`get_table_colnames()` is called also in `prepare_dm_for_flatten()` and there it keeps the default behaviour.

``` r
library(dm, warn.conflicts = FALSE)
library(tidyverse)

person  <- tibble(id = 1, person = "Joe")
hobby   <- tibble(id = c(1,1), hobby = c("R Programming", "Basketball"))
country <- tibble(id = 1, place = "Switzerland")
city    <- tibble(country_id = c(1,1), city = c("Zurich", "Geneva"))

dm <-
  dm(person, hobby, country, city) %>%
  dm_add_pk(person, id) %>%
  dm_add_pk(country, id) %>%
  dm_add_fk(hobby, id, person, id) %>% 
  dm_add_fk(city, country_id, country, id)

dm <- dm_disambiguate_cols(dm)
#> Renamed columns:
#> * id -> person.id, hobby.id, country.id
dm_draw(dm, view_type = "all")
```

<sup>Created on 2021-11-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Note: `get_table_colnames()` has a `tables` argument, `NULL` by default and never used unless set to `NULL`, so it can be removed without affecting the code, is it there for a reason, future plans ? flexibility ? or is it dead code that we should trim off ?

### DIFF
--- a/R/disambiguate.R
+++ b/R/disambiguate.R
@@ -44,19 +44,20 @@ get_table_colnames <- function(dm, tables = NULL, exclude_pk = TRUE) {
     unnest_col("column", character())
 
   if(exclude_pk) {
-  pks <- dm_get_all_pks_def_impl(def)
+    pks <- dm_get_all_pks_def_impl(def)
 
-  keep_colnames <-
-    pks[c("table", "pk_col")] %>%
-    set_names(c("table", "column")) %>%
-    unnest_col("column", character())
+    keep_colnames <-
+      pks[c("table", "pk_col")] %>%
+      set_names(c("table", "column")) %>%
+      unnest_col("column", character())
 
-  table_colnames <-
-    table_colnames %>%
-    # in case of flattening, the primary key columns will never be responsible for the name
-    # of the resulting column in the end, so they do not need to be disambiguated
-    anti_join(keep_colnames, by = c("table", "column"))
+    table_colnames <-
+      table_colnames %>%
+      # in case of flattening, the primary key columns will never be responsible for the name
+      # of the resulting column in the end, so they do not need to be disambiguated
+      anti_join(keep_colnames, by = c("table", "column"))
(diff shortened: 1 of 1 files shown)


moodymudskipper:
Oops, the output isn't right, let me fix this

krlmlr (commented):
Looks good. Can you fix the indent please, or use a shortcut?

krlmlr (approved):
Thanks!

Issue #707, opened 2021-12-09
dm_pixarfilms has unsatisfied constraints


moodymudskipper:
``` r
library(dm, warn.conflicts = FALSE)
dm_examine_constraints(dm_pixarfilms())
#> ! Unsatisfied constraints:
#> • Table `pixar_films`: primary key `film`: has 1 missing values
#> • Table `pixar_people`: primary key `film`     , `role_type`: has duplicate values: Toy Story 4, Storywriter (8), Cars, Screenwriter (6), The Good Dinosaur, Storywriter (5), Brave, Screenwriter (4), Cars 3, Storywriter (4), …
```

<sup>Created on 2021-12-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

AFAII the missing film is for films TBA, in our case it's in 2023, so I don't believe it's a mistake.
We could either : 

* remove the primary keys
* remove the problematic records t
* replace the NA with a new value

`film` is the most important key of the dm so the dm has little value without it. Removing the row seems the easiest. A problem is we don't match {pixarfilms}'s data anymore.

The second issue is simply because one can have several persons with the same role on a movie, so we might have no pk or a pk on all 3 columns. The key doesn't have any child table so removing it seems the best solution to me.

Both issues happened because `dm_add_pk()` was used with `check = FALSE`(default), so I'll use `check=TRUE` in the fixed version.

PR #708, opened 2021-12-09
B707 fix dm pixarfilms

moodymudskipper:
Closes https://github.com/cynkra/dm/issues/707


moodymudskipper:
need to test on windows machine to understand what's going on

krlmlr (commented):
Thanks. We can keep both versions with an option to the function? We could use this as an example for an inconsistent dm, perhaps with auto-repair when we have it?

`check = TRUE` slows down the creation, perhaps let's add a test?

krlmlr:
Let's not spend too much time on it. Getting rid of the duckdb failure seems more urgent.

R/pixarfilms.R
```diff
@@ -44,7 +49,6 @@ dm_pixarfilms <- function(color = TRUE) {
   dm <-
     dm %>%
     dm_add_pk(pixar_films, film) %>%
-    dm_add_pk(pixar_people, c(film, role_type)) %>%
```
krlmlr, line 47: Did you mean to remove this primary key?

moodymudskipper: Yes, this is not unique and shouldn't be, you can have several screenwriters for a given movie for instance.

krlmlr (approved):

krlmlr:
Thanks!

PR #709, opened 2021-12-11
Refactor copy dm to

### LINKED ISSUE #701: Fix compatibility with duckdb 0.3.1
dm uses a strategy (read: ugly hack) for injecting primary and foreign key information during the creation of a table: `PRIMARY KEY` and `FOREIGN KEY` clauses are added to the type of the last column. This kind of works with most backends, but duckdb changed how tables are written in duckdb/duckdb@8aed295921e65ea4320348026e74f461708bc043.

I think we need our own `sql_create_table()` generic that supports primary and foreign key constraints. In `copy_dm_to()` we would switch to that generic.

moodymudskipper:
Closes #701 

Still needs proper tests, tested manually with fixed pixar dm with Postgres sqlite and ducked for temp and permanent tables

```
pixar_dm <- 
  # fetch sample dm 
  dm_pixarfilms() %>% 
  # make it regular
  dm_filter(pixar_films, !is.na(film)) %>% 
  dm_apply_filters() %>% 
  dm_select_tbl(-pixar_people)
```


moodymudskipper:
I couldn't find a nice way to unquote sql names created with `quote_ids()`, as it was needed for `DBI::dbAppendTable` so I used `substr` to remove a character on each side. I think it's robust, just not so pretty.
`dbUnquoteIdentifier()`didn't work on those because they don't have the right class.

Note: {dbplyr}'s doc advices to use `sql()` rather than `ident_q()` but {dm} uses `ident_q()` in several places.

krlmlr:
Thanks. `copy_dm_to()` is used everywhere in the database tests. To replicate locally, use e.g.:

```
DM_TEST_SRC=sqlite R -q -e 'pkgload::load_all(); bad_dm()'
```

For development, you can set the `DM_TEST_SRC` environment variable in the shell with `export` and start RStudio from that shell, so that the setting is available after R is restarted and for all child processes (e.g. tests).

Snapshot tests for the generated SQL would be good. Do we have an intermediate structure (e.g. a named list) that makes it easy to generate such tests?

Happy to review the code when the smoke test and at least one database test pass.

R/db-interface.R
```diff
@@ -201,20 +195,36 @@ copy_dm_to <- function(dest, dm, ...,
...
+  walk(queries$create_table_queries$sql, ~{
+    rs <- DBI::dbSendStatement(dest_con, .x)
+    DBI::dbClearResult(rs)
+  })
+
+  # populate tables
+  pwalk(
+    queries$create_table_queries[c("table", "remote_table")],
+    ~  {
+      DBI::dbAppendTable(dest_con, DBI::SQL(.y), dm[[.x]])
```
krlmlr, line 210: To fix the failures on SQL Server, we need to ensure that `dbExistsTable()` works for the table names we create. The implementation in {odbc} uses `dbUnquoteIdentifier()`: https://github.com/r-dbi/odbc/blob/main/R/Table.R#L204-L224 . As a workaround we could try `dbWriteTable(append = TRUE)`:

```suggestion
      # https://github.com/r-dbi/odbc/issues/480
      DBI::dbWriteTable(dest_con, DBI::SQL(.y), dm[[.x]], append = TRUE)
```

Internally, `odbc::dbAppendTable()` calls `dbWriteTable()` anyway...

krlmlr (commented):
Thanks, looks great!

- Can we now remove `build_copy_data()` ?
- Can you please add snapshot tests for `build_copy_queries()` ?

R/build_copy_queries.R
```diff
@@ -0,0 +1,131 @@
+
+build_copy_queries <- function(con, dm, set_key_constraints = TRUE, temporary = TRUE, table_names) {
+  ## apply filters
+  dm <- dm_apply_filters(dm)
+
+  ## helper to quote all elements of a column and enumerate (concat) element wise
+  quote_enum_col <- function(x) {
+    map_chr(x, ~toString(map_chr(.x, DBI::dbQuoteIdentifier, con = con)))
```
krlmlr, line 8: At least one of the `map_chr()` calls might be unnecessary here?

R/global.R
```diff
@@ -110,5 +110,13 @@ utils::globalVariables(c(
...
   "role_type",
   "award_type",
-  "genre"
+  "genre",
+  #
+  # build_copy_queries
+  "col_def",
+  "fk_def",
+  "index_name",
+  "na.omit",
```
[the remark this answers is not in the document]
moodymudskipper: I used it but changed to use `Filter` instead

R/build_copy_queries.R
```diff
@@ -0,0 +1,131 @@
+
+build_copy_queries <- function(con, dm, set_key_constraints = TRUE, temporary = TRUE, table_names) {
+  ## apply filters
+  dm <- dm_apply_filters(dm)
```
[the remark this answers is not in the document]
moodymudskipper: No

```
library(dm, warn.conflicts = FALSE)
dm_nyc <- dm_nycflights13()
dm_nyc_filtered <-
  dm_nycflights13() %>%
  dm_filter(airports, name == "John F Kennedy Intl")

identical(dm_ptype(dm_nyc), dm_ptype(dm_nyc_filtered))
#> [1] TRUE
```

I took it from build_copy_data, should we remove this line ?

R/build_copy_queries.R
```diff
@@ -0,0 +1,131 @@
...
+  get_sql_col_types <- . %>%
+    tbl_impl(dm, .) %>%
+    DBI::dbDataType(con, .) %>%
+    enframe("col", "type")
+
+  col_defs <-
+    dm %>%
+    src_tbls_impl() %>%
+    set_names() %>%
+    map_dfr(get_sql_col_types, .id = "table") %>%
```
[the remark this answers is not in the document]
moodymudskipper: `.id` is not a parameter of `map`, `get_sql_col_types` gives us two columns data frames that we stack with `map_dfr()` so we have a 3 col df with `table`, `col` and `type`.

R/db-interface.R
```diff
@@ -201,20 +195,36 @@ copy_dm_to <- function(dest, dm, ...,
...
+  walk(queries$index_queries$sql, ~ {
+    rs <- DBI::dbSendStatement(dest_con, .x)
+    DBI::dbClearResult(rs)
+  })
+
+  # build remote dm
+  remote_tables <-
+    queries$create_table_queries$remote_table %>%
+    set_names(queries$create_table_queries$table) %>%
+    map(tbl, src = dest_con)
```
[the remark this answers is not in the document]
moodymudskipper: Which `.vars` arg are you referring to ?

R/build_copy_queries.R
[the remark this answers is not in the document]
krlmlr: Yes, please.

R/db-interface.R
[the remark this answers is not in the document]
krlmlr: It's `vars`, a bit obscure: https://github.com/tidyverse/dbplyr/blob/63f59ace7e381fdc1ac8908cd38a40356613dff2/R/tbl-sql.R#L15

R/build_copy_queries.R
moodymudskipper: Not in general, the outer `map_chr()`  to iterates through rows, the inner `map_chr()` maps through elements contained in the cell which might be a list (not always) and quotes them.

krlmlr:
I misread the code for `odbc::dbWriteTable()` . I wonder how dbplyr solves this internally.

moodymudskipper:
We make indexes unique using make.unique, since only separating tables and columns with `__` wasn't fully robust. We keep the readable names though, which might help the user working on a database. A caveat is that we have long names and if the DBMS has restriction on name length we might get into trouble should we worry about this ?

moodymudskipper:
In order to print the full queries the snapshot tests we simply print the tibbles as data frames since data frames don't trim character columns.

moodymudskipper:
The tests include a snapshot test showing the full build_copy_queries output for the Pixar dm, and a check that we have unique indexes on an artificial dm made to create ambiguous index names.

krlmlr:
The tests still fail, see https://github.com/cynkra/dm/runs/4575296197?check_suite_focus=true.

tests/testthat/test-dm-from-src.R
```diff
@@ -22,6 +22,7 @@ test_that("table identifiers are quoted", {
     )
   })
 
+  skip("FIXME: adapt test")
```
krlmlr, line 25:
```suggestion
```

tests/testthat/test-db-interface.R
```diff
@@ -197,3 +197,60 @@ test_that("copy_dm_to() fails with schema argument for databases other than MSSQ
...
+    variant = if (packageVersion("testthat") > "3.1.0") "testthat-new" else "testthat-legacy",
+    pixar_dm %>%
+      build_copy_queries(
+        src_db,
+        .,
+        table_names = names(.) %>%
+          repair_table_names_for_db(temporary = FALSE, con = src_db, schema = NULL) %>%
+          map(dbplyr::ident_q)) %>%
+      map(as.data.frame) # to print full queries
+  )
```
krlmlr, line 226: Braces are important for srcrefs. I'd like the query to be printed on several lines, we could use `glue::as_glue()` for the `sql` component.

```suggestion
  expect_snapshot({
    variant = if (packageVersion("testthat") > "3.1.0") "testthat-new" else "testthat-legacy",
    pixar_dm %>%
      build_copy_queries(
        src_db,
        .,
        table_names = names(.) %>%
          repair_table_names_for_db(temporary = FALSE, con = src_db, schema = NULL) %>%
          map(dbplyr::ident_q)) %>%
      map(as.list) # to print full queries
  })
```

R/build_copy_queries.R
```diff
@@ -0,0 +1,143 @@
...
+  graph <- create_graph_from_dm(dm, directed = TRUE)
+  topo <- igraph::topo_sort(graph, mode = "in")
+  idx <- match(names(topo), create_table_queries$table)
+
+  if (length(idx) == nrow(create_table_queries)) {
+    create_table_queries <- create_table_queries[idx, ]
+  }
+
+  ## Return a list of both type of queries
+  lst(create_table_queries, index_queries)
```
krlmlr, line 142: Can we return a tibble here, with one row per table, and perhaps columns `name`, `columns` (list of character), `sql_table` (character), `sql_index` (character)?

R/db-interface.R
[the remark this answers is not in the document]
krlmlr: - Use `match()` below instead of setting names here and later unnaming
- Use `map2()` with table names and column names

```suggestion
  map(..., ~ tbl(dest_con, .x, vars = .y))
```

krlmlr (commented):
Thanks, we're getting there!

krlmlr:
Tests should be good now. Would you like to take another stab?

tests/testthat/test-db-interface.R
moodymudskipper: With `as.list()` instead of `as.data.frame` queries are printed on several lines already because they have the SQL class and printing method.

krlmlr:
Tests still fail, do you need to update the snapshots?

moodymudskipper:
I'm still fixing other things

Le lun. 27 déc. 2021 à 12:13, Kirill Müller ***@***.***> a
écrit :

R/build_copy_queries.R
moodymudskipper: done, it contains a couple additional column, `remote_name` is needed for populating tables and `index_name` to check for duplicate indexes. `sql_index` is not a character but a character list since a table can have several index queries in general.

krlmlr:
Looks like we do need test variants for the different database versions. I can take this on, unless you have started already.

moodymudskipper:
you can take it

krlmlr (dismissed):

krlmlr (approved):

krlmlr:
Thanks!

PR #713, opened 2021-12-11
Switch to actions v2

krlmlr:
CC @gaborcsardi: I have defined a R CMD check workflow that consists of only a few composite actions. This simplifies smoke testing: run only one element of the test matrix, the other elements are started only after the smoke test is successful. Are there plans in r-lib/actions to move in a similar direction?

### DIFF
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -26,7 +26,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      - uses: r-lib/actions/setup-r@v1
+      - uses: r-lib/actions/setup-r@v2
         with:
           install-r: false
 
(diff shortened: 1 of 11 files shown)


gaborcsardi:
I am not sure if you need anything from `r-lib/actions` for this, I think you can use the actions as is to create workflows with smoke tests.

If you mean having example workflows with smoke tests, we can consider that. OTOH IDK if there is a good smoke test for R packages, each OS might fail for some reason that does not affect others.

krlmlr:
Thanks. This is what I did.

More often than not my checks fail on all OSes, a smoke test on any OS helps here.

gaborcsardi:
Many people run `R CMD check` before committing/pushing and that serves as a smoke test. But yeah, we can have an example workflow for this, can you open an issue please?

krlmlr:
Integrated now.

PR #714, opened 2021-12-17
Add `na_matches` argument to `rows_*()`

### LINKED ISSUE #619: `rows_*()`: add `na_matches` argument

mgirlich:
Closes #619


krlmlr:
Thanks. I saw check errors with the last two test runs, would you like to take another look?

krlmlr:
Moving over to the dbplyr PR.

PR #722, opened 2021-12-28
`pack_join()`

### LINKED ISSUE #721: pack_join
For #595

moodymudskipper:
closes #721


krlmlr:
Thanks. A few basic tests would be great.

krlmlr:
And we also need to add to `_pkgdown.yml` .

moodymudskipper:
Edited to support `copy` arg.

Do we need `.dm` and `.zoomed_dm` methods ?

krlmlr (approved):
Thanks, looks good!

krlmlr:
Do we need the zoomed methods for implementing the serialization? If not, let's leave it for now.

moodymudskipper:
No we don't need it for the serialisation. 

What we might need for the serialisation though, if we want to keep the option full join the parent tables is to have a `full = FALSE` arg to `pack_join`. or a `full_pack_join()` function. What do you think ?

krlmlr:
`nest_join()` doesn't have it yet either. Let's keep it simple for now.

krlmlr:
Thanks!

Issue #672, opened 2021-10-14
Improve db_schema_*() API


krlmlr:
- [ ] Avoid proactive checks if a schema exists, let the database determine that check and fail accordingly
- [ ] Use `dm_meta()` from #517 for `db_schema_list()` and `db_schema_exists()`

Follow-up to https://github.com/cynkra/dm/issues/668#issuecomment-942958596.

PR #771, opened 2022-01-27
no proactive checks for schema existence and tweak MSSQL-schema-code

TSchiefer:
partly addresses #672: the proactive checks for schema existence are abolished.

For listing schemas on MSSQL we also now avoid matching `schema owner id` to `user id`. I just ran into trouble here (`schema owner id` didn't exist in `user id` list and therefore schema wasn't listed, even though it existed). 
Now used 3054912 as a workaround, that schemas that exist by default for backward compatibility are filtered out.

### DIFF
--- a/R/error-helpers.R
+++ b/R/error-helpers.R
@@ -594,19 +594,6 @@ warn_if_arg_not <- function(arg,
 
 # Errors for schema handling functions ------------------------------------
 
-abort_no_schema_exists <- function(schema, dbname = NULL) {
-  abort(error_txt_no_schema_exists(schema, dbname),
-    class = dm_error_full("no_schema_exists")
-  )
-}
-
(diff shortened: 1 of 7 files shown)


TSchiefer:
Test failures seem to be fairly independent of the changes in the PR

krlmlr:
Thanks. How is this tested?

TSchiefer:
> How is this tested?

I am not sure how to reach this state, that the principal_id for a schema is not available in the user_id list. Apparently it is possible, since I've come across it, but I haven't found a way yet to get there. I have tested it in this specific setting though, and it worked...

(sorry BTW for the closing of the PR, misclicked)

krlmlr (approved):
Thanks!

krlmlr:
Please merge with a bullet point in the merge commit message.

Issue #817, opened 2022-01-30
Consistent API for new functions


moodymudskipper:
Needed for  #223

There is a lot to say, so sorry for writing a lot.

I'll try to be as terse as I can and answer in several parts so we can fold some as resolved for clarity. 

We discuss : 

* nest/pack functions' naming (the most urgent part)
* renaming dm_wrap()
* dm level function based on `reunite_parent_child()`
* dm level function based on `decompose_table()`
* dm level dm_flatten_to_tbl()
* changes to the nest/pack API for consistency

## review of the naming conventions

We can separate dm's functions by suffix :

* `dm_verb_tbl()` : `dm_add_tbl()`, `dm_rm_tbl()`, 
  `dm_mutate_tbl()`, `dm_select_tbl()`, `dm_rename_tbl()`
  `dm_nest_tbl()`, `dm_unnest_tbl()`, `dm_pack_tbl()`, `dm_unpack_tbl()`, 
   (unexported: `dm_patch_tbl()`),
* `dm_verb()` (with dplyr verbs): `dm_filter()`, `dm_mutate()`, `dm_select()`
* `dm_verb_to_tbl()` : `dm_apply_filters_to_tbl()`, `dm_flatten_to_tbl()`, 
  `dm_join_to_tbl()`, `dm_squash_to_tbl()`
  
These correspond to categories of behaviours :

* `dm_verb_tbl()` : These functions target a table and change the dm's prototype, 
they might or might not modify tables. The only counter example might be `dm_patch_tbl()` which doesn't restructure the dm per say, but it replaces the tables entirely so can still be
considered a "dm" op rather than a table op. Anyway it is unexported so consistency is less
important there.
* `dm_verb()` : These functions target a table in the dm, and modify it without affecting
  the structure of the dm
* `dm_verb_to_tbl()`: These functions all return a tibble

The functions `reunite_parent_child()` and `decompose_table()` don't follow these formats,
but we want the analogues that take and return a dm to follow a consistent api.

moodymudskipper:
## should we rename new nest/pack functions ?

Since these do a join should we mention the join ?
To me, looking at the dm structure level, we are nesting a table into another,
so talking about "join" is not necessarily useful, it's an implementation detail.
Just like `dm_mutate_tbl()` and `dm_select_tbl()` don't really do dplyr style
mutating not selection, `dm_nest_tbl()` doesn't do dplyr style nesting. 
A function `dm_nest()` would nest a specific table.

[...]

PR #820, opened 2022-02-02
Improve API consistency

krlmlr:
For #817.


moodymudskipper:
Regarding plural tables this doesn't quite fit what I had in mind.

I was thinking of packing or wrapping into several tables, so the "into" was the potentially plural argument in my idea. This way we could pack a parent table into its different children and handle the cycles.
Being able to pack several parent tables at the same time is mostly useful to save typing I think, but could be done in several steps. 

So I thought `dm_unnest_tbl()` and `dm_unpack_tbl()` were the ones that needed their `table` argument pluralised (and specified as parent/child) and `dm_nest_tbl()`, `dm_pack_tbl()` would keep a singular argument, but renamed as parent/child for clarity.

krlmlr:
Ouch. You're right. I'll fix in the CRAN branch.

krlmlr:
Maybe let's take some time to discuss this.

PR #827, opened 2022-02-08
Build sequence of actions first and then execute

### LINKED ISSUE #752: Split planning and doing in dm_wrap()
Recursively traverse the relationship graph to create a recipe that is then applied in one swoop (e.g. via `purrr::reduce2()`).

This allows for earlier detection of failure and better tests.

moodymudskipper:
closes #752


krlmlr (dismissed):
We can fix #864 on top of this PR if it helps.

R/dm_wrap.R
```diff
@@ -41,34 +41,42 @@ dm_wrap_tbl <- function(dm, root, strict = TRUE) {
...
       graph <- igraph::delete.vertices(graph, parent_name)
       positions <- node_type_from_graph(graph, drop = root_name)
     }
     if (!has_terminal_child && !has_terminal_parent) break
   }
 
+  wrapped_dm <- purrr::reduce2(
+    wrap_sequence$action,
+    wrap_sequence$table,
+    function(dm, f, table) eval(rlang::call2(f, dm, table)),
```
krlmlr, line 67: Would `rlang::exec()` work here?

krlmlr (commented):
Thanks. Can you please make the split explicit by extracting two functions -- one that creates the recipe and one that applies it? We will work on the recipe-creating function later.

krlmlr (approved):
Thanks. I meant also to extract the `reduce2()` call into another function, let's tackle this in a separate PR.

PR #852, opened 2022-03-10
check_suggested() is memoised, except for testthat

krlmlr:
Suggested messages are given only once per session, per message.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -10,6 +10,10 @@
 
   dm_has_financial <<- memoise::memoise(dm_has_financial, cache = cache_attach())
 
+  if (!identical(Sys.getenv("TESTTHAT"), "true")) {
+    check_suggested <<- memoise::memoise(check_suggested)
+  }
+
   register_pkgdown_methods()
 
   s3_register("waldo::compare_proxy", "dm")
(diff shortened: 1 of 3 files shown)


R/zzz.R
```diff
@@ -10,6 +10,10 @@
 
   dm_has_financial <<- memoise::memoise(dm_has_financial, cache = cache_attach())
 
+  if (!identical(Sys.getenv("TESTTHAT"), "true")) {
+    check_suggested <<- memoise::memoise(check_suggested)
+  }
```
TSchiefer, line 15: It looks like the env variable "TESTTHAT" is set only at a later stage. The function is thus memoised even for the tests.

krlmlr: Making it unconditional then.

TSchiefer (approved):
LGTM

R/zzz.R
moodymudskipper: We can also define `check_suggested <- environment(check_suggested)$'_f'` in helper scripts, or keep an unmemoized copy in the namespace if we want to avoid using implementation details.

moodymudskipper (commented):
LGTM, just pointing out a way to have check_suggested unmemoized for tests

R/zzz.R
krlmlr: Good idea, it seems that the tests work well enough even when always memoizing.

PR #853, opened 2022-03-10
Use default width for progress bar

krlmlr:
Why was it set to 60 @moodymudskipper @TSchiefer?


TSchiefer (approved):
LGTM, failing tests are independent of this

moodymudskipper:
I see no reason. The `clear = FALSE` seems arbitrary too

Issue #850, opened 2022-03-09
Progress bar for copy_dm_to()


krlmlr:
no longer shows, most likely after we changed to using `db_append_table()` .

Can we use this opportunity to improve the code to:

- send the table data in chunks (of e.g. 1 MB or 10 MB size each)
- precompute the total number of chunks
- show the progress bar in terms of chunks sent?

krlmlr:
On it now.

PR #855, opened 2022-03-10
Progress for copy_dm_to(), speed up writes for SQL server

krlmlr:
For #850.

### DIFF
--- a/R/db-interface.R
+++ b/R/db-interface.R
@@ -197,21 +197,42 @@ copy_dm_to <- function(dest, dm, ...,
 
   queries <- build_copy_queries(dest_con, dm, set_key_constraints, temporary, table_names_out)
 
+  ticker_create <- new_ticker(
+    "creating tables",
+    n = length(queries$sql_table),
+    progress = progress,
+    top_level_fun = "copy_dm_to"
+  )
+
   # create tables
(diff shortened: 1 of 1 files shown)


R/db-interface.R
```diff
@@ -242,19 +263,31 @@ check_naming <- function(table_names, dm_table_names) {
...
+      paste0("inserting into ", remote_table),
+      n = n_chunks,
+      progress = progress,
+      top_level_fun = top_level_fun
+    )
+
+    walk(seq_len(n_chunks), ticker(~ {
+      end <- .x * chunk_size
+      idx <- seq2(end - (chunk_size - 1), min(end, nrow(table)))
+      values <- as_tibble(map(table[idx, ], dbplyr::escape, parens = FALSE, collapse = NULL, con = con))
```
TSchiefer, line 286: I believe the `as_tibble()`-call is redundant, since you are converting `values` to a list again the only time you're using it later on

TSchiefer (approved):
nice

R/db-interface.R
krlmlr: Good catch!

TSchiefer (approved):
LGTM

PR #857, opened 2022-03-11
Fix column tracking for foreign keys

### LINKED ISSUE #856: Column renames are not reflected in foreign keys
Our tests don't catch this, this isn't detected by `validate_dm()` either.

``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter
dm_nycflights13() %>% 
  dm_rename(airports, airport_id = faa) %>% 
  dm_get_all_fks() %>% 
  filter(parent_table == "airports")
#> # A tibble: 1 × 5
#>   child_table child_fk_cols parent_table parent_key_cols on_delete
#>   <chr>       <keys>        <chr>        <keys>          <chr>    
#> 1 flights     origin        airports     faa             no_action
```

<sup>Created on 2022-03-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
@TSchiefer @moodymudskipper: Can you please finish this?

- [x] Tests
- [x] Ensure that `dm_validate()` catches this condition

Closes #856.

### DIFF
--- a/R/validate.R
+++ b/R/validate.R
@@ -39,12 +39,21 @@ validate_dm <- function(x) {
 
   dm_col_names <- set_names(map(def$data, colnames), table_names)
 
-  fks <- c_list_of(def$fks)
+  fks <-
+    def %>%
+    select(ref_table = table, fks) %>%
+    unnest_list_of_df("fks")
+
   check_fk_child_tables(fks$table, table_names)
 
   fks %>%
     unnest_col("column", character()) %>%
     check_colnames(dm_col_names, "FK")
+
+  fks %>%
+    unnest_col("ref_column", character()) %>%
+    select(table = ref_table, column = ref_column) %>%
+    check_colnames(dm_col_names, "Parent key")
 
   pks <-
     def %>%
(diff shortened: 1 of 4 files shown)


krlmlr:
Thanks, good catch. I'll review locally too.

R/validate.R
```diff
@@ -39,13 +39,22 @@ validate_dm <- function(x) {
...
 
   fks %>%
     unnest_col("column", character()) %>%
     check_colnames(dm_col_names, "FK")
 
+  fks %>%
+    unnest_col("ref_column", character()) %>%
+    select(table = ref_table, column = ref_column) %>%
+    check_colnames(dm_col_names, "Parent key")
+
```
krlmlr, line 57: Do we need to keep the original code too, because the fks have table and ref-table? (How do we test `dm_validate()` ?)

moodymudskipper: I thought both would work this way but better test indeed. I think the easiest is to take a valid dm, edit its data artificially then test with expect_error, I ll update the PR

moodymudskipper: 2 tests were added to test for respective existence of `column` and `ref_column`

Issue #751, opened 2022-01-19
Wrapping dm objects with circular references


krlmlr:
Doesn't work yet after #737. I wonder if it even should.

Ultimately we want a way to wrap and unwrap `dm_nycflights13(cycle = TRUE)` and other similar constructs. One way to achieve this is to create two "views" for the `airports` table. I remember we had an issue that describes this operation, but haven't looked.

Preprex:

``` r
library(dm)

dm_nycflights13(cycle = TRUE) %>%
  dm_treeify()
```

![](https://i.imgur.com/FZ6rJYk.png)

<sup>Created on 2022-01-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

The inverse operation could again take a dm ptype as argument.

TSchiefer:
> I remember we had an issue that describes this operation, but haven't looked.

Think it was this one: #55

krlmlr:
Fixed with `dm_disentangle()` when it's exported.

PR #862, opened 2022-03-17
implement dm_disentangle()

### LINKED ISSUE #55: New operation: disentangle

TSchiefer:
I know this wasn't high priority, but I felt like taking a stab at this once more.

It seems to work also in corner cases: 
- dm whose graph has less or equal edges than vertices, but one or more of the components has a cycle
- no disentangling when multiple incoming edges but no cycle

Details:
- the newly created tables have names: `<old_name>_n` (with n = 1, 2, ...); should one or more of those new names already exist, a name repair would take place (`repair = "unqiue"`), which leads to admittedly rather ugly results (e.g. `<old_name>_n...1`); I am not sure how to best implement it for consistency reasons; we could allow the user to provide a naming template
- Implementation: 
  1. graph representation of dm is broken down in its independent components
  1. each component is tested for cycles
  1. if at least one cycle is detected, all tables in components with cycles are checked for multiple incoming FKs
  1. those tables that have 2 or more incoming FKs are disentangled

``` r
suppressMessages(devtools::load_all("~/git/cynkra/dm"))
suppressPackageStartupMessages(library(tibble))

# no action if no cycle detected
dm_disentangle(dm_nycflights13())
#> No cycle detected, returning original `dm`.
#> Tables: `airlines`, `airports`, `flights`, `planes`, `weather`
#> Columns: 53
#> Primary keys: 4
#> Foreign keys: 4

# creating a dm consisting of 4 independent parts, 2 of which contain a cycle
t8 <- tibble(r = 1:3)
t9 <- tibble(s = c(2L, 2L, 3L))
t10 <- tibble(t = c(1L, 1L, 2L))
three_part_cycle_dm <- dm_add_tbl(dm_for_filter_w_cycle(), t8, t9, t10) %>% 
  dm_add_pk(t8, r) %>% 
  dm_add_fk(t9, s, t8) %>% 
  dm_add_fk(t10, t, t8) %>% 
  dm_bind(dm_nycflights_small_cycle())
dm_get_all_fks(three_part_cycle_dm)
#> # A tibble: 13 × 5
#>    child_table child_fk_cols parent_table parent_key_cols on_delete
#>    <chr>       <keys>        <chr>        <keys>          <chr>    
#>  1 tf_2        d             tf_1         a               no_action
#>  2 tf_7        q             tf_2         c               no_action
#>  3 tf_2        e, e1         tf_3         f, f1           no_action
#>  4 tf_4        j, j1         tf_3         f, f1           no_action
#>  5 tf_5        l             tf_4         h               cascade  
#>  6 tf_5        m             tf_6         n               no_action
#>  7 tf_6        o             tf_7         p               no_action
#>  8 t9          s             t8           r               no_action
#>  9 t10         t             t8           r               no_action
#> 10 flights     carrier       airlines     carrier         no_action
#> 11 flights     dest          airports     faa             no_action
#> 12 flights     origin        airports     faa             no_action
#> 13 flights     tailnum       planes       tailnum         no_action

dm_disentangle(three_part_cycle_dm) %>% 
  dm_get_all_fks()
#> # A tibble: 13 × 5
#>    child_table child_fk_cols parent_table parent_key_cols on_delete
#>    <chr>       <keys>        <chr>        <keys>          <chr>    
#>  1 tf_2        d             tf_1         a               no_action
#>  2 tf_7        q             tf_2         c               no_action
#>  3 tf_5        l             tf_4         h               cascade  
#>  4 tf_5        m             tf_6         n               no_action
#>  5 tf_6        o             tf_7         p               no_action
#>  6 t9          s             t8           r               no_action
#>  7 t10         t             t8           r               no_action
#>  8 flights     carrier       airlines     carrier         no_action
#>  9 flights     tailnum       planes       tailnum         no_action
#> 10 tf_2        e, e1         tf_3_1       f, f1           no_action
#> 11 tf_4        j, j1         tf_3_2       f, f1           no_action
#> 12 flights     dest          airports_1   faa             no_action
#> 13 flights     origin        airports_2   faa             no_action
  
# dm_draw does unfortunately not work (anymore?) with reprex(), since html output is not allowed
```

<sup>Created on 2022-03-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

closes #55


TSchiefer:
in addition to the question about naming I also need to transfer the `on_delete` values

TSchiefer:
It looks now to me that it's still doing a little too much in some cases.
If there are 2+ incoming FKs it doesn't necessarily mean it needs to be disentangled, see below (maybe use dm_draw() instead of dm_get_all_fks()).
I think the "too much" is always related to a split of the graph in an additional component. Maybe that could help solving this.

``` r
suppressMessages(devtools::load_all("~/git/cynkra/dm"))
suppressPackageStartupMessages(library(tibble))

dm_for_filter_w_cycle() %>% 
  dm_add_tbl(tf_8 = tibble(r = 1)) %>% 
  dm_add_fk(tf_8, r, tf_1) %>% 
  dm_disentangle() %>% 
  dm_get_all_fks()
#> # A tibble: 8 × 5
#>   child_table child_fk_cols parent_table parent_key_cols on_delete
#>   <chr>       <keys>        <chr>        <keys>          <chr>    
#> 1 tf_7        q             tf_2         c               no_action
#> 2 tf_5        l             tf_4         h               cascade  
#> 3 tf_5        m             tf_6         n               no_action
#> 4 tf_6        o             tf_7         p               no_action
#> 5 tf_2        d             tf_1_1       a               no_action
#> 6 tf_8        r             tf_1_2       a               no_action
#> 7 tf_2        e, e1         tf_3_1       f, f1           no_action
#> 8 tf_4        j, j1         tf_3_2       f, f1           no_action
```

<sup>Created on 2022-03-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

TSchiefer:
status update:
1. `on_delete` is transferred to the new relations
1. `quiet` arg: if `FALSE`, names of new tables are messaged
1. `naming_template` arg: can make use of:
    - original parent table name (`.pt`)
    - primary key column names of parent table (`.pkc`; if compound key, the column names will be separated by underscore
    - child table name (`.ct`)
    - foreign key colum names (`.fkc`)
    - number of newly created table based on each parent table (`.ntn`; from "new table number"): if a parent table is split into three you get `1,2,3`.
1. the "doing too much"-problem is solved with a graph theoretical examination: 
    - first, the components of the graph which have a cycle are determined
    - next, the parent tables within that graph are determined, which have 2+ incoming FKs
    - then it is tested for each pair of tables participating in each FK relation, how many "simple paths" there are between the two tables. If it is just one path, then for that relation the parent table doesn't have to be replicated.
    - if there is at least one component with an "endless cycle" (meaning a cycle in a directed graph that you can keep on walking in the direction of the arrows endlessly), the function is a no-op and the original `dm` is returned.
1. in case of a very convoluted `dm` (cf. `entangled_dm()`), there might be more iterations of disentangling required. Therefore the function now checks in the end if there are still any components with cycles and calls itself on the new `dm` if there are.
1. tests are in place
1. added to _pkgdown.yml

krlmlr:
Thanks. I like the behavior, it's an important macro that will help us a lot in the medium term.

For `dm_wrap_tbl()` we have designed it as a composition of individual operations (`dm_nest_tbl()` and `dm_pack_tbl()`) from the start. Before merging this, I would like to check if we can follow a similar approach for `dm_disentangle()` . I'm envisioning a single operation (let's call it `dm_clone_tbl()` for now, tbd) so that `dm_disentangle()` is a repeated application of `dm_clone_tbl()` .

Cloning would accept a table argument as input and create one table in the dm for each foreign key into that table. Similarly to `dm_nest_tbl()` and `dm_pack_tbl()`, it's dm in, dm out. The main work that `dm_disentangle()` then has to do is to figure out which tables need to be cloned, and in what order.

Is it feasible to work on this pull request to extract an explicit `dm_clone_tbl()` operation, perhaps in a separate `.R` file? This would be easier to review and to merge, and is already useful even without `dm_disentangle()` .

Please also review with @moodymudskipper.

TSchiefer:
I think that's a very good idea and it should be feasible under the condition that we add an (optional) argument e.g. `included_child_tables` (or similar), that defaults to `NULL` which results in the behaviour described by you. This is necessary, because in general not all of the incoming FKs warrant a duplication of the PT.

TSchiefer:
After our discussion my next steps would be:
1. add replication of all tables that are "upstream" (against direction of FKs) from a parent table that is to be replicated, but are not part of a cycle
1. add replication of all tables that are "downstream" (in direction of FKs) from a parent table that is to be replicated

Potentially I would try the idea of providing code for sequential "atomic clone" calls, but I'd like to see how `dm_disentangle()` works with said improvements.

TSchiefer:
@krlmlr @moodymudskipper I had another idea how to call the atomic operation function: `dm_rep_parent_tbl()`
I think it might be suitable, because `rep()` has a slightly similar behavior, even though the number of replications cannot be provided by the user in our case.
Really unsure about the inverse function in that case though, maybe `dm_fuse_parent_tbls()`?

moodymudskipper:
`dm_rep_parent_tbl()` is fine by me, though I'd go for the full `dm_replicate_parent_tbl()` name for clarity and no confusion with `repair`. Since it's not a common function I think we can afford a long name.

I'm less sold on `fuse` and in general pure synonyms of "merge", maybe `unify` ? It's a bit more specific.

The other day we were also thinking about do we need a name that describe the proximate/technical vs ultimate/conceptual action, the above is more proximate. This gift me thinking today about `tie()` / `untie()`, `knot()` / `unknot()`, `lace()` / `unlace()`.  If we see cycles as creating knots in the diagram this is quite intuitive.

I'm also wondering if we'd be better off having the general function `dm_disentangle` be named the same as the atomic function, with a `_all` suffix. This wouldn't work with `dm_wrap_tbl()` because we're both packing and nesting there, but in the current case it might be clearer to avoid multiplying names.

krlmlr:
Does it help if we designate a start node, enumerate all cycle-free paths from that start node and create a tree-like structure from that? This might help for both disentangling and wrapping (#751). `igraph::all_simple_paths()` doesn't help right away, we need to enumerate edges sequences. I tried constructing the [line graph](https://en.wikipedia.org/wiki/Line_graph) to enumerate edge sequences, but this gives too many paths.

I have drafted a sketch, this still needs some postprocessing, but the table names and versions should be correct. Does it work for our "enfant terrible"?

``` r
library(tidyverse)
library(dm)

graph <-
  dm_nycflights13(cycle = TRUE) %>%
  dm:::create_graph_from_dm(directed = TRUE)

graph_df <-
  graph %>%
  igraph::as_data_frame() %>%
  group_by(from, to) %>%
  mutate(idx = row_number()) %>%
  ungroup()

graph_df_ud <-
  graph_df %>%
  rename(from = to, to = from) %>%
  bind_rows(graph_df)

enumerate_all_paths <- function(node, path = character()) {
  if (length(path) > 0) {
    tbl_count[[node]] <<- (tbl_count[[node]] %||% 0) + 1
    usage_idx <- paste0("-", tbl_count[[node]])
  } else {
    usage_idx <- ""
  }
  path <- c(path, set_names(node, paste0(node, usage_idx)))

  out <-
    graph_df_ud %>%
    filter(from == !!node) %>%
    filter(!(to %in% path)) %>%
    select(to, idx)

  if (nrow(out) == 0) {
    print(names(path))
    return()
  }

  pwalk(out, ~ {
    enumerate_all_paths(.x, path)
  })
}

tbl_count <- list()
enumerate_all_paths("flights")
#> [1] "flights"    "airlines-1"
#> [1] "flights"    "airports-1"
#> [1] "flights"    "airports-2"
#> [1] "flights"  "planes-1"
#> [1] "flights"   "weather-1"
tbl_count <- list()
enumerate_all_paths("airlines")
#> [1] "airlines"   "flights-1"  "airports-1"
#> [1] "airlines"   "flights-1"  "airports-2"
#> [1] "airlines"  "flights-1" "planes-1" 
#> [1] "airlines"  "flights-1" "weather-1"
tbl_count <- list()
enumerate_all_paths("airports")
#> [1] "airports"   "flights-1"  "airlines-1"
#> [1] "airports"  "flights-1" "planes-1" 
#> [1] "airports"  "flights-1" "weather-1"
#> [1] "airports"   "flights-2"  "airlines-2"
#> [1] "airports"  "flights-2" "planes-2" 
#> [1] "airports"  "flights-2" "weather-2"
```

<sup>Created on 2022-04-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

moodymudskipper:
Do we need a `disentangle` or `break_cycles` argument in `dm_wrap_tbl(dm, root, strict = TRUE)`.
Should I start working on this in another PR ?

krlmlr:
If we get `dm_disentangle()` and the reverse operation to work, we don't need to change `dm_wrap_tbl()` for now. Let's wait a bit.

krlmlr:
Next iteration, without igraph and printing each edge of the new data model only once. Next steps: use event-based approach similar to https://github.com/r-lib/pillar/blob/main/R/ctl_colonnade.R .

``` r
library(tidyverse)
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter

all_fks <-
  dm_nycflights13(cycle = TRUE) %>%
  dm_get_all_fks() %>%
  rename(child_cols = child_fk_cols, parent_cols = parent_key_cols) %>%
  select(-on_delete)

graph_df_ud <-
  all_fks %>%
  rename(
    child_table = parent_table,
    child_cols = parent_cols,
    parent_table = child_table,
    parent_cols = child_cols,
  ) %>%
  bind_rows(all_fks) %>%
  group_by(child_table, parent_table) %>%
  mutate(idx = row_number()) %>%
  ungroup()

enumerate_all_paths <- function(node, path = character()) {
  if (length(path) > 0) {
    tbl_count[[node]] <<- (tbl_count[[node]] %||% 0) + 1
    usage_idx <- paste0("-", tbl_count[[node]])
    print(c(names(path)[[length(path)]], paste0(node, usage_idx)))
  } else {
    usage_idx <- ""
  }
  path <- c(path, set_names(node, paste0(node, usage_idx)))

  out <-
    graph_df_ud %>%
    filter(child_table == !!node) %>%
    filter(!(parent_table %in% path)) %>%
    select(parent_table, idx)

  if (nrow(out) == 0) {
    # print(names(path))
    return()
  }

  pwalk(out, ~ {
    enumerate_all_paths(.x, path)
  })
}

tbl_count <- list()
enumerate_all_paths("flights")
#> [1] "flights"    "airlines-1"
#> [1] "flights"    "airports-1"
#> [1] "flights"    "airports-2"
#> [1] "flights"  "planes-1"
#> [1] "flights"   "weather-1"
tbl_count <- list()
enumerate_all_paths("airlines")
#> [1] "airlines"  "flights-1"
#> [1] "flights-1"  "airports-1"
#> [1] "flights-1"  "airports-2"
#> [1] "flights-1" "planes-1" 
#> [1] "flights-1" "weather-1"
tbl_count <- list()
enumerate_all_paths("airports")
#> [1] "airports"  "flights-1"
#> [1] "flights-1"  "airlines-1"
#> [1] "flights-1" "planes-1" 
#> [1] "flights-1" "weather-1"
#> [1] "airports"  "flights-2"
#> [1] "flights-2"  "airlines-2"
#> [1] "flights-2" "planes-2" 
#> [1] "flights-2" "weather-2"
```

<sup>Created on 2022-04-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

TSchiefer:
closing, because #923 was preferable

Issue #864, opened 2022-03-28
dm_wrap_tbl() fails if column and table have the same name


krlmlr:
``` r
library(dm)
library(tibble)

dm(a = tibble(b = integer()), b = tibble(b = integer())) %>%
  dm_add_fk(a, b, b, b) %>%
  dm_wrap_tbl(a)
#> New names:
#> * b -> b...1
#> * b -> b...2
#> Error in `check_join_vars()`:
#> ! Join columns must be present in data.
#> x Problem with `b`.
```

<sup>Created on 2022-03-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This seems to be a useful naming scheme. @moodymudskipper: what would it take to support this?

moodymudskipper:
There was a FIXME in pack_join :) 

    # FIXME: handle potential conflict between name_var and existing variables

PR #865, opened 2022-03-28
pack_join handle names robustly and consistently with nest_join

moodymudskipper:
closes https://github.com/cynkra/dm/issues/864

There are 2 possible conflicts, within `b` when (self) packing, which should be invisible to the user, and within `a` where I would have preferred to fail but for consistently with `dplyr::nest_join` we overwrite the column silently. 2 new snapshot tests illustrate this.


krlmlr (approved):

krlmlr:
Thanks!

PR #923, opened 2022-04-13
WIP: dm_disentangle() based on enumerate_all_paths()

### LINKED ISSUE #55: New operation: disentangle

TSchiefer:
I decided to start a new PR regarding this issue. I will close the other PR as soon as we can be sure this here is what we want.

Now I am using the approach suggested in https://github.com/cynkra/dm/pull/862#issuecomment-1095260560
I already implemented a first version of `dm_disentangle()` that is based on the function `enumerate_all_paths()` here, which provides a recipe for the new sets of tables and key relations.
The results are very different from what we got with the version of `dm_disentangle()` from #862, but they nevertheless seem to make a lot of sense. 
Also, that was to be expected, since now we don't try to resolve the cycles by only duplicating the parent tables. 
The results vary strongly depending on which `start` table you use (which was also to be expected).

I am storing intermediate results in a specific environment.
I believe I might have not quite captured the logic/approach from https://github.com/r-lib/pillar/blob/main/R/ctl_colonnade.R, but it worked quite well with the current approach. If there are downsides to my approach please let me know.

Example for the current state of things:
``` r
suppressPackageStartupMessages(library(dm))

cycle_dm <- dm_nycflights13() %>% 
  dm_add_fk(flights, dest, airports, on_delete = "cascade")

dm:::enumerate_all_paths(cycle_dm, "flights")
#> # A tibble: 2 × 7
#>   child_table child_cols parent_table parent_cols new_child_table
#>   <chr>       <keys>     <chr>        <keys>      <chr>          
#> 1 flights     origin     airports     faa         flights        
#> 2 flights     dest       airports     faa         flights        
#> # … with 2 more variables: new_parent_table <chr>, on_delete <chr>
dm:::dm_disentangle(cycle_dm, "flights") %>% 
  dm_get_all_fks()
#> # A tibble: 5 × 5
#>   child_table child_fk_cols     parent_table parent_key_cols   on_delete
#>   <chr>       <keys>            <chr>        <keys>            <chr>    
#> 1 flights     carrier           airlines     carrier           no_action
#> 2 flights     tailnum           planes       tailnum           no_action
#> 3 flights     origin, time_hour weather      origin, time_hour no_action
#> 4 flights     origin            airports-1   faa               no_action
#> 5 flights     dest              airports-2   faa               cascade

dm:::enumerate_all_paths(cycle_dm, "airports")
#> # A tibble: 8 × 7
#>   child_table child_cols        parent_table parent_cols       new_child_table
#>   <chr>       <keys>            <chr>        <keys>            <chr>          
#> 1 flights     origin            airports     faa               flights-1      
#> 2 flights     carrier           airlines     carrier           flights-1      
#> 3 flights     tailnum           planes       tailnum           flights-1      
#> 4 flights     origin, time_hour weather      origin, time_hour flights-1      
#> 5 flights     dest              airports     faa               flights-2      
#> 6 flights     carrier           airlines     carrier           flights-2      
#> 7 flights     tailnum           planes       tailnum           flights-2      
#> 8 flights     origin, time_hour weather      origin, time_hour flights-2      
#> # … with 2 more variables: new_parent_table <chr>, on_delete <chr>
dm:::dm_disentangle(cycle_dm, "airports") %>% 
  dm_get_all_fks()
#> # A tibble: 8 × 5
#>   child_table child_fk_cols     parent_table parent_key_cols   on_delete
#>   <chr>       <keys>            <chr>        <keys>            <chr>    
#> 1 flights-1   origin            airports     faa               no_action
#> 2 flights-2   dest              airports     faa               cascade  
#> 3 flights-1   carrier           airlines-1   carrier           no_action
#> 4 flights-1   tailnum           planes-1     tailnum           no_action
#> 5 flights-1   origin, time_hour weather-1    origin, time_hour no_action
#> 6 flights-2   carrier           airlines-2   carrier           no_action
#> 7 flights-2   tailnum           planes-2     tailnum           no_action
#> 8 flights-2   origin, time_hour weather-2    origin, time_hour no_action
```

<sup>Created on 2022-04-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

closes #55


TSchiefer:
@krlmlr 

It would be nice if you could have a look at some point if this is going in the right direction. I wanted to push it early (this time) so there won't be so much lost time changing everything around.
Adding the arguments `naming_template` and `quiet` will be very straightforward, documentation and tests would come at the end.

Once `dm_disentangle()` works satisfactorily, adding the disentangling functionality to `dm_wrap()` will be super easy too (we could provide an argument `resolve_cycles = TRUE/FALSE` and then just use `dm_disentangle()` on the dm)

krlmlr:
Thanks. I like `enumerate_all_paths()`, but it's too early to discuss `dm_disentangle()`.

How does the mapping between new and original table work? How can we implement the flexible naming scheme from the previous pull request?

TSchiefer:
> How does the mapping between new and original table work?

Basically like you suggested in https://github.com/cynkra/dm/pull/862#issuecomment-1095260560. I just included a few more columns in `out` than `parent_table` and `idx` (the latter I don't use anymore): in addition I also use `node = parent_table, node_key_cols = parent_cols, former_node = child_table, former_key_cols = child_cols` and I pass on all of these to the next step. That way we have all the info we require. `path` is still used to determine which nodes have been visited already.

> How can we implement the flexible naming scheme from the previous pull request?

It was a bit premature of me to say that implementing the naming scheme would be straightforward: there is the problem that both child and parent tables can be duplicated, and giving one fixed schema that works for both of these is not a trivial task (apart from the schema that's currently implemented).
I think an approach might be to provide 2 schemas: 1 for child tables that are replicated and 1 for parent tables, both defaulting to the current implementation.

In any case all the info for flexible naming is available, even when only looking at the result of `enumerate_all_paths()`

krlmlr:
Thanks, I see now. Maybe `enumerate_all_paths()` should return a dm with two data frames, one contains the table map (which is now stored redundantly in two columns), the other contains the new definition of foreign keys (current return value minus three columns)? We can use `dm_get_tables()` for the snapshot tests.

Once we have the data in this format, we can start looking at `dm_disentangle()`. I think the implementation can be very simple:

- Check we're not zoomed
- Get `dm_def()`
- Subset table rows according to the table map, update table names
- Update fk col with information from the new key definition

We shouldn't need to reuse any existing foreign keys. I might take this on if in doubt.

TSchiefer:
These are very good points, mainly the alternative implementation for `dm_disentangle()`, thank you.

In my new implementation `enumerate_all_paths()` outputs a named list though, and not a dm, since the additional functionalities of dm objects aren't necessary (at least the way I implemented it).

To make it easier to update the new FKs, I added a new function `fk_table_to_def_fks()` which accumulates the output like from `dm_get_all_fks()` for each parent table and provides the result in a column `fks`, just as it would appear in the `def` table of a `dm`.
This function might come in handy in other places too, when dealing with a tibble of FKs that should be added to a `dm`.

I realized that the way I did it at first, FKs in unconnected components to the component we're disentangling are not contained in the output of `enumerate_all_paths`. Therefore I readd them now explicitly at the end of the function.

If you feel that what you had in mind is more straightforward, please tweak.

Haven't touched the naming issue yet. Still thinking the "2-template-approach" could work.

R/enumerate_all_paths.R
```diff
@@ -0,0 +1,160 @@
...
+        map_lgl(child_cols, ~ identical(sort(.x), sort(node_key_cols))) &
+        parent_table == former_node &
+        map_lgl(parent_cols, ~ identical(sort(.x), sort(former_key_cols)))
+    ) |
+      (
+        parent_table == node &
+          map_lgl(parent_cols, ~ identical(sort(.x), sort(node_key_cols))) &
+          child_table == former_node &
+          map_lgl(child_cols, ~ identical(sort(.x), sort(former_key_cols)))
+      )
```
krlmlr, line 127: Can we simplify this (and similarly below)? In `graph_df_ud`, each original edge in the data model graph should correspond to two rows, one for each direction.

R/foreign-keys.R
```diff
@@ -568,6 +568,21 @@ check_fk <- function(t1, t1_name, colname, t2, t2_name, pk) {
   )
 }
 
+fk_table_to_def_fks <- function(table,
+                                child_table = "child_table",
+                                child_fk_cols = "child_fk_cols",
+                                parent_table = "parent_table",
+                                parent_key_cols = "parent_key_cols") {
+  group_by(table, !!ensym(parent_table)) %>%
```
krlmlr, line 576: When using the pipe, the LHS should be naked and not a function call. It's easier to read and also makes it easier to insert something between `table` and `group_by()` .

```suggestion
  table %>%
    group_by(!!ensym(parent_table)) %>%
```

R/enumerate_all_paths.R
```diff
@@ -0,0 +1,160 @@
...
+    graph_df_ud %>%
+    filter(child_table == !!node) %>%
+    filter(!(parent_table %in% path)) %>%
+    rename(node = parent_table, node_key_cols = parent_cols, former_node = child_table, former_key_cols = child_cols) %>%
+    select(-on_delete)
+
+  if (nrow(out) == 0) {
+    return()
+  }
+
```
krlmlr, line 77: Do we need this corner case?

```suggestion
```

R/enumerate_all_paths.R
```diff
@@ -0,0 +1,160 @@
...
+rename_unique <- function(all_paths) {
+  all_names <- bind_rows(
+    select(all_paths, table = child_table, new_table = new_child_table),
+    select(all_paths, table = parent_table, new_table = new_parent_table)
+  ) %>%
+    distinct() %>%
+    arrange(table, new_table) %>%
+    group_by(table) %>%
+    count()
```
krlmlr, line 89: This is shorter and will also result in an ungrouped data frame:

```suggestion
    count(table)
```

R/enumerate_all_paths.R
```diff
@@ -0,0 +1,160 @@
...
+  left_join(all_paths, rename(all_names, n_c = n), by = c("child_table" = "table")) %>%
+    left_join(rename(all_names, n_p = n), by = c("parent_table" = "table")) %>%
+    mutate(
+      new_child_table = if_else(n_c == 1, child_table, new_child_table),
```
krlmlr, line 93: A simple lookup in a named vector (which needs to be prepared in a suitable way) might be easier to digest:

```suggestion
      new_child_table = (!!tables)[new_child_table],
```

krlmlr (commented):
Thanks, this looks good! Can we do one more iteration, I'd play with the code after that.

R/enumerate_all_paths.R
TSchiefer: Would be great if we can make it simpler. I am still wondering though:
Here I am searching for the connection that's really existing and subsequently add it to the result table. 
Since we arrived here via an edge in the undirected graph, we do not know if node is the child table or the parent table. I figured I need to filter for both possibilities. In the code below I then take the first hit, since that is the correct one (since we earlier did `bind_rows(correct_graph_tbl, inverse_graph_tbl)`)

It is not enough to store the sequence of tables in the result, since in cases like e.g. `dm_nycflights13()` we then do not know which FK we took from `flights` to `airports`.
It is also not enough to just filter:
```r
filter(
	child_table %in% c(node, former_node),
	map_lgl(child_cols, ~ identical(sort(.x), sort(node_key_cols)) | identical(sort(.x), sort(former_key_cols))),
	parent_table %in% c(node, former_node),
	map_lgl(parent_cols, ~ identical(sort(.x), sort(node_key_cols)) | identical(sort(.x), sort(former_key_cols)))
	)
```
since then we could run into problems with ambiguities.

Maybe I am missing something.

R/enumerate_all_paths.R
TSchiefer: It seems to me you have a different approach in mind than how I implemented it.
In this function (`rename_unique(all_paths)`) I am just doing a little cleanup of the result table, such that tables that were visited only once retain their original name.
At this stage the result table itself already contains all old names, new names (a few of which might be reversed in this function), column names, on_delete.

I feel it might be best if you really work with the code and choose what you deem the best solutions. It's completely fine with me if you choose a very different implementation eventually. I assume you will at least be able to make use of the existing tests for both `enumerate_all_paths()` and `dm_disentangle()`.

TSchiefer:
1. FYI: After you looked at it I still pushed something that I had locally: https://github.com/cynkra/dm/pull/923/commits/d728ceda9ab2bbde1a8dbfb47007e8b4695cc521
2. Regarding naming: I really hadn't properly thought through the naming template question: the problem is that contrary to the first version of `dm_disentangle()` from #862, we have here in general more than 1 table that's connected to a newly created table, which means if we e.g. include all child tables in the name of a parent table the names could get very unwieldy.
My newest idea was to again just use 1 naming template, and provide as building blocks: "former_node", "former_node_key_cols", "node", "node_key_cols", "number". That way the naming is dependent on the way the dm was explored, starting from the `start` node. TBD

R/enumerate_all_paths.R
krlmlr: This might be easier if the foreign keys were named. That shouldn't become a blocker, though.

I'll take a look.

R/enumerate_all_paths.R
krlmlr: I think I understand the intent of the code. I'll try to simplify (and `(!!tables)[...]` isn't exactly simple or intuitive...).

krlmlr:
I changed the code to traverse the directed graph instead, seems to be simpler. Also, each edge now gets a unique numeric ID by which it can be identified much easier than by table names.

Snapshots have changed at some point, because the graph is now traversed in a different order. All commits should make sense and pass tests individually. Does the code still produce correct results?

The new code now uses lookups instead of joins in two places, for the special case of character -> character mapping this is a bit more compact.

Let's postpone flexible naming. Are we happy with the default naming scheme, in particular with the non-syntactic separator?

krlmlr:
`dm_disentangle()` seems to produce invalid objects, https://github.com/cynkra/dm/runs/6103498871?check_suite_focus=true#step:11:80.

R/enumerate_all_paths.R
```diff
@@ -0,0 +1,133 @@
...
+  )
+}
+
+split_to_list <- function(all_paths) {
+  table_mapping <- bind_rows(
+    select(all_paths, new_table = new_child_table, table = child_table),
+    select(all_paths, new_table = new_parent_table, table = parent_table)
+  ) %>%
+    filter(new_table != table) %>%
+    distinct() %>%
```
TSchiefer, line 121:
```suggestion
    distinct() %>%
    mutate(new_table = unname(new_table)) %>%    
```
> dm_disentangle() seems to produce invalid objects

this is because the new column `table` in `def` is named.

krlmlr (approved):
Let's merge it and export + document + rethink in a forthcoming iteration.

krlmlr:
Thanks!

PR #975, opened 2022-04-21
feat: New `json_pack.tbl_lazy()` and `json_nest.tbl_lazy()`

### LINKED ISSUE #969: `json_pack.tbl_lazy()`, `json_nest.tbl_lazy()`
Related to #913 and #918 

@krlmlr I'm happy to take it but if I do I'd appreciate a bit of guidance 

* Should this live in {dm} or {dbplyr} ? (or be moved later)
* Do we want to support every dbplyr supported DBMS or only those who support json ?
* Is there an existing function that would work better as a base to build on ? AFAICT we need a 'tbl_lazy' method but also `op_json_pack()` etc on the lazy side, and  `sql_render.json_pack_query()` and more on the collect side

moodymudskipper:
Scope: postgres + mssql

Closes #969


R/json_pack.R
```diff
@@ -34,3 +34,70 @@ to_packed_json <- function(x) {
...
+        select_subquery <- paste("SELECT", toString(paste(
+          paste0('"', cols_to_pack, '"'), " AS ", paste0('"', cols_to_pack_new, '"')
+        )))
+      }
+      to_json_subquery <- sprintf("TO_JSON((SELECT d FROM (%s) d))", select_subquery)
+      packed_data <-
+        packed_data %>%
+        mutate(!!dot_nms[[i]] := sql(to_json_subquery)) %>%
+        # don't remove cols that have or will be overwritten
+        select(-!!setdiff(cols_to_pack, dot_nms))
```
krlmlr, line 81: `all_of()` or perhaps `any_of()` ?

krlmlr (commented):
Thanks. I wonder if we can do without a `for` loop?

R/json_nest.R
```diff
@@ -32,3 +32,23 @@ json_nest.data.frame <- function(.data, ..., .names_sep = NULL) {
...
+    expr = expr(c(!!!unname(dots))),
+    data = tidyselect_env
+  ) %>%
+    names()
+  group_cols <- setdiff(tidyselect_env, all_cols_to_nest)
+
+  packed_data <- json_pack_tbl_lazy_impl(.data, dots, tidyselect_env, group_cols, .names_sep)
+
+  nested_data <- packed_data %>%
+    group_by(across(!!group_cols)) %>%
```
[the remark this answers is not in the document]
moodymudskipper: https://github.com/cynkra/dm/pull/975/commits/79e6281aca9baf3189ae45f5a2e785ccbb2d02ba

R/json_pack.R
moodymudskipper: https://github.com/cynkra/dm/pull/975/commits/79e6281aca9baf3189ae45f5a2e785ccbb2d02ba

R/json_pack.R
```diff
@@ -34,3 +34,70 @@ to_packed_json <- function(x) {
...
+  if (is_mssql(con)) {
+    dbms <- "mssql"
+  } else if (is_postgres(con)) {
+    dbms <- "postgres"
+  } else {
+    abort("Unsupported DBMS")
+  }
+  # FIXME: move dot checking in generic (also for df methods), that's what `nest` does
+  dot_nms <- names(dots)
+  if (is_null(dot_nms) || "" %in% dot_nms) {
```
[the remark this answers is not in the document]
moodymudskipper: https://github.com/cynkra/dm/pull/975/commits/25447f0fecf40f6b069dd0e2cf1f17945aac62c6

moodymudskipper:
* Should we add a "pq_json" class to our data frame methods for consistency ?
* Should we apply a transformation to the output of our data frame methods to add spaces around `,` and `:` to enforce consistency with remote methods ?

R/json_nest.R
```diff
@@ -24,11 +24,43 @@ json_nest <- function(.data, ..., .names_sep = NULL) {
...
+    group_by(across(all_of(id_cols))) %>%
+    summarize(!!!sql_exprs) %>%
+    ungroup()
+}
+
+sql_json_nest <- function(con, cols, names_sep, packed_col) {
+  UseMethod("sql_json_nest")
+}
+
+sql_json_nest.PqConnection <- function(con, cols, names_sep, packed_col) {
```
krlmlr, line 59: Here and below?

```suggestion
#' @export
sql_json_nest.PqConnection <- function(con, cols, names_sep, packed_col) {
```

moodymudskipper:
> * What's the purpose of a special `pq_json` class?

just to make the collected remote data frame identical to the locally computed version, but probably not necessary considering your next comment

> * It's more than sufficient if the JSON data is equivalent for databases and data frames, it doesn't have to be identical

I agree that them being functionally equivalent is the most important, but for tests it's nice to be able to check equivalence. Should we have a transformation function just for tests ? could take care of both the class and the spaces.

For the spaces it seems that this works ok, at least enough for an unexpected function used for tests only : 

```
library(magrittr)
"{\"Sepal.Length\":5.1,\"Sepal.Width\":3.5}" %>% 
  gsub(":(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", " : ", ., perl = TRUE) %>% 
  gsub(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", ", ", ., perl = TRUE)
#> [1] "{\"Sepal.Length\" : 5.1, \"Sepal.Width\" : 3.5}"
```

R/json_pack.R
```diff
@@ -34,3 +34,70 @@ to_packed_json <- function(x) {
   jsonlite::stream_out(x, con, digits = NA)
   textConnectionValue(con)
 }
+
+json_pack_tbl_lazy_impl <- function(.data, dots, tidyselect_env, group_cols, .names_sep) {
+  con <- dbplyr::remote_con(.data)
+  if (is_mssql(con)) {
```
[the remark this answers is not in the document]
moodymudskipper: https://github.com/cynkra/dm/pull/975/commits/2f9f557c108c7d3c5c3119ed877736937ebc712f

tests/testthat/test-json_nest.R
```diff
@@ -5,3 +5,21 @@ test_that("`json_nest()` works", {
...
+  # FIXME: add "mssql" when ready
+  skip_if_src_not("postgres")
+  con <- my_test_src()$con
+
+  withr::defer(
+    try(dbExecute(con, "DROP TABLE iris"))
+  )
+
+  dbWriteTable(con, "iris", iris)
+  iris_remote <- tbl(con, "iris")
```
krlmlr, line 19: We have `test_db_src_frame()` for that. I wouldn't mind using a simpler test data frame here, and also avoid iris for new code.

tests/testthat/test-json_nest.R
```diff
@@ -5,3 +5,21 @@ test_that("`json_nest()` works", {
...
+
+  expect_snapshot({
```
krlmlr, line 21: The snapshot test is likely to fail for other databases. We can make it exclusive to Postgres, or try to test an invariant.

moodymudskipper:
> Can we specify an invariant: converting the JSON to a tibble yields identical results (after `collect()`)?

At the moment a locally json packed column has no class (regular character columns) so we cannot robustly do this, what about the following ? We should also take care about factors, maybe more if we .

```
local_db <- test_src_postgres()
# DBI::dbWriteTable(local_db$con, "iris", iris)
iris_remote <- tbl(local_db, "iris")
local <- json_pack(as_tibble(iris), Sepal = 1:2, Petal = starts_with("Petal"), .names_sep = ".")
remote <- json_pack(iris_remote, Sepal = 1:2, Petal = starts_with("Petal"), .names_sep = ".")

equivalent <- function(local, remote) {
  remote_collected <- collect(remote)
  remote_processed <- map_dfc(remote_collected, ~ {
    if (!inherits(., "pq_json")) return(.)
    gsub(" ?([:,]) (?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", "\\1", unclass(.), perl = TRUE)
  })
  local_no_factor <- map_dfc(local, ~ if (is.factor(.)) as.character(.) else .)
  identical(local_no_factor, remote_processed)
}
equivalent(local, remote)
#> [1] TRUE
```

In hindsight, pq_json columns also print nicer in tibble than character columns so there's a case to use the class for data.frame methods.

R/json_nest.R
```diff
@@ -24,11 +24,44 @@ json_nest <- function(.data, ..., .names_sep = NULL) {
...
+  UseMethod("sql_json_nest")
+}
+
+#' @export
+sql_json_nest.PqConnection <- function(con, cols, names_sep, packed_col) {
+  inside_cols <- remove_prefix_and_sep(cols, prefix = packed_col, sep = names_sep)
+  inside_cols_idented <- dbplyr::ident(inside_cols)
+  n <- length(inside_cols)
+  # alternate names and expressions for `json_build_object`
+  exprs <- c(syms(cols), inside_cols_idented)[rep(1:n, each = 2) + c(n, 0)]
```
krlmlr, line 65: `1:n` is prone to corner cases, `seq_len(n)` is always safer.

I used `as.vector(t(matrix(..., ncol = 2)))` in similar cases, not sure which is better. With `!!!` we can ditch the `as.vector()`. Python has `zip()`, maybe we can add it as a helper function here?

moodymudskipper: This is what I use normally, but in our case it's a bit more complex since we have a vector of characters and a list of symbols. We can adapt it though but it's less efficient.

``` r
x <- letters[1:3]
y <- LETTERS[1:3]
z <- rlang::syms(y)
as.vector(t(matrix(c(x,y), ncol = 2)))
#> [1] "a" "A" "b" "B" "c" "C"
as.vector(t(matrix(c(x,z), ncol = 2)))
#>      [,1] [,2] [,3]
#> [1,] "a"  "b"  "c" 
#> [2,] ?    ?    ?

unlist(t(matrix(c(x, purrr::map(z, list)), ncol = 2)))
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> A
#> 
#> [[3]]
#> [1] "b"
#> 
#> [[4]]
#> B
#> 
#> [[5]]
#> [1] "c"
#> 
#> [[6]]
#> C
n <- length(x)
c(z, x)[rep(seq_len(n), each = 2) + c(n, 0)]
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> A
#> 
#> [[3]]
#> [1] "b"
#> 
#> [[4]]
#> B
#> 
#> [[5]]
#> [1] "c"
#> 
#> [[6]]
#> C

X <- rep(x, 1e4)
Z <- rep(z, 1e4)
bench::mark(
  rep = c(Z, X)[rep(seq_len(n <- length(X)), each = 2) + c(n, 0)],
  matrix = unlist(t(matrix(c(X, purrr::map(Z, list)), ncol = 2)))
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 rep          1.21ms   1.33ms     739.     1.95MB     57.8
#> 2 matrix      10.22ms  10.28ms      96.6    2.06MB    869.
```

<sup>Created on 2022-04-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Good idea to have a helper function
* `alternate <- function(...)` with `...` vectors/lists of same length ?
*  if we want it type stable in the `purrr` style we need `alternate()`, `alternate_chr()` etc
*  In this PR or a new one ?

moodymudskipper:
> Can we test, for our test cases, that `json_nest() %>% collect() %>% unjson_nested()` gives results equivalent to `collect() %>% json_nest() %>% unjson_nested()` and `collect() %>% nest()` ? Here, `unjson_nested()` would be a (yet to be implemented?) operation to convert JSON columns to nested or packed columns -- later with a database implementation too on Postgres and dbplyr.

OK, in this PR ?

R/json_nest.R
krlmlr: Let's keep this code as is, I opened a new issue for `alternate()` .

krlmlr:
In this PR we can do a very specialized version of `unjson_nested()` -- on data frames, perhaps also only for a specific test case. Let's think about generalizing later.

R/json_nest.R
moodymudskipper: https://github.com/cynkra/dm/pull/975/commits/4592c36123f95504b79e8951328e65165ee771dd and https://github.com/cynkra/dm/issues/982

tests/testthat/test-json_nest.R
moodymudskipper: https://github.com/cynkra/dm/pull/975/commits/7c030ebc5cb1e910ddc3f2ca234713888fa0cab4

tests/testthat/test-json_nest.R
moodymudskipper: It was exclusive to Postgres due to the `skip_if_src_not("postgres")` but we now also use the `variant` argument so we can have multiple snapshots when we support other databases. https://github.com/cynkra/dm/pull/975/commits/7c030ebc5cb1e910ddc3f2ca234713888fa0cab4

R/json_nest.R
mgirlich: Sorry to interrupt but I just stumbled over your discussion about `alternate()`. I think that `vctrs::vec_interleave()` is what you are looking for

``` r
x <- letters[1:3]
y <- LETTERS[1:3]
vctrs::vec_interleave(x,y)
#> [1] "a" "A" "b" "B" "c" "C"

z <- rlang::syms(y)
vctrs::vec_interleave(as.list(x), z, .ptype = list())
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> A
#> 
#> [[3]]
#> [1] "b"
#> 
#> [[4]]
#> B
#> 
#> [[5]]
#> [1] "c"
#> 
#> [[6]]
#> C
```

<sup>Created on 2022-04-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

moodymudskipper:
In the end I believe I need `OPENJSON` because your energie example was nesting, not merging.

In your example you had some names and aliases hardoded, since I could not do that here the SQL editing is a bit convoluted, especially for the nest part. I'm happy to consider better ways of doing this.

When we're happy with the general implementation these should be tested and tweaked with non syntactic columns/table names, and with no grouping column.

moodymudskipper:
Testing from my machine on SQL Server it works fine : 

``` r
library(dplyr, warn = F)
library(dm)
#> 
#> Attaching package: 'dm'
#> 
#>     filter
con_mssql <- DBI::dbConnect(
  odbc::odbc(),
  server = Sys.getenv("TMP_DB_MSSQL_HOST"),
  uid = Sys.getenv("TMP_DB_MSSQL_USER"),
  pwd = Sys.getenv("TMP_DB_MSSQL_PASSWORD"),
  port = Sys.getenv("TMP_DB_MSSQL_PORT"),
  driver = "ODBC Driver 17 for SQL Server"
)
DBI::dbExecute(con_mssql, "USE test;", immediate = TRUE)
#> [1] 0
remote <- tbl(con_mssql, "remote")
remote
#> # Source:   table<remote> [?? x 3]
#> # Database: Microsoft SQL Server 15.00.4073[mssqldb@EC2AMAZ-P2L9J21/test]
#>     grp a_i   a_j  
#>   <dbl> <chr> <chr>
#> 1     1 a     A    
#> 2     1 b     B    
#> 3     2 c     C    
#> 4     2 d     D
remote %>% json_nest(A = starts_with("a")) %>% arrange(grp) %>% collect()
#> # A tibble: 2 × 2
#>     grp A                                                              
#>   <dbl> <chr>                                                          
#> 1     1 "[{\"a_i\":\"a\",\"a_j\":\"A\"},{\"a_i\":\"b\",\"a_j\":\"B\"}]"
#> 2     2 "[{\"a_i\":\"c\",\"a_j\":\"C\"},{\"a_i\":\"d\",\"a_j\":\"D\"}]"
```

<sup>Created on 2022-05-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

However in CI it seems that every row is nested for every group. This seems to be my substitution code going wrong.

krlmlr:
Can you SSH into the worker to understand what's happening?

https://github.com/mxschmitt/action-tmate

https://tmuxcheatsheet.com/, section "Windows"

krlmlr:
On CI:

```
> remote %>% json_nest(A = starts_with("a")) %>% show_query()
<SQL>
SELECT "grp", (SELECT "a_i", "a_j"
FROM "#test_frame_2022_05_09_12_26_58_17832_1"
WHERE ("grp" = "#test_frame_2022_05_09_12_26_58_17832_1"."grp") FOR JSON PATH) AS "A"
FROM "#test_frame_2022_05_09_12_26_58_17832_1"
GROUP BY "grp"
```

moodymudskipper:
locally I have : 

```
<SQL> SELECT
  "grp",
  (SELECT "a_i", "a_j"
FROM "remote"
WHERE ("grp" = "q01"."grp") FOR JSON PATH) AS "A"
FROM (
  SELECT "grp", "a_i", "a_j"
  FROM "remote"
) "q01"
GROUP BY "grp"
```

moodymudskipper:
@krlmlr I think we could make the dbplyr code look much better if we allow the SQL code to look a little less good, would the following query be acceptable for you for nesting ? Here we don't use the joining trick from above, instead we extract without grouping, creating json objects rowdies, then concatenate by group with `STRING_AGG` and build arrays.

With this solution I believe I just need to edit a subquery replacing "FROM.*" with "FOR JSON PATH, WITHOUT_ARRAY_WRAPPER", and the rest would be idiomatic dbplyr code.


```
SELECT "grp", "A" = \'[\' + STRING_AGG("A", \',\') + \']\' FROM
(SELECT "grp", (SELECT "a_i", "a_j" FOR JSON PATH, WITHOUT_ARRAY_WRAPPER) A
FROM "remote"
GROUP BY "grp", "a_i", "a_j") B GROUP BY "grp"
```

krlmlr:
Thanks. I like the other query better.

In the query rendered with CRAN dbplyr, it seems that we have a problem with table aliases. A correct query would be something like (note the `"nested_001"`): 

```
> remote %>% json_nest(A = starts_with("a")) %>% show_query()
<SQL>
SELECT "grp", (SELECT "a_i", "a_j"
FROM "#test_frame_2022_05_09_12_26_58_17832_1" "nested_001"
WHERE ("nested_001"."grp" = "#test_frame_2022_05_09_12_26_58_17832_1"."grp") FOR JSON PATH) AS "A"
FROM "#test_frame_2022_05_09_12_26_58_17832_1"
GROUP BY "grp"
```

moodymudskipper:
Do we need code that works for both current cran and new {dbplyr} ?
I could tweak the code to make it work with more string manipulation but I don't know how to guarantee that it is  robust to further translation modifications, hence the above alternative proposal.

krlmlr:
Let's target dbplyr >= 2.2.0 .

krlmlr:
dbplyr 2.2.0 is on CRAN now, dm 0.3.0 will follow suit very soon.

R/json_nest.R
```diff
@@ -24,11 +24,127 @@ json_nest <- function(.data, ..., .names_sep = NULL) {
...
+  # build final query
+  query <- glue::glue(
+    'SELECT {toString(id_cols)}, {toString(nest_col_queries)} ',
+    'FROM (SELECT {toString(col_nms)} FROM {tbl_name}) {temp_alias} GROUP BY {id_cols}'
+  )
+
+  # HACK! see FIXME at top
+  query <- gsub("'", '"', query)
+
+  tbl(con, sql(query))
```
krlmlr, line 149:
```suggestion
  tbl(con, sql(query), vars = ...)
```

to avoid an extra query to the database.

krlmlr (commented):
If it works, looks good!

moodymudskipper:
* `json_nest.tbl_Microsoft SQL Server` now uses `sql_render()` rather than assuming it's passed an existing table
* It handles .names_Sep
* The code is restructured so subqueries and different step of building those are stored in a tibble called "nesting plan"
* Tests are fixed so they make sense with `.names_sep`

* `json_pack.tbl_Microsoft SQL Server` is implemented and structured overall like `json_nest.tbl_Microsoft SQL Server`, just a bit simpler

moodymudskipper:
@krlmlr Ready for review

krlmlr:
Tests are brittle, see https://github.com/cynkra/dm/pull/1445/files and https://github.com/cynkra/dm/pull/1444/files . How can we stabilize?

R/json_nest.R
```diff
@@ -24,11 +24,120 @@ json_nest <- function(.data, ..., .names_sep = NULL) {
 json_nest.data.frame <- function(.data, ..., .names_sep = NULL) {
   check_suggested("jsonlite", use = TRUE, top_level_fun = "json_nest")
   dot_nms <- ...names()
-  # `{tidyr}` only warns but since we don't need backward compatibility we're
-  #   better off failing
-  if (is_null(dot_nms) || "" %in% dot_nms) {
-    abort("All elements of `...` must be named.")
-  }
```
krlmlr, line 31: Why is this gone now?

R/json_pack.R
```diff
@@ -31,6 +31,117 @@ json_pack.data.frame <- function(.data, ..., .names_sep = NULL) {
...
+    names_sep = .names_sep,
+    packed_col = .y,
+    data = .data
+  ))
+
+  .data %>%
+    transmute(!!!syms(id_cols), !!!sql_exprs)
+}
+
+sql_json_pack <- function(con, cols, names_sep, packed_col, data) {
```
krlmlr, line 61: All methods need an ellipsis, otherwise it's difficult to extend once they are exported. (We don't need it right now, but it's so easy to forget later.)

krlmlr (commented):
Thanks, looks good once we have dealt with the technicalities. We'll see how this works in practice when we can use the new methods as a backend to `dm_wrap_tbl()` .

R/json_nest.R
moodymudskipper: It happened in 2f9f557c108c7d3c5c3119ed877736937ebc712f when refactoring, it might have been a mistake, I'll put it back

moodymudskipper:
> Tests are brittle, see https://github.com/cynkra/dm/pull/1445/files and https://github.com/cynkra/dm/pull/1444/files . How can we stabilize?

On Postgres it seems we have time dependent aliases, I tried  `gsub("test_frame_[_0-9]+", "test_frame_...", query)` to get a simplified output

krlmlr:
Thanks! Please merge when the tests are green.

I have squash-merged #1444 and #1445, please watch out for those autogenerated PRs too.

moodymudskipper:
@krlmlr It's all green except for codecov/patch, should I add more tests ?

krlmlr:
Thanks!

krlmlr:
It's weird that the patch coverage is so low, the database tests should have covered the new code and also registered it with codecov.

PR #988, opened 2022-04-29
Compatibility with dplyr 1.0.9

krlmlr:
and dbplyr update.

This is a breaking change:

- Adding required `conflict` or `unmatched` arguments
- Removing unimplemented `check` argument
- Removed message when `in_place` is missing
- Unexported `sql_rows_*()` methods
- Ready for dbplyr > 2.1.1, using dbplyr methods if provided, without breaking tests
- Added `Matching, by = ...` message when `by` is omitted (with dbplyr > 2.1.1)
- New `rows_append()` and `dm_rows_append()`
- Fix detection of availability of primary keys in `dm_rows_*()` operations


R/rows-db.R
```diff
@@ -62,9 +57,13 @@ NULL
...
 #' @rdname rows-db
 rows_insert.tbl_dbi <- function(x, y, by = NULL, ...,
-                                in_place = NULL, copy = FALSE, check = NULL,
+                                in_place = FALSE,
+                                conflict = NULL,
+                                copy = FALSE,
                                 returning = NULL) {
 
+  stopifnot(identical(conflict, "ignore"))
+
```
moodymudskipper, line 70: Should our doc explain that (and why) only "ignore" is supported for `conflict` and `unmatched` ?

krlmlr: This is only compatibility code that will be removed as soon as dbplyr 2.2.0 hits CRAN. No need to spend time here, I guess.

R/rows-db.R
```diff
@@ -292,11 +305,6 @@ rows_upsert.tbl_dbi <- function(x, y, by = NULL, ...,
   name <- target_table_name(x, in_place)
 
   if (!is_null(name)) {
-    # Checking optional, can rely on primary key constraint
-    if (is_true(check)) {
-      check_db_superset(x, y, by)
-    }
-
     if (is_empty(new_columns)) {
       return(invisible(x))
```
moodymudskipper, line 304: Are we sure we want to return sometimes visibly sometimes invisibly ?

R/aaa-dbplyr.R
```diff
@@ -1,3 +1,16 @@
+replace_if_dbplyr_has <- function(fun) {
+  dbplyr_ns <- asNamespace("dbplyr")
+
+  fun <- as_string(ensym(fun))
+  value <- mget(fun, dbplyr_ns, mode = "function", ifnotfound = list(NULL))[[1]]
+  if (!is.null(value)) {
+    assign(fun, value, inherits = TRUE)
+    "dbplyr"
+  } else {
+    "dm"
```
moodymudskipper, line 14: should side effect functions return invisibly ?

R/aaa-dbplyr.R
```diff
@@ -1,4 +1,8 @@
 replace_if_dbplyr_has <- function(fun) {
+  if (!requireNamespace("dbplyr", quietly = TRUE)) {
+    return()
```
moodymudskipper, line 3: `return(invisible())` ?

R/aaa-dbplyr.R
```diff
@@ -12,15 +16,18 @@ replace_if_dbplyr_has <- function(fun) {
...
 
-  if (is_empty(methods)) {
-    return()
+    # Register our method implementations only if dbplyr doesn't provide them
+    dbplyr_methods <- mget(names(methods), dbplyr_ns, mode = "function", ifnotfound = list(NULL))
+
+    methods <- methods[map_lgl(dbplyr_methods, is.null)]
+
+    if (is_empty(methods)) {
+      return()
```
moodymudskipper, line 29: invisible ?

R/rows-db.R
krlmlr: For `in_place = TRUE` we do return invisibly, because this is an operation with side effects.

R/aaa-dbplyr.R
krlmlr: You're right, but it's not too important here and below, because this is an internal function called in `.onLoad()` .

krlmlr:
@moodymudskipper: Please approve when it's "good enough".

moodymudskipper (approved):

PR #997, opened 2022-05-11
feat: New `json_unnest()` and `json_unpack()`, currently implemented for data frames only

### LINKED ISSUE #991: json_unnest(), json_unpack()
after #969 / #975 , for #913

data.frame and tbl_lazy methods, Postgres and mssql

On data.frames we might consider formalising a 2 step process : 
* json packed -> packed -> unpacked
* json nested -> nested -> unnested

At the end of the first step our nested/packed column might be a nested lists OR a list of tibbles.
At the moment we have unexported drafts of `unjson_nested()` and `unjson_packed()` which do the latter thanks to `jsonlite::fromJSON(simplifyDataFrame = TRUE)`'s magic. But we could use {tibblify} too.

On remote databases we wouldn't have this intermediate step

moodymudskipper:
Closes #991

For now we have data frame methods only, and roundtrip tests.

For remote data bases the challenge is to map the data types dynamically. The following works but I wouldn't know how to generate the `WITH` clause dynamically in an efficient way.

```
CREATE TABLE tbl1_have (id INT, json_col VARCHAR(100))
INSERT INTO tbl1_have VALUES 
(1,   '{"a_i":"a","a_j":1}'),
(1,   '{"a_i":"b","a_j":2}'),
(2,   '{"a_i":"c","a_j":3}'),
(2,   '{"a_i":"d","a_j":4}')

CREATE TABLE tbl2_have (id INT, json_col VARCHAR(100))
INSERT INTO tbl2_have VALUES 
(1,   '[{"a_i":"a","a_j":1},{"a_i":"b","a_j":2}]'),
(2,   '[{"a_i":"c","a_j":3},{"a_i":"d","a_j":4}]')

SELECT id, a_i, a_j FROM tbl1_have CROSS APPLY OPENJSON(json_col) 
WITH(a_i VARCHAR(100), a_j INT)

id  a_i a_j                             
1   a   1
1   b   2
2   c   3
2   d   4

SELECT id, a_i, a_j FROM tbl2_have CROSS APPLY OPENJSON(json_col) 
WITH(a_i VARCHAR(100), a_j INT)

id  a_i a_j                             
1   a   1
1   b   2
2   c   3
2   d   4
```

I've asked on [SO](https://stackoverflow.com/questions/72102467) and got a nice answer but there is a small bit of string concatenation to make json objects stored outside of arrays into length one arrays. 

```
select string_agg(quotename(k) + case t
    when 0 then ' nchar(1)'       -- javascript null
    when 1 then ' nvarchar(max)'  -- javascript string
    when 2 then ' float'          -- javascript number
    when 3 then ' bit'            -- javascript boolean
    else ' nvarchar(max) as json' -- javascript array or object
end, ', ') within group (order by k)
from (
    select j2.[key], max(j2.[type])
    from test
    cross apply openjson(case when json_col like '{%}' then '[' + json_col + ']' else json_col end) as j1
    cross apply openjson(j1.value) as j2
    group by j2.[key]
) as kt(k, t)
```

May we proceed with this code @krlmlr ?


krlmlr:
Thanks! 

I wonder if the JSON stuff should live in a separate package. Downsides are that we'd need to replicate the database testing setup there as well, but this would be mitigated by moving database methods to dbplyr. No action needed.

Issue #999, opened 2022-05-20
Remove all calls of soft deprecated functionalities


pawelru:
an example:
https://github.com/cynkra/dm/blob/0cd852445b5e41fcc93fc2f428bf02c22d2cc165/vignettes/tech-dm-join.Rmd#L98-L100

results the following:
![image](https://user-images.githubusercontent.com/12943682/169556872-3d39aeed-7560-4b7d-92a8-8b6c9a19979e.png)

Looking into the code - there are more soft-deprecated functions. Would be worth to clean this up.

krlmlr:
Thanks, good catch. Would you like to help out with a PR?

PR #1000, opened 2022-05-20
remove soft deprecated dplyr-src funs

pawelru:
close https://github.com/cynkra/dm/issues/999

I have found many other calls to those functions but I decided to leave them - here goes my explanation:
- `dm_get_src()`
image
  - `R/` - only deprecated function itself
  - `scratch` - I have left your notes untouched
  - `tests/` - as you soft-deprecated then that means that you still expect it to work - would be good to keep that check

- `tbl()`
image
  - I left `demo/` directory as it is as I assume it's for demoes that already happened so this code is unlikely to be executed again. If yes - should be done on the current version at that point in time.
  - `R/` - only comments and deprecated function itself
  - `tests/` - same as above

- `src_tbls()`
image
  - `demo/` - as above
  - I think that the last occurence is intended to call `dplyr::src_tbls` (would be good to prefix though) - please confirm

- `copy_to()`
image
  - `R/` - same as above
  - `tests/` - same as above
  - `vignettes/` - here I probably need your help with assessment. Should it stay or not?

### DIFF
--- a/R/flatten.R
+++ b/R/flatten.R
@@ -31,7 +31,7 @@
 #' Then all involved foreign tables are joined to the `start` table successively, with the join function given in the `join` argument.
 #'
 #' **Case 2**, filter conditions are set for at least one table that is connected to `start`:
-#' First, disambiguation will be performed if necessary. The `start` table is then calculated using `dm[["start"]]`.
+#' First, disambiguation will be performed if necessary. The `start` table is then calculated using `dm[[start]]`.
 #' This implies
 #' that the effect of the filters on this table is taken into account.
 #' For `right_join`, `full_join` and `nest_join`, an error
--- a/R/flatten.R
+++ b/R/flatten.R
@@ -31,7 +31,7 @@
 #' Then all involved foreign tables are joined to the `start` table successively, with the join function given in the `join` argument.
 #'
 #' **Case 2**, filter conditions are set for at least one table that is connected to `start`:
-#' First, disambiguation will be performed if necessary. The `start` table is then calculated using `tbl(dm, "start")`.
+#' First, disambiguation will be performed if necessary. The `start` table is then calculated using `dm[["start"]]`.
 #' This implies
 #' that the effect of the filters on this table is taken into account.
 #' For `right_join`, `full_join` and `nest_join`, an error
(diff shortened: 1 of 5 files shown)


R/flatten.R
```diff
@@ -265,7 +265,7 @@ prepare_dm_for_flatten <- function(dm, tables, gotta_rename) {
   red_dm <-
     dm_reset_all_filters(dm) %>%
     dm_select_tbl(!!!tables)
-  # Only need to compute `tbl(dm, start)`, `dm_apply_filters()` not necessary
+  # Only need to compute `dm[[start]]`, `dm_apply_filters()` not necessary
```
krlmlr, line 268:
```suggestion
  # Only need to compute `dm$start`, `dm_apply_filters()` not necessary
```

vignettes/tech-dm-join.Rmd
```diff
@@ -96,11 +96,11 @@ The difference is the `name` column from the `airlines` table.
 
 ```{r}
 dm %>%
-  tbl("flights") %>%
+  .[["flights"]] %>%
```
krlmlr, line 99: I'd prefer `dm$flights` here, without pipe.

krlmlr (commented):
Thanks a lot! Is there a reason to prefer `[[` over `$` ?

krlmlr:
The demos are part of `R CMD check`, we will catch when things start to fail. But then they are only of historic interest, so if this happens we could prune all except the most recent one.

krlmlr:
For the vignettes we'd want to look for idiomatic replacements. Happy to provide guidance as needed, pinging @TSchiefer and @moodymudskipper too.

vignettes/tech-dm-join.Rmd
pawelru: sure - a commit is coming

pawelru:
> Thanks a lot! Is there a reason to prefer `[[` over `$` ?

Nothing really strong. To me, `[[` deals better with list name contains a `"$"`. You would then have to add some back-ticks which doesn't look good.

```
> x <- list("a$b" = 1, a = 2, b = 3)
> x$a$b
Error in x$a$b : $ operator is invalid for atomic vectors
> x$`a$b`
[1] 1
> x[["a$b"]]
[1] 1
```

But this is rather extremely case. You are the owner so I would do what you prefer.

R/flatten.R
```diff
@@ -31,7 +31,7 @@
 #' Then all involved foreign tables are joined to the `start` table successively, with the join function given in the `join` argument.
 #'
 #' **Case 2**, filter conditions are set for at least one table that is connected to `start`:
-#' First, disambiguation will be performed if necessary. The `start` table is then calculated using `tbl(dm, "start")`.
+#' First, disambiguation will be performed if necessary. The `start` table is then calculated using `dm[["start"]]`.
```
krlmlr, line 34: `start` is an argument to this function:?

```suggestion
#' First, disambiguation will be performed if necessary. The `start` table is then calculated using `dm[[start]]`.
```

krlmlr (approved):
Thanks! Let's get this merged so that future PRs don't require confirmation to run workflows.

krlmlr:
Thanks!

PR #1012, opened 2022-06-02
FIX: avoid NULL if no fks after disentangle

TSchiefer:
NULL appears in `def$fks`, need to avoid, otherwise something like this fails (and in general no valid dm is produced):
```
dm_disentangle(dm_nycflights13(cycle = TRUE), flights) %>% 
  dm_select_tbl(-`airports-1`)
```

### DIFF
--- a/R/disentangle.R
+++ b/R/disentangle.R
@@ -31,5 +31,6 @@ dm_disentangle <- function(dm, start, quiet = FALSE) {
     left_join(fk_table, by = c("table" = "new_parent_table")) %>%
     select(-fks) %>%
     relocate(fks = new_fks, .after = pks) %>%
+    mutate(fks = vctrs::as_list_of(map(fks, ~ if (is.null(.x)) {new_fk()} else {.x}))) %>%
     new_dm3()
 }
--- a/R/disentangle.R
+++ b/R/disentangle.R
@@ -31,10 +31,6 @@ dm_disentangle <- function(dm, start, quiet = FALSE) {
     left_join(fk_table, by = c("table" = "new_parent_table")) %>%
     select(-fks) %>%
     relocate(fks = new_fks, .after = pks) %>%
-    mutate(fks = vctrs::as_list_of(map(fks, ~ if (is.null(.x)) {
-      new_fk()
-    } else {
-      .x
-    }))) %>%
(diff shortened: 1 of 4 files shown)


TSchiefer:
actually `validate_dm()` doesn't even detect the NULL in `def$fks`. The class of the column is correct (`<list<tibble[,4]>>`). Probably needs a dedicated check in `validate_dm()`.

krlmlr (commented):
Thanks. Could this be considered a bug in vctrs?

krlmlr:
Should we fix `validate_dm()` here too.

TSchiefer:
Good point about the vctrs requirements that are not actually met. I'll try to make a reprex for a potential issue there, let's see

krlmlr:
Scratch that, I'm now pretty sure it's the correct vctrs behavior.

TSchiefer:
hm, ok, I'd have this here, but it is true that vctrs couldn't necessarily avoid that:
``` r
suppressPackageStartupMessages({
  library(vctrs)
  library(tibble)
  library(dplyr)
})

a <- tibble(col_a = 1:3)
b <- tibble(col_a = 2:4, col_b = as_list_of(as.list(1:3)))
join_res <- left_join(a, b, by = "col_a")
join_res
#> # A tibble: 3 × 2
#>   col_a       col_b
#>   <int> <list<int>>
#> 1     1            
#> 2     2         [1]
#> 3     3         [1]
join_res$col_b
#> <list_of<integer>[3]>
#> [[1]]
#> NULL
#> 
#> [[2]]
#> [1] 1
#> 
#> [[3]]
#> [1] 2
```

<sup>Created on 2022-06-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

Ok, I'll leave it at that

krlmlr:
`dm_financial()` is not available now. Can you please fix the test so that it is skipped if the DB is not available, here or in a separate PR?

TSchiefer:
Do you think vctrs could give a warning/message that a join might result in NULL entries in columns which are otherwise strictly controlled regarding the class? Actively checking the class might be too expensive(?)

krlmlr:
Thanks!

PR #1035, opened 2022-06-06
API: Add ellipses to `dm_disambiguate_cols()`, `dm_draw()`, `dm_examine_constraints()`, `dm_nycflights13()` and `dm_pixarfilms()`

krlmlr:
for stable API.

### DIFF
--- a/R/disambiguate.R
+++ b/R/disambiguate.R
@@ -1,7 +1,11 @@
 #' Resolve column name ambiguities
 #'
+#' @description
+#' `r lifecycle::badge("stable")`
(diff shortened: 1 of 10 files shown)


krlmlr:
Merging, because I'm basing another PR on this. A second pair of eyes would help nevertheless.

TSchiefer:
Looks good.
I could imagine that mainly the ellipses in dm_nycflights13() and dm_pixarfilms() could break some existing code. We should definitely mention it in NEWS.md

PR #1038, opened 2022-06-06
API: Deprecate dm_join_to_tbl(), dm_is_referenced() and dm_get_referencing_tables()

krlmlr:
.


krlmlr:
I'd like to keep the API surface as small as possible. Let me know if I'm missing a functionality that's gone now.

Issue #1064, opened 2022-06-13
Use cynkratemplate


maelle:

krlmlr:
Happy to, as soon as we find a way to integrate https://github.com/cynkra/cynkraweb-assets.

maelle:
https://github.com/cynkra/cynkratemplate/pull/18

krlmlr:
Done now?

PR #1067, opened 2022-06-13
docs: BS5 for greater readability

maelle:
- bigger font
- search

IMHO #1064 should come after cynkratemplate itself uses BS5 and therefore has a search bar.

image

### DIFF
--- a/pkgdown/_pkgdown.yml
+++ b/pkgdown/_pkgdown.yml
@@ -167,48 +167,42 @@ reference:
   - dm_validate
 
 navbar:
-  title: ~
-  type: default
-  left:
-  - text: Intro
-    href: articles/dm.html
-  - text: Reference
-    href: reference/index.html
-  - text: Tutorials
-    menu:
-    - text: Introduction to relational data models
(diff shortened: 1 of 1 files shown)


maelle:
The navbar links should not be blue, fix would be https://github.com/r-lib/webfakes/commit/7aeb062e92e6d5e05d15cdc8853c0032e23af50a, will do in a sec

maelle:
I've been happily reading the docs in a local build with this config.

krlmlr:
Thanks. https://github.com/cynkra/dm/actions/runs/2488843602 will put it into https://cynkra.github.io/dm/dev/ in a few minutes (because the branch name is prefixed with `docs-`), I'll take a look.

krlmlr:
Thanks, I like it! Where does the logo come from?

image

How can we make the logos (cynkra, energie360°, ...) look good in dark mode?

maelle:
@krlmlr the logo comes from https://github.com/cynkra/dm/blob/main/man/figures/logo.png which is a location automatically recognized by pkgdown. :smile_cat:

PR #1075, opened 2022-06-13
docs: Minimal tweaks of intro vignette

maelle:


krlmlr (approved):
Thanks, looks great! Let's wait for the BS5 run before merging.

Issue #366, opened 2020-05-17
dm_learn_from_db and MySQL


ericemc3:
looking forward to playing with this within MySQL...

krlmlr:
I wonder if `postgres_learn_query()` works well enough for MariaDB? I never tried, would review a pull request.

ericemc3:
It won't probably work as if, because you won't find in MySQL the same exact column names.
We are (in my company) implementing the same db model with various databases (such as MySQL, Postgres or MsSQL).

Useful requests for MySQL would be something like:
```sql
/*foreign keys*/    
SELECT column_name, constraint_name, referenced_table_name, referenced_column_name, table_name, ordinal_position
		FROM information_schema.key_column_usage
		WHERE referenced_table_name IS NOT NULL
		and table_schema = 'myschema'
		order by table_name,ordinal_position
			
/*primary keys*/  			
SELECT column_name, constraint_name, referenced_table_name, referenced_column_name, table_name, ordinal_position
		FROM information_schema.key_column_usage 
		WHERE table_schema = 'myschema'
		AND constraint_name = 'primary'
		
/*column info*/  			
SELECT * FROM information_schema.columns WHERE table_schema = 'myschema'
```

noamross:
Would you take a PR for this or is it dependent on your expected changes in #342 and #229?

krlmlr:
Thanks. Let me take another stab getting the internal structure right, happy to take PRs after that.

krlmlr:
@noamross: Now that we have an idea what `dm_meta()` should look like, we could take another stab.

I've found that, across databases, `constraint_column_usage` is the most inconsistent. It seems that `REFERENTIAL_CONSTRAINTS` is the thing to use in MySQL/MariaDB?

Another thing that gave me headaches for Postgres are compound keys -- `constraint_column_usage` seems to lack a crucial column that helps infer compound keys correctly, I worked around with a custom query against the native views.

krlmlr:
We're very close to an implementation. Do you have a chance to test it in the next few days?

noamross:
[...]

PR #1130, opened 2022-06-20
Exclude `information_schema` tables from dm by default if `schema = NULL`

noamross:
Addressing #366.  Does not address case of MSSQL using `dbname` instead of schema (yet), as I'm unfamiliar with how that differs.


krlmlr (commented):
Thanks!

krlmlr:
I'm sorry, I was confused.

#1123 now uses only objects from the current database if `DBI::dbConnect()` has been called with the `dbname` argument. This is consistent with the current behavior. We'll want to find a way to query all schemas even in that case, need to revisit the semantics of `schema = NA` vs. `schema = NULL` .

krlmlr:
`schema = NA` done in 78aa189b4adfb9edfaacdef891410a82d237c74a, but no tests.

Closing for now, I believe we want `information_schema` with `schema = NA` .

Issue #1076, opened 2022-06-13
Interactive creation of dm code


TSchiefer:
Since this morning we were questioning the way disentangling currently works, I was wondering if after all a small Shiny App wouldn't actually solve our problems.

The function starting the app would take one argument: the `dm` to disentangle.
The app has several elements:
1. Visual representation of the current state of the dm
2. A box with the code that can be copied to produce the current state of the `dm` from the input dm.
3. A "reset" button
4. An "undo last action" button
5. An info box that says one of: "Cycle detected" (red) or "Cycle-free" (green)
6. Four possible actions:
  - remove table (providing Dropdown menu with table list)
  - remove FK (providing Dropdown menu with FK list)
  - duplicate/multiply table (providing Dropdown menu with list of candidate tables for multiplication)
  - Current implementation of `dm_disentangle()`, taking a `start` table as an argument (providing Dropdown menu with table list); this one can only be used once, but the other three actions can still be used afterwards.

@moodymudskipper said he believes that people tend to prefer thinking about data models in a visual way, which I agree with. This approach would take this into account, plus, it's not very hard to implement. And we would give the user all the flexibility (and also responsibility) to create the cycle-free dm they want.

krlmlr:
Good idea, I'm curious what the "duplicate/multiply table" operation looks like in general.

Should we budget a day for a prototype?

TSchiefer:
One more thought: Maybe we shouldn't make it too sophisticated and with this operation really just add one copy of the table in question. 
If we do this via `dm_zoom_to() -> dm_insert_zoomed() (-> dm_rename_tbl()???)`, the FKs would be part of the new table and could be removed later on.

On the other hand, if we just add a copy of the table without the key info, we could in addition provide the option to add PKs and FKs in the Shiny App, potentially widening the range of the app's use cases.
We could also make it possible to rename and select tables in general in order to provide more flexibility.

krlmlr:
How about "splitting a table" as an atomic operation -- create one copy for each incoming foreign key into the table, copy all outgoing foreign keys?

As discussed yesterday, an esquisse-like Shiny app to prepare a dm object might be a useful addition anyway, with support for much simpler operations such as table selection, column selection/reordering, key removal/creation, deconstruction, ... . Let's sketch out the operations we want to support (and the code that we generate for those operations) before implementing the app.

TSchiefer:
> create one copy for each incoming foreign key into the table, copy all outgoing foreign keys

[...]

PR #1133, opened 2022-06-20
Draft middleware

krlmlr:
The idea here is to implement `enum_*()` functions for the entire functionality of the Shiny app. No logic in the UI part!

CC @christophsax.


scratch/code_generation_middleware.R
```diff
@@ -0,0 +1,67 @@
...
+ops <- enum_ops(dm)
+ops
+
+enum_ops(dm, op_name = "dm_add_pk")
+enum_ops(dm, op_name = "dm_add_pk", table_name = "parent")
+
+final_ops <- enum_ops(dm, op_name = "dm_add_pk", table_name = "parent", column_names = "a")
+final_ops
+
+cg_block <- new_cg_block(dm)
```
TSchiefer, line 60: We would normally use `expr(dm)`.
The way it is done here works too, but then the print method for code block prints it as: `<dm>`.
If we do use `expr(dm)` however, we run into a problem with environments, because `dm` is interpreted as the function `dm()`.
Shall we just use `quo()` for the input object and `eval_tidy()` in `cg_eval_block()` and tweak the print method slightly to avoid it being printed like `^dm`?

TSchiefer (approved):
This is very neat. I assume the idea is to display (or activate) dropdown menus in the shiny app for the newest list item that's been added? And that this should be extended to other functions at some point?

Only one remark, see comment

scratch/code_generation_middleware.R
krlmlr: Weird, `new_cg_block(expr(dm))` doesn't work for me.

TSchiefer: That's because of
> If we do use expr(dm) however, we run into a problem with environments, because dm is interpreted as the function dm().

scratch/code_generation_middleware.R
```diff
@@ -0,0 +1,68 @@
...
+        "dm_add_pk", # only if not all tables have PK
+        "dm_add_fk",
+        NULL
+      )
+    )
+  } else {
+    list2(
+      op_name = op_name,
+      !!!exec(paste0("enum_ops_", op_name), dm = dm, ...)
+    )
```
egnha, line 23: Just a minor question/comment (as I familiarize myself with the code)—Could this be simpler as as a direct function call?:
```R
c(op_name = op_name,
  get(paste0("enum_ops_", op_name))(dm = dm, ...))
```
(Ditto for `enum_ops_dm_add_pk()`.)

egnha:
Am I correct that, according to points 7 and 8 of Re: https://github.com/cynkra/dm/issues/1076#issuecomment-1154887556, the next steps of the middleware draft would be to add the following functions?
```R
enum_ops_dm_rm_pk <- function() ...
enum_ops_dm_rm_pk_table <- function() ...
```
(And additionally, register these names in `enum_ops()`.)

Presuming that, would these `enum_ops_dm_rm_pk_*()` functions be essentially the same as the corresponding `enum_ops_dm_add_pk_*()` functions, different only in the final call that they generate?

egnha:
I am a little confused about what these `enum_ops_*()` functions are supposed to do, because depending on whether `op_name`/`table_name` is supplied, they return rather different kind of objects: in the former case, essentially a bunch of names, in the latter case, a call (plus a record of supplied names, which I am thinking is a kind of attribute of the generated call). The downstream consumer `cg_add_call()` only seems to work on a call.

Granted that, should each `enum_ops_*()` not be rather two functions?

Other than `cg_add_call()` what other function could represent typical use of `enum_ops_*()`?

egnha:
About the public API: from the sample code, it seems to me that `enum_ops()` is intended to be exported, and the other functions are implementation details. Is that correct?

krlmlr:
The shiny app will only call enum_ops(). None of this needs to be exported for now.

scratch/code_generation_middleware.R
```diff
@@ -0,0 +1,68 @@
...
+  if (is.null(column_names)) {
+    # enumerate all columns that are not list
+    list(column_names = colnames(dm[[table_name]]))
+  } else {
+    list2(
+      column_names = column_names,
+      call = expr(dm_add_pk(., !!sym(table_name), !!!syms(column_names)))
+    )
+  }
+}
```
TSchiefer, line 49: Missed this earlier:
If column names is a vector of length > 1 this fails. Maybe:
```suggestion
enum_ops_dm_add_pk_table <- function(dm = NULL, ..., table_name, column_names = NULL) {
  if (is.null(column_names)) {
    # enumerate all columns that are not list
    list(column_names = colnames(dm[[table_name]]))
  } else if (length(column_names) > 1) {
    list2(
      column_names = column_names,
      call = expr(dm_add_pk(., !!sym(table_name), c(!!!syms(column_names))))
    )
  } else {
    list2(
      column_names = column_names,
      call = expr(dm_add_pk(., !!sym(table_name), !!sym(column_names)))
    )
  }
}
```

scratch/code_generation_middleware.R
krlmlr: Not sure, are there limitations with the current approach?

krlmlr:
Thanks. `enum_ops()` will be called by the Shiny app and will inspect the last element of the list returned by the function. This will typically represent the options available to the user. The return value is designed so that it can be reused in another call to `enum_ops()` after a choice has been made. Things will become clearer when we start using it from the Shiny app, for now it's fine to implement "similar" behavior for `dm_rm_pk()` and `dm_add_fk()` . Tests would also be a good idea to have at this stage.

Please feel free to push to this branch, or to a new branch -- whatever works best.

egnha:
Here is the beginning of a sketch for an alternative construction that simplifies some of the internal bookkeepking:
```R
ops <- local({
  dm_add_pk <- function(dm, ...) "Make names or dm_add_pk call."
  dm_add_fk <- function(dm, ...) "Make names or dm_add_fk call."
  # etc.
  environment()
})

enum_ops <- function(dm = NULL, ..., op_name = NULL) {
  if (is.null(op_name)) {
    names(ops)
  } else {
    op <- get(op_name, envir = ops)
    op(dm, ...)
  }
}
```
This directly expresses the idea that `enum_ops(op_name = nm)` generates a called named `nm` or a bunch of names, and that this generating function might as well just be called `nm`, thereby obviating a level of name indirection.

Moreover, it looks to me that the `ops` environment could be generated since its functions all seem to be cut from the same template (qua macro).

Thoughts?

egnha:
Here is the beginning of a sketch for an alternative construction that simplifies some of the internal bookkeepking:
```R
ops <- local({
  dm_add_pk <- function(dm, ...) "Make names or dm_add_pk call."
  dm_add_fk <- function(dm, ...) "Make names or dm_add_fk call."
  # etc.
  environment()
})

enum_ops <- function(dm = NULL, ..., op_name = NULL) {
  if (is.null(op_name)) {
    names(ops)
  } else {
    op <- get(op_name, envir = ops)
    op(dm, ...)
  }
}
```
This directly expresses the idea that `enum_ops(op_name = nm)` generates a called named `nm` or a bunch of names, and that this generating function might as well just be called `nm`, thereby obviating a level of name indirection.

Moreover, it looks to me that the `ops` environment could be generated since its functions all seem to be cut from the same template (qua macro).

Thoughts?

krlmlr:
Merging for now, let's discuss in new issues.

krlmlr:
Regarding possible simplifications with generic code: I'd like to do the first few steps manually.

In my sketch, `op_name` always stays `"dm_add_pk"` etc.

egnha:
Certainly, I'll open a new issue.

krlmlr:
The caller isn't aware of the detailed op names, and shouldn't be concerned.

krlmlr:
For the caller it's all `"dm_add_pk"` plus options.

krlmlr:
I'll start working on a first draft Shiny app with @christophsax now, with first results within 1-2 hours I hope.

krlmlr:
@egnha: Could you please take a look at what a first set of tests could look like for the existing functionality, even if it's not perfect?

krlmlr:
We will adapt later as necessary.,

krlmlr:
Does any of this make sense?

scratch/code_generation_middleware.R
egnha: > @egnha: Could you please take a look at what a first set of tests could look like for the existing functionality, even if it's not perfect?

Sure. As a new issue, first?

egnha:
> @egnha: Could you please take a look at what a first set of tests could look like for the existing functionality, even if it's not perfect?

Sure. As a new issue, first?

egnha:
> A first set of tests could go into a new PR, yes.
> 
> I opened #1148 which tweaks the middleware a bit further and adds two example scripts. I think we can use this pattern for now until we come up with something better or more suited for the UI. We now have a clearer idea what the UI should look like, and the middleware might require yet another round of modifications -- but we want to start somewhere.

Thanks for the heads-up.

I notice that three additional sample scripts have been added (code_generation_middleware_n.R, n=2..4). Just so that I'm clear about your intent (and expectation), am I to regard the output values in these scripts as assertions? (I.e., possible test expectations?)

egnha:
OK, thanks for the clarification. I'll extract some tests from them.

PR #1119, opened 2022-06-19
feat: New `nest_join()` and `pack_join()` for zoomed dm objects

IndrajeetPatil:
Closes #1055

Starting to get my feet wet in the codebase 😅

PR #1138, opened 2022-06-21
docs: Use selection helpers in `_pkgdown.yml`

IndrajeetPatil:
I think this is tidier and also future-proof.

For example, #1119 added `nest_join.zoomed_dm` and `pack_join.zoomed_dm`, but they are currently missing from the reference index. But, with the current selection helper approach, this wouldn't happen since it would pick up all functions with the pattern ".zoomed_dm$".

### DIFF
--- a/pkgdown/_pkgdown.yml
+++ b/pkgdown/_pkgdown.yml
@@ -28,27 +28,16 @@ reference:
 
 - title: Primary keys
   contents:
-  - dm_add_pk
-  - dm_has_pk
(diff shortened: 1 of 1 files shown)


IndrajeetPatil:
@maelle Can you please also have a look at this and see what you think? Thanks.

maelle (approved):
Looks good, good idea!

krlmlr:
Thanks!

PR #1145, opened 2022-06-21
use eval_tidy() (and expect quo(input_object))

TSchiefer:
then I think the script in https://github.com/cynkra/dm/pull/1133 would work with `cg_block <- new_cg_block(quo(dm))`


krlmlr:
Thanks!

PR #1150, opened 2022-06-21
Draft of test expectations for enum_ops()

egnha:
WIP: the initial draft consists of test expectations distilled from the scripts code_generation_middleware_*.R. The terrible test description, and layout of the tests, are, of course, provisional.

@krlmlr: For starters, are these test assertions correct?

### DIFF
--- a/tests/testthat/test-enum-ops.R
+++ b/tests/testthat/test-enum-ops.R
@@ -1,0 +1,33 @@
+test_that("enum_ops() works", {
+  parent <- tibble(a = c(1L, 1:3), b = -1)
+  child <- tibble(a = 1:4, c = 3)
+  dm <- dm(parent, child)
+  expect_equal(
+    enum_ops(dm),
+    list(input = list(dm = dm),
+         single = list(op_name = "dm_rm_fk"),
+         multiple = list(table_names = names(dm)))
+  )
+  expect_equal(
+    enum_ops(dm, op_name = "dm_add_pk"),
+    list(input = list(dm = dm, op_name = "dm_add_pk"),
+         single = `names<-`(list(), character(0)),
+         multiple = list(table_names = names(dm)))
+  )
+  expect_equal(
+    enum_ops(dm, op_name = "dm_add_pk", table_names = "parent"),
+    list(input = list(dm = dm,
+                      table_names = "parent",
+                      column_names = NULL,
+                      op_name = "dm_add_pk"),
+         multiple = list(column_names = colnames(dm[["parent"]])))
+  )
+  expect_equal(
+    enum_ops(dm, op_name = "dm_add_pk", table_names = "parent", column_names = "a"),
+    list(input = list(dm = dm,
+                      table_names = "parent",
+                      column_names = "a",
+                      op_name = "dm_add_pk"),
+         call = quote(dm_add_pk(., parent, a)))
+  )
+})


krlmlr:
Thanks, good start.

- Are you familiar with snapshot tests? Because we're trying to catch up with a moving target here, using this kind of tests will be easier to write and to maintain for now: https://testthat.r-lib.org/articles/snapshotting.html
- We need many more tests, using the above technique allows us to write them very quickly. We may have to blindly accept large-scale changes but will certainly see if a minor tweak leads to unexpected behavior anywhere
- Are we in a happy place for fast-paced development in this project? The existing test base is already quite large and takes some time to run, but work on the new API requires only a fraction of them. Thoughts?

egnha:
Thanks for your review and suggestions. Good idea. I'll have a look a snapshotting.

egnha:
One question: where would I get more tests?

The tests that I have written, limited in number though they are, already cover the range of behavior in the aforementioned scripts (code_generation_middleware_*.R).

egnha:
And about snapshotting (just so we're on the same page): are you suggesting to imprint code_generation_middleware_*.R as snapshots?

(The fact that they are filed under "scratch" implies to me that they might be provisional.)

egnha:
(As for the pace of the testing cycle, that is not yet an issue in the context of the testing here.)

krlmlr:
We would add more tests as we proceed with the development.

Please note that the example scripts differ subtly in the order of the argument specification. The last script specifies:

1. dm
2. table
3. columns
4. operation

because that's what seems to be most natural for the user interface. Is this all covered by the tests?

The functionality for removing PKs should be similar.

egnha:
> Please note that the example scripts differ subtly in the order of the argument specification. The last script specifies:
> 
> 1. dm
> 2. table
> 3. columns
> 4. operation
> 5. `force` argument if needed, with warning message
> 
> because that's what seems to be most natural for the user interface. Is this all covered by the tests?

I see. In this case, might it not be simpler to test that the signature (i.e., `formals()`) of `enum_ops()` has a certain form? (The validity of various calling forms would then be derived from R's calling convention.)

krlmlr:
We could test the signature, but we also want to test functionality.

egnha:
Of course. Sorry if I was unclear: I am merely suggesting that they be separate tests. (E.g., `f(x, y)` has expected values vs. `f(x, y)` and `f(y, x)` have same value.)

krlmlr:
`f(x, y)` and `f(y = y, x)` will have the same value by R's matching rules?

egnha:
No, not in general, of course, that would depend on the particular form of `formals(f)`. Maybe an example helps to clarify my point: If the point is that you want to ensure that `f(x, y)` and `f(y = y, x)` have the same value, it would follow from knowing that `f` has signature `function(x, y, ...) ...`. (This doesn't uniquely determine the formals—not the point here.) Does that make sense?

But back to my original question: it seems you want to express some invariants about the signature of `enum_ops()`, so can we just make such an assertion directly? (Which is not to say that the signature is uniquely determined or fixed.)

krlmlr:
Of course. We can agree to put all arguments past the ellipsis to ensure that all arguments are named.

Let's keep the tests as simple as possible for now, snapshot tests seem just fine. We need much much more functionality first, with tests, to get a better idea what the final product looks like. Moving target and such.

egnha:
Sure. Will snapshotting scratch/code_generation_middleware_*.R covering everything you have in mind?

krlmlr:
For now, yes. Thanks!

egnha:
Presently the following `exec()` is erroring and preventing snapshotting:
https://github.com/cynkra/dm/blob/0956cf0021b19f19ad62667ac42c6ddc8a8e1c1c/scratch/code_generation_middleware_2.R#L56-L67
The problem is a name ambiguity: Both `ops_3$input` and `user_choice_3` have a component named `column_names`.

More than one fix is possible (change `enum_ops()` or modify the present call, ...). The context here doesn't suggest what is right.

How should this be resolved?

krlmlr:
Let's skip this for now and discuss in a separate issue.

egnha:
The problematic script is skipped (325fb27bcc0e1d0646b0ac57462e7a90617d4f73).

Issue #1154, opened 2022-06-22
Named arguments to successive enum_ops() calls are not overridden


egnha:
Presently, the following `enum_ops()` call is erroring:
https://github.com/cynkra/dm/blob/0956cf0021b19f19ad62667ac42c6ddc8a8e1c1c/scratch/code_generation_middleware_2.R#L59-L67
The problem is a name clash: Both `ops_3$input` and `user_choice_3` have a component named `column_names`.

More than one fix is possible. (Change `enum_ops()`? Doctor the call?) The context here doesn't suggest what is right/intended.

Cf. https://github.com/cynkra/dm/issues/1150#issuecomment-1162858354_

egnha:
Yes, I think it makes sense to simplify `enum_ops()`, esp. as preparation for its extension to additional op-names (e.g., `"dm_add_fk"`, `"dm_rm_pk"`, etc.). I could try work on that.

egnha:
I'll work out a simplification of `enum_ops()`.

egnha:
I will fill in the details and test against the snapshot tests (tomorrow). PR TBD.

egnha:
I feel the same way. However, the `local()` is specifically here to constrain name-lookup. I think that legitimizes its use.

egnha:
> Thanks. This might work, should we take a look at what this looks like in practice?

Sure, I'll fill in the elided parts tomorrow. (Essentially, a copy-paste of the branching logic, no new idea vis-à-vis code structure.)

egnha:
> If we need to constrain, could we switchpatch too?

Sorry, I don't quite understand this question. Could you elaborate, please?

egnha:
Could this remark help: `enum()` is basically a `get()`. The `get()` has to be constrained—there is a fixed set of allowed `dm_*` operations.

krlmlr:
I believe I see what it does. I just don't particularly like it. All those functions inside the `local()` need to be also accessible via direct function call. The same results could also be achieved with a fixed prefix or suffix that is only used for those functions, and we could add a snapshot test for those.

Switchpatching would be something like the following, if I understand that term correctly:

```r
switch(op,
  dm_add_pk = op_dm_add_pk,
  dm_add_fk = op_dm_add_fk,
  ...
)
```

egnha:
I see what you mean. Something like that could also work.

PR #1162, opened 2022-06-23
Simplification of enum_ops()

egnha:
Cf. #1154.

This PR simplifies `enum_ops()` as follows:

- Makes explicit the use of `...` to ensure that certain arguments (`table_names`, etc.) are called by name; in particular, superfluous dots are removed downstream
- Expresses `enum_ops()` as a function composition to express the fact that (list) components of the output are created in a specific order, with specific argument dependencies
- Expresses call creation as an explicit dispatch on the operator name (the `switch()` in `enum()`)

CC @krlmlr

### DIFF
--- a/R/enum_ops.R
+++ b/R/enum_ops.R
@@ -1,119 +1,83 @@
-enum_ops <- function(dm = NULL, ..., table_names = NULL, column_names = NULL, op_name = NULL) {
-  # FIXME: Implement choosing dm or connection object from .GlobalEnv
-  stopifnot(!is.null(dm))
-
-  if (!is.null(table_names) && !is.null(column_names) && !is.null(op_name)) {
-    return(
-      list2(
-        input = list2(
-          dm = dm,
-          table_names = table_names,
-          column_names = column_names,
-          op_name = op_name,
-          ...,
-        ),
-        !!!exec(
-          paste0("enum_ops_", op_name),
-          dm = dm, ...,
-          table_names = table_names,
-          column_names = column_names,
-          op_name = op_name
-        )
-      )
-    )
-  }
-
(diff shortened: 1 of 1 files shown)


R/enum_ops.R
```diff
@@ -1,119 +1,77 @@
...
-    if (length(dm) > 0 && is.null(column_names)) {
-      op_name <- c(op_name, "dm_rm_fk")
+    if (length(dm) > 0 && is.null(cols)) {
+      op <- c(op, "dm_rm_fk")
     }
+    list(input = list(dm = dm), single = list(op_name = op))
   } else {
-    input <- c(input, op_name = op_name)
-    op_name <- NULL
+    nil <- `names<-`(list(), character())
```
krlmlr, line 31: From rlang:

```suggestion
    nil <- set_names(list(), character())
```

R/enum_ops.R
```diff
@@ -1,119 +1,77 @@
...
-  if (dm_has_pk(eval_tidy(dm), !!sym(table_names))) {
-    out$call <- as.call(c(as.list(out$call), force = TRUE))
-    out$confirmation_message <- "This table already has a primary key. Please confirm overwriting the existing primary key."
   }
-
   out
 }
+
+any_null <- function(...) {
+  any(vapply(list(...), is.null, TRUE))
```
krlmlr, line 76: From purrr:

```suggestion
  detect_index(list(...), is.null) != 0
```

krlmlr (approved):
Thanks! I see how `[[` is stricter than `$`, but it also hampers readability a bit. No need to take concrete action now, though.

krlmlr:
Thanks!

PR #1164, opened 2022-06-23
Fix regression: compound columns must be grouped with c()

egnha:
#1162 introduced a regression: it produced incorrect `dm_add_pk()` calls for compound primary keys, e.g., `dm_add_pk(., table, col1, col2)` instead of `dm_add_pk(., table, c(col1, col2))`.


egnha:
Cf. https://github.com/cynkra/dm/pull/1133#discussion_r902452398

krlmlr:
Thanks, good catch (and great we have a test too!)

PR #1184, opened 2022-07-04
get_src_table_names() uses dm_meta()

krlmlr:
This is more complicated than it should be, because `get_src_table_names()` does some implicit filtering on the default schema, and the learn tests fail (but in parallel only):

```sh
DM_TEST_SRC=postgres R -q -e 'testthat::test_local()'
```


krlmlr:
Will make a clean start.

Issue #676, opened 2021-10-18
Which functions and parameters are stable now?


krlmlr:
Review, mark explicitly.

krlmlr:
Done now.

PR #1192, opened 2022-07-05
Stabilize dm_disambiguate_cols()

krlmlr:
Part of #1039. For #676.


krlmlr:
The dots are not used, no need to take action.

Issue #979, opened 2022-04-25
Some unexported functions are not called by any other function


moodymudskipper:
```
"abort_key_constraints_need_db"  "abort_no_src_or_con"           
"abort_temp_table_requested"     "abort_unnamed_table_list"      
"abort_wrong_col_names"          "class_to_db_class"             
"dm_test_load"                   "get_db_table_names"          
"nest_join_zoomed_dm"            "sql_schema_table_list_mssql"   
"sql_schema_table_list_postgres"
```

Most or all of them have tests written for them too, are they still useful ?

krlmlr:
How did you come up with that list? I see `sql_schema_table_list_postgres()` is used in tests.

I wonder if we can parse the full coverage results (running against all backends) to determine functions that aren't called by any test.

PR #1197, opened 2022-07-05
Remove unused internal functions

IndrajeetPatil:
IINM, these are no longer needed; at least, they are not used anywhere in the codebase.


krlmlr:
Thanks. I prefer to keep the `aaa-*.R` for now, we might use them at some point, and by that time I'll have forgotten where to look. We could exclude them from covr though.

IndrajeetPatil:
Sounds good!

moodymudskipper:
@IndrajeetPatil relevant: https://github.com/cynkra/dm/issues/979

Issue #1131, opened 2022-06-20
dm_examine_constraints() should honor implicit unique keys created by foreign keys


krlmlr:
Both dm objects should return the same results for `dm_examine_constraints()`, that's what databases enforce too.

``` r
library(dm)
library(tibble)

parent <- tibble(a = c(1L, 1:3))
child <- tibble(a = 1:4)

dm_unique_key <-
  dm(parent, child) %>%
  dm_add_fk(child, a, parent, a)

dm_unique_key %>%
  dm_draw()
```

![](https://i.imgur.com/UVUkWlV.png)

``` r
dm_unique_key %>%
  dm_examine_constraints()
#> ! Unsatisfied constraints:
#> • Table `child`: foreign key `a` into table `parent`: values of `child$a` not in `parent$a`: 4 (1)


dm_primary_key <-
  dm(parent, child) %>%
  dm_add_pk(parent, a) %>%
  dm_add_fk(child, a, parent)

dm_primary_key %>%
  dm_draw()
```

![](https://i.imgur.com/LWoukYt.png)

``` r
dm_primary_key %>%
  dm_examine_constraints()
#> ! Unsatisfied constraints:
#> • Table `parent`: primary key `a`: has duplicate values: 1 (2)
#> • Table `child`: foreign key `a` into table `parent`: values of `child$a` not in `parent$a`: 4 (1)
```

<sup>Created on 2022-06-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

smgogarten:
This has the (perhaps unintended?) consequence of preventing users from creating many-to-many tables (see #1365 ). Is there a compromise possible?

PR #1209, opened 2022-07-07
feat: honor unique keys in parent table for weak fks

TSchiefer:
a test is implemented as well.

closes https://github.com/cynkra/dm/issues/1131

### DIFF
--- a/R/examine-constraints.R
+++ b/R/examine-constraints.R
@@ -109,6 +109,7 @@ check_pk_constraints <- function(dm, progress = NA, top_level_fun = NULL) {
       kind = character(),
       column = new_keys(),
       ref_table = character(),
+      is_pk = logical(),
       is_key = logical(),
       problem = character()
     ))
--- a/R/examine-constraints.R
+++ b/R/examine-constraints.R
@@ -77,7 +77,7 @@ print.dm_examine_constraints <- function(x, ...) {
       # FIXME: Use cli styles
       mutate(text = paste0(
         "Table ", tick(table), ": ",
-        kind_to_long(kind, is_pk), " ",
+        kind_to_long(kind), " ",
         format(map(problem_df$columns, tick), justify = "none"),
         into,
         ": ", problem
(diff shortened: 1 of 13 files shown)


R/examine-constraints.R
```diff
@@ -90,11 +90,23 @@ print.dm_examine_constraints <- function(x, ...) {
...
 
 check_pk_constraints <- function(dm, progress = NA, top_level_fun = NULL) {
-  pks <- dm_get_all_pks_impl(dm)
+  pks <- dm_get_all_pks_impl(dm) %>%
+    mutate(kind = "PK") %>%
+    bind_rows(
+      dm_get_all_fks_impl(dm) %>%
+        select(table = parent_table, pk_col = parent_key_cols) %>%
+        distinct() %>%
+        mutate(kind = "UK")
```
krlmlr, line 107: Extract `dm_get_all_uks_impl(dm)` ?

R/examine-constraints.R
```diff
@@ -90,11 +90,23 @@ print.dm_examine_constraints <- function(x, ...) {
...
+    kind == "UK" ~ "unique key",
+    .default = "foreign key"
+  )
 }
 
 check_pk_constraints <- function(dm, progress = NA, top_level_fun = NULL) {
-  pks <- dm_get_all_pks_impl(dm)
+  pks <- dm_get_all_pks_impl(dm) %>%
+    mutate(kind = "PK") %>%
+    bind_rows(
```
krlmlr, line 103: `bind_rows(PK = pks, UK = uks, .id = "kind")` ?

krlmlr:
The snapshot update is extra weird, sorry about that.

krlmlr (approved):
Thanks, I expect more PRs coming in after the weird snapshot update. We could do one more iteration and merge, or leave it as is.

R/examine-constraints.R
TSchiefer: sadly, `.id = "kind"` doesn't take the info from the existing column "kind", but instead creates a new one with values "1" or "2". Can still make use of it though.

krlmlr: Also with named arguments?

krlmlr: "1" and "2" seem dangerous.

TSchiefer: one step ahead of me, there, nice trick!

TSchiefer:
BTW, had chosen "fix", because the issue was labeled as a bug.

krlmlr:
Yeah, let's not be too strict with ourselves. It was *also* labeled an enhancement. 🙃

TSchiefer:
So I think this PR might be OK now (confused about the check-failures though). Let me know if I can still do something here.

PR #1222, opened 2022-07-08
docs: Add contributing guide

### LINKED ISSUE #757: CONTRIBUTING.md
Guidelines to collaborate on {dm}

I suppose a build ignored md file in the root folder is fine ? `COLLAB.md` ? unless there is a standard for that already ?

The goal is to facilitate external collaboration and onboarding, and structure our approach of development, i.e. write the unwritten rules and some more.

For example in no special order: 

* function names
* tests
* how we do to test multiple data sources
* where to declare global variables
* reference to tidyverse principles
* error helpers
* ...

It should be built incrementally and with no pretense of exhaustivity or perfect structure at first, it can live on its [own branch](https://github.com/cynkra/dm/tree/f757-collaboration-guide) until it has enough content and is structured enough, then we can merge it and still update it regularly.

I suppose when we get enough there we can isolate specific dm guidelines from general Cynkra guidelines.

I think this thread is also a good place for questions and short discussions on specific aspects.

maelle:
Fix #757

Using text by @moodymudskipper added on top of the tidy contributing guide

### DIFF
--- a/.github/.gitignore
+++ b/.github/.gitignore
@@ -1,1 +1,2 @@
 /pkg.lock
+*.html
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+# Contributing to dm
+
+This outlines how to propose a change to dm. 
+For more detailed info about contributing to this, and other tidyverse packages, please see the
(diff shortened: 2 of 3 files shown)


krlmlr:
Thanks!

Issue #1030, opened 2022-06-06
Decide about "questioning" APIs


krlmlr:
Create issues in this milestone if necessary.

- [x] Filtering
- [x] Enumeration of primary and foreign keys
- [x] Table surgery

krlmlr:
Uncertain only about filtering, but might as well promote it to "stable".

krlmlr:
Done now.

PR #1236, opened 2022-07-11
api: Stabilize `dm_filter()`, deprecate `dm_apply_filters()`, `dm_apply_filters_to_tbl()` and `dm_get_filters()`

krlmlr:
For #1030.


krlmlr:
This is a big PR, the commits kind of make sense individually. I'm looking for feedback specifically regarding API design, did I forget any use cases we had in the past?

PR #1245, opened 2022-07-11
docs: Fix column names in `?dm_get_all_pks`

TSchiefer:
man page says the output tibble contains a column `pk_cols`, but currently it's `pk_col`.

I didn't update the snapshots, since my package setup differs slightly from what is needed to produce the "correct" snapshots, so tests will fail initially.


krlmlr:
Yeah, let's adapt the man page.

Snapshots are also updated on GitHub, as PRs to the branch.

TSchiefer:
I now backtracked and just adapted the man page.
The snapshot PR after a235fcc looked partly fine, but there were some stranger parts that I didn't like so much.

krlmlr:
Thanks!

PR #1250, opened 2022-07-12
docs: Improve contrast for display of `dm_financial()` (#1073)

### LINKED ISSUE #1073: Contrast in diagrams
In the diagrams of the intro vignette I wonder whether the contrast between background and foreground is enough (for all eyes).

When choosing a dark color the header is fine but the body of the table might not be. Should the body of the table get no background but instead the "frame" of the table would?

maelle:
Fix #1073


krlmlr:
Thanks!

PR #1289, opened 2022-07-17
chore: error message when list supplied to dm

### LINKED ISSUE #1248: Provide better error when `list` is provided to `dm()`
The current error message is a bit cryptic since users can't know what is `def$data`:

``` r
library(dm)
library(nycflights13)

dm(list(weather, airports))
#> Error in `abort_dm_invalid()`:
#> ! This `dm` is invalid, reason: Not all entries in `def$data` are of class `data.frame` or `tbl_dbi`. Check `dm_get_tables()`.
```

<sup>Created on 2022-07-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

IndrajeetPatil:
Closes #1248

Snapshot incoming from GHA.


krlmlr:
Good point. Should we check the class of the objects, e.g. introduce `is_tbl` in addition to `is_dm` further below?

IndrajeetPatil:
This has too big of an [impact](https://github.com/cynkra/dm/pull/1658) to be considered in an off the cuff PR.

PR #1324, opened 2022-07-20
api: deprecate `dm_squash_to_tbl()` and instead provide argument `.recursive` for `dm_flatten_to_tbl()`; rename `start` to `.start` and `join` to `.join` (#1272)

### LINKED ISSUE #1272: Could flatten and squash be the same function with a different argument?
Something like `dm_flatten_to_tbl(recursive = TRUE)`

TSchiefer:
closes #1272


krlmlr:
@TSchiefer: I went "all in" here. What did I miss

Issue #1333, opened 2022-07-21
count() doesn't keep primary keys


krlmlr:
I tried implementing `dplyr_reconstruct()`, with no success so far.

``` r
library(conflicted)
library(dm)
library(tidyverse)

# pkgload::load_all()
library(dm)
library(dplyr)

dm <- dm_nycflights13()

tbls <-
  dm %>%
  dm_get_tables(keyed = TRUE)

flights <- tbls$flights
airports <- tbls$airports

by_origin <-
  flights %>%
  count(origin)

by_origin
#> # A tibble: 3 × 2
#> # Keys:     — | 0 | 4
#>   origin     n
#>   <chr>  <int>
#> 1 EWR      641
#> 2 JFK      602
#> 3 LGA      518

# PK for by_origin is missing
dm(flights, airports, by_origin) %>%
  dm_get_all_pks()
#> # A tibble: 1 × 2
#>   table    pk_col
#>   <chr>    <keys>
#> 1 airports faa
```

<sup>Created on 2022-07-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
More action items:

- `rename()` can't currently work, need `rename.keyed_tbl()`
- check all other verbs

TSchiefer:
FWIW, `count()` seems to behave properly when using a `dm_zoomed`:

``` r
suppressPackageStartupMessages({
  library(dm)
  library(dplyr)
})

dm_nycflights13() %>% dm_zoom_to(airlines) %>% count(carrier) %>% dm_update_zoomed() %>% dm_get_all_pks()
#> # A tibble: 4 × 3
#>   table    pk_col            autoincrement
#>   <chr>    <keys>            <lgl>        
#> 1 airlines carrier           FALSE        
#> 2 airports faa               FALSE        
#> 3 planes   tailnum           FALSE        
#> 4 weather  origin, time_hour FALSE
dm_nycflights13() %>% dm_zoom_to(airlines) %>% count(name) %>% dm_update_zoomed() %>% dm_get_all_pks()
#> # A tibble: 3 × 3
#>   table    pk_col            autoincrement
#>   <chr>    <keys>            <lgl>        
#> 1 airports faa               FALSE        
#> 2 planes   tailnum           FALSE        
#> 3 weather  origin, time_hour FALSE
```

<sup>Created on 2023-08-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

[...]

PR #1334, opened 2022-07-21
Try dplyr_reconstruct()

krlmlr:
For #1333.


krlmlr:
The current attempt looks better. Action items:

- In the resurrected key info from the template, check if columns still exists; remove keys where columns don't exist
- Combine key info from data and template

End result: `count()` works.

Issue #835, opened 2022-02-10
More progress bars


krlmlr:
- [ ] `dm_wrap_tbl()`
- [ ] `dm_unwrap_tbl()`
- [ ] `dm_apply_filters()`

moodymudskipper:
needs #752

moodymudskipper:
@krlmlr given that `dm_apply_filters()` is deprecated, should it be modified to get the feature ? 
It would be small work but not sure if it's good practice to update a deprecated function.

PR #1447, opened 2022-09-01
chore: Refactor `dm_unwrap_tbl()` so it builds a "unwrap plan" first

### LINKED ISSUE #1446: Split planning and doing in dm_unwrap_tbl()
Complement to #752

moodymudskipper:
Closes #1446 

This paves the way for printing the atomic nest/pack operations (is an issue open for this ?) and #835

I stumbled on an interesting bug with `bind_rows()` along the way, which we might keep in mind since we're playing a lot with packed tables now : https://github.com/tidyverse/dplyr/issues/6447


R/dm_wrap.R
```diff
@@ -119,37 +119,50 @@ dm_wrap_tbl_plan <- function(dm, root) {
...
-  }
+  unpack_plan <-
+    tibble(
+      action = "dm_unpack_tbl",
+      table = table_name,
+      col = parents
+    )
+
+  unwrap_plan_from_children <-
+    # note: we cannot use bind_rows() because of https://github.com/tidyverse/dplyr/issues/6447
```
krlmlr, line 156:
```suggestion
    # note: we cannot use bind_rows() because of https://github.com/tidyverse/dplyr/issues/6447,
    # or even vec_rbind() because of https://github.com/r-lib/vctrs/issues/1640
```

krlmlr (commented):
Thanks! `map_dfr()` and `transpose()` are doing a ton of unnecessary work, should we use simple `map()` instead and omit `transpose()` ?

I also wonder how this will work for JSON data where we don't have access to the structure of the children, especially when this is a lazy source.

R/dm_wrap.R
moodymudskipper: > `map_dfr()` and `transpose()` are doing a ton of unnecessary work, should we use simple `map()` instead and omit `transpose()` ?

In `imap_fr(dm_get_tables(dm), dm_unwrap_tbl_plan)` `dm_unwrap_tbl_plan()` returns data frames, not just rows. So just using `imap()`and `transpose()` is not enough, I'd need to `split.data.frame()` every input from `imap()` too. Would it be better ?

> I also wonder how this will work for JSON data where we don't have access to the structure of the children, especially when this is a lazy source.

Yes I figured we'd tackle this as a next step. A problem with json is that I believe there are no zero length elements, so no structure to be found in nested elements where we'd have a zero row data frame in the standard version. To avoid collecting everything we might filter on non empty elements and fetch the first.
We can use a generic at this place in the code.

moodymudskipper:
See latest commit "Avoid map_dfr() and transpose()". I'm not sure if it's better.

krlmlr:
Thanks! You're right it's not better, but I forgot this is transient anyway, because we can't rely on "nested" and "packed" classes.

PR #1448, opened 2022-09-02
Use default schema in `dm_learn_from_db()`

### LINKED ISSUE #1440: `dm_from_con()` does not default to `schema = "public"` for Postgres
Even though it says so in the documentation.

This is due to an early exit

```r
if (is_null(table_names)) {
  return(dm_learned)
}
```

mgirlich:
Fixes #1440.
There already is a test
`test_that("Standard learning from MSSQL (schema 'dbo') or Postgres (schema 'public') and get_src_tbl_names() works?"`
but it seems it doesn't work correct. I had a quick look at it but wasn't quite sure how to fix it. @krlmlr maybe you want to have a look at the test.


R/db-helpers.R
```diff
@@ -170,9 +187,14 @@ schema_postgres <- function(con, schema) {
   schema
 }
 
+ident_q <- function(...) {
+  # to avoid dependency on dbplyr define `ident_q()` here (not exported from dplyr)
+  structure(c(...), class = c("ident_q", "ident", "character"))
+}
+
```
krlmlr, line 194: We're using `dbplyr::ident_q()` elsewhere.

R/db-helpers.R
```diff
@@ -170,9 +187,14 @@ schema_postgres <- function(con, schema) {
...
 schema_mariadb <- function(con, schema) {
   if (is_null(schema)) {
-    schema <- sql("database()")
+    schema <- ident_q("database()")
```
krlmlr, line 197: What's the advantage of using `ident_q()` here?

R/db-helpers.R
```diff
@@ -155,6 +147,31 @@ get_src_tbl_names <- function(src, schema = NULL, dbname = NULL) {
...
+
+  if (is_mssql(src)) {
+    # MSSQL
+    schema_mssql(con, schema)
+  } else if (is_postgres(src)) {
+    # Postgres
+    schema_postgres(con, schema)
+  } else if (is_mariadb(src)) {
+    # MariaDB
+    schema_mariadb(con, schema)
```
krlmlr, line 171: Weird, we're using a filter with `DATABASE()` elsewhere. Do we still need that?

krlmlr (commented):
Thanks.

@TSchiefer: Could you please help with the tests?

R/db-helpers.R
mgirlich: It works better with `%in% !!...`:

``` r
library(dplyr, warn.conflicts = FALSE)
library(dbplyr, warn.conflicts = FALSE)
lf <- lazy_frame(x = 1)

lf %>% filter(x %in% !!sql("1"))
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (`x` IN 1)
lf %>% filter(x %in% !!ident_q("1"))
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (`x` IN (1))
```

<sup>Created on 2022-09-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

krlmlr:
@mgirlich: Would you like to do another iteration?

krlmlr:
Thank you for your patience. I think we want `schema = NULL` to get all tables from all schemas in dm. It's way easier to change the documentation 😏 .

krlmlr:
Scratch that, `schema = NULL` will do all schemas, `schema = NA` will do the "sensible" thing depending on the database. I still might go with a different implementation.

R/db-helpers.R
krlmlr: Seems like we do.

krlmlr (approved):

krlmlr:
Thanks!

PR #1449, opened 2022-09-02
feat: `dm_from_con()` can use multiple schemata

### LINKED ISSUE #1441: `dm_from_con()` should accept multiple schemas
Currently, only one schema is accepted. 

```r
dm::dm_from_con(
  con,
  schema = c("a", "b")
)

#> Error in `abort_parameter_not_correct_length()`:
#> ! Parameter `schema` needs to be of length `1` but is of length 2 (`a`, `b`).
```

In my case I want to create a dm for two schemas (public + another one).

mgirlich:
Fixes #1441.


R/meta.R
```diff
@@ -241,7 +245,7 @@ filter_dm_meta <- function(dm_meta, catalog = NULL, schema = NULL) {
     constraint_column_usage <- constraint_column_usage %>% filter(table_catalog %in% !!catalog)
   }
 
-  if (!is.null(schema) && !is.na(schema)) {
+  if (!is.null(schema) && isTRUE(!is.na(schema))) {
     schemata <- schemata %>% filter(schema_name %in% !!schema)
     tables <- tables %>% filter(table_schema %in% !!schema)
     columns <- columns %>% filter(table_schema %in% !!schema)
```
mgirlich, line 251: The condition below

`} else if (!isTRUE(is.na(schema)) && is_mariadb(dm_get_con(dm_meta))) {`

is in my opinion confusing. I think the idea is to use a default for MariaDB if `schema` is `NULL`. If so, this should use `is.null(schema)`. Also, it might be better to implement these defaults differently. There is a function `schema_mariadb()` which seems to implement this default. But it always uses `database()` and here the default handling is different in case `database()` is NULL.

krlmlr: IIRC `DATABASE()` can be `NULL` in MariaDB/MySQL if connecting without a default database, and this could be the reason for this ... hack?

R/meta.R
```diff
@@ -225,6 +225,10 @@ filter_dm_meta <- function(dm_meta, catalog = NULL, schema = NULL) {
   force(catalog)
   force(schema)
 
+  if (length(schema) > 1 && any(is.na(schema))) {
```
krlmlr, line 228:
```suggestion
  if (length(schema) > 1 && anyNA(schema)) {
```

tests/testthat/test-learn.R
```diff
@@ -483,4 +483,9 @@ test_that("dm_from_con() with mariaDB", {
   expect_snapshot_output(my_dm <- dm_from_con(my_db))
   expect_snapshot(dm::dm_get_all_fks(my_dm))
   expect_snapshot(dm::dm_get_all_pks(my_dm))
+
+  # multiple schemata work
+  expect_snapshot_output(my_dm <- dm_from_con(my_db, schema = c("Accidents", "Ad")))
+  expect_snapshot(dm::dm_get_all_fks(my_dm))
+  expect_snapshot(dm::dm_get_all_pks(my_dm))
```
krlmlr, line 490: The test output looks odd, perhaps because the connection uses `dbname = "Financial_ijs"` ?

I wonder what's a good test design on Postgres and MSSQL for this feature.

krlmlr (commented):
Thanks! Should we document this explicitly?

tests/testthat/test-learn.R
mgirlich: Maybe simply setup our own DB.

R/meta.R
```diff
@@ -241,7 +245,7 @@ filter_dm_meta <- function(dm_meta, catalog = NULL, schema = NULL) {
     constraint_column_usage <- constraint_column_usage %>% filter(table_catalog %in% !!catalog)
   }
 
-  if (!is.null(schema) && !is.na(schema)) {
+  if (!is.null(schema) && all(!is.na(schema))) {
```
krlmlr, line 248:
```suggestion
  if (!is.null(schema) && !anyNA(schema)) {
```

krlmlr:
Thanks!

PR #1455, opened 2022-09-05
remove buggy and redundant test

TSchiefer:
see https://github.com/cynkra/dm/pull/1448#issue-1359846466

the test "Standard learning from MSSQL (schema 'dbo') or Postgres (schema 'public') and get_src_tbl_names() works?" was actually duplicated. The second version (https://github.com/cynkra/dm/blob/main/tests/testthat/test-learn.R#L44-L108) worked as intended, but for Postgres all sorts of additional (system?) tables are also included which made it necessary to manually subset the dm-tables in the test (was already like that all along; seems we neglected to turn this bug into an issue when we noticed it).

Since it will be solved in https://github.com/cynkra/dm/pull/1448, we should remove the subsetting in the test in this PR.


krlmlr:
Thanks!

PR #1459, opened 2022-09-06
Fix raw `NAMESPACE` for `{glue}`

IndrajeetPatil:
There is no function called `glue::collapse()`.

This causes problems for lintr:
https://github.com/r-lib/lintr/issues/1529


MichaelChirico:
There is `glue::glue_collapse()`, maybe that was intended?

MichaelChirico:
yes -- just suggesting there's two options for this PR, (1) assume the exclude= is spurious, delete it (2) assume glue_collapse was intended, fix the typo

I think we can decide (1) or (2) based on glue_collapse usage internally? my guess is (1) is right if this issue hasn't come up before btw.

krlmlr:
Thanks. Do we even need `rawNamespace` then?

krlmlr:
I'm sympathetic to the use of lintr, but we need to close a few PRs first to fully embrace it.

IndrajeetPatil:
> Thanks. Do we even need rawNamespace then?

No, we don't. I will change that.

> I'm sympathetic to the use of lintr, but we need to close a few PRs first to fully embrace it.

Sounds good.

krlmlr:
Thanks!

PR #1453, opened 2022-09-04
Handle name collisions when wrapping

moodymudskipper:
* Draft for discussion, implementing ideas from #1046 .
* Closes #867 (if we go this route)
* On top of  #1451 , better to merge after #1446 , #1450, #1447 

We use suffixes in col names and remove the need for a ptype and for"nested" and "packed" column classes.

* Nested cols are suffixed with ">"
* Packed cols are suffixed with "<"
* Joining columns are kept in nested and packed tables and are renamed `x=y` where `x` is the original name and `y` is the name of the key in the packing or nesting table
* Primary keys in packed/nested data are suffixed with "*", on top of possible name modification above

The code is not cleaned up yet (`ptype` is still there though not used, a couple comments to address for robustness, some commented code kept for convenience in case we want backward compatibility)

See below a demo, the commented code is how it worked before.

``` r
suppressMessages(devtools::load_all("~/git/dm"))

dm_packed <- dm_pack_tbl(dm_for_filter(), tf_1)
dm_packed
#> ── Metadata ────────────────────────────────────────────────────────────────────
#> Tables: `tf_2`, `tf_3`, `tf_4`, `tf_5`, `tf_6`
#> Columns: 19
#> Primary keys: 5
#> Foreign keys: 4
dm_packed$tf_2
#> # A tibble: 6 × 5
#>   c            d e        e1 `tf_1<`$b $`a=d*`
#>   <chr>    <int> <chr> <int> <chr>       <int>
#> 1 elephant     2 D         4 B               2
#> 2 lion         3 E         5 C               3
#> 3 seal         4 F         6 D               4
#> 4 worm         5 G         7 E               5
#> 5 dog          6 E         5 F               6
#> 6 cat          7 F         6 G               7

dm_packed_nested <- dm_nest_tbl(dm_packed, tf_2)
dm_packed_nested
#> ── Metadata ────────────────────────────────────────────────────────────────────
#> Tables: `tf_3`, `tf_4`, `tf_5`, `tf_6`
#> Columns: 15
#> Primary keys: 4
#> Foreign keys: 3
dm_packed_nested$tf_3
#> # A tibble: 10 × 4
#>    f        f1 g     `tf_2>`         
#>    <chr> <int> <chr> <list>          
#>  1 C         2 one   <tibble [0 × 5]>
#>  2 C         3 two   <tibble [0 × 5]>
#>  3 D         4 three <tibble [1 × 5]>
#>  4 E         5 four  <tibble [2 × 5]>
[...]

Issue #867, opened 2022-03-29
Handle name collisions in dm_wrap_tbl()


krlmlr:
`nest_join()` and `pack_join()` (after #865) silently overwrite matching column names: if the LHS contains a column named the same as the `name` argument, it is gone after the join. This is fine, but I wonder if we can handle this better (or more consistently) in `dm_wrap_tbl()`:

``` r
library(dm)
library(tidyverse)

dm(a = tibble(b = integer()), b = tibble(b = integer())) %>%
  dm_add_fk(a, b, b, b) %>%
  dm_wrap_tbl(a) %>% 
  pull_tbl(a)
#> # A tibble: 0 × 1
#> # … with 1 variable: b <packed[,0]>

dm(a = tibble(c = integer()), b = tibble(b = integer())) %>%
  dm_add_fk(a, c, b, b) %>%
  dm_wrap_tbl(a) %>% 
  pull_tbl(a)
#> # A tibble: 0 × 2
#> # … with 2 variables: c <int>, b <packed[,0]>
```

<sup>Created on 2022-03-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
I suspect there are many more edge cases if we also take `dm_unwrap_tbl()` into account. Should we counter this by disambiguating table names instead (i.e., if tables are named the same as any columns, they get a `...0` suffix)?

``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(tidyverse)

dm <- 
  dm(a = tibble(b = 0L), b = tibble(b = 0L)) %>% 
  dm_add_fk(a, b, b, b)

dm %>%
  dm_wrap_tbl(a) %>% 
  dm_unwrap_tbl(dm) %>% 
  dm_get_tables()
#> Error in def$table[[where]]: attempt to select less than one element in get1index

dm <- 
  dm(a = tibble(c = 0L), b = tibble(b = 0L)) %>% 
  dm_add_fk(a, c, b, b)

dm %>%
  dm_wrap_tbl(a) %>% 
  dm_unwrap_tbl(dm) %>% 
  dm_get_tables()
#> Error in `select()`:
#> ! Names must be unique.
#> x These names are duplicated:
#>   * "b" at locations 1 and 2.
```

<sup>Created on 2022-03-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

[...]

PR #1516, opened 2022-09-12
Handle name collisions when wrapping (no suffix)

moodymudskipper:
Towards #867.

After discussion in #1453 we propose unwrapping without using nesting/packing cols nor nested/packed cols to store constraints metadata.

When unnesting :
* we know the PK of the parent but we lost the FK relationship and the PK of the child
* The parent key in the nesting parent is assumed to be its PK
* The foreign key in the nested child is the first found combination of columns of compatible types which contains a subset of the parent key 
* The pk of the child to unnest is found by testing if the nth leftmost columns form a unique key, iterating from 1 to the number of cols.

When unpacking :
* We know nothing about the constraints relevant to the unpacking operation
* The pk of the parent to unpack is found by testing if the nth leftmost columns form a unique key, iterating from 1 to the number of cols.
* The parent key is assumed to be the parent's pk
* The FK in the packing child is the first found combination of columns of compatible types which contains a subset of the parent key 

See below the original `dm_for_filter()` followed by its wrapped then unwrapped counterpart.

``` r
dm1 <- dm_for_filter()
dm_draw(dm1, view_type = "all")
```

``` r
dm_wrapped <- dm_wrap_tbl(dm1, tf_4)
dm_unwrapped <- dm_unwrap_tbl(dm_wrapped, dm1)
dm_draw(dm_unwrapped, view_type = "all")
```

We see a lot of constraints got mixed up but we get the right diagram shape.

Next steps as I see them if we pursue in this direction:
* More efficient detection of constraints, even on the flights dm this runs forever because of expand_grid() + rowwise(). I think that for each new candidate we need to test if the n first candidates are a subset of the n first parent keys. Right now we compute every type compatible combination first and it goes into the millions.
* Add the key information to the unwrap plan arguments
* update `dm_unnest_tbl()` and `dm_unpack_tbl()` with arguments to set the keys.
  * `dm_unnest_tbl(dm, parent_table, col, ptype)` might become `dm_unnest_tbl(dm, parent_table, col, parent_key, child_fk, child_pk)`
  *  `dm_unpack_tbl(dm, child_table, col, ptype)` might become `dm_unnest_tbl(dm, child_table, col, child_fk , parent_key, parent_pk)`
  * OR instead of `parent_key` and `child_fk` args we use a {dplyr} join like syntax (as in the by arg), 
  * Print calls built from plan so user can adjust the keys.

### DIFF
--- a/R/dm_nest_tbl.R
+++ b/R/dm_nest_tbl.R
@@ -150,8 +150,18 @@ dm_pack_tbl <- function(dm, parent_table, into = NULL) {
   def <- dm_get_def(dm, quiet = TRUE)
   table_data <- def$data[def$table == table_name][[1]]
   child_data <- def$data[def$table == child_name][[1]]
-  packed_data <- pack_join(child_data, table_data, by = set_names(parent_fk, child_fk), name = table_name)
-  class(packed_data[[table_name]]) <- c("packed", class(packed_data[[table_name]]))
+  #packed_data <- pack_join(child_data, table_data, by = set_names(parent_fk, child_fk), name = table_name)
+  #class(packed_data[[table_name]]) <- c("packed", class(packed_data[[table_name]]))
+  parent_fk_suffixed <- paste0(parent_fk, "=", child_fk)
+  # FIXME: fail if weird names exist already
+  table_data <- rename(table_data, !!! set_names(parent_fk, parent_fk_suffixed))
+  packed_data <- pack_join(
+    child_data,
+    table_data,
+    by = set_names(parent_fk_suffixed, child_fk),
+    name = paste0(table_name, "<"),
+    keep = TRUE
+  )
 
   # update def and rebuild dm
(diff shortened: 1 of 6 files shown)


moodymudskipper:
The detection of constraints work ok in the example above in the sense that some type and value compatible keys are found. This is not the general case, however, in this case we'd fail without having a plan to edit...

Another approach is to not test compatibility at all, just assume unknown keys are first and single when the info is not available, and don't bother checking types and subsets. This would be fast and robust (in the sense of not crashing).

krlmlr:
Can you please resolve conflicts?

krlmlr:
Closing for now, added a comment to the original issue. This is too much to unpack right now. (Couldn't resist the pun.)

krlmlr:
Thanks! I took what I could from this PR before closing it.

Issue #1533, opened 2022-09-22
`dm_from_con()` with multiple schemata fails with opaque error


owenjonesphe:
## Cause

Following the changes from #1449, we can now have multiple schemata passed through - so the [filter condition](https://github.com/cynkra/dm/blob/e4c9f834ab20671c19fbb9956f1b253337127d8a/R/db-helpers.R#L149) in `dm:::get_src_tbl_names()` needs to be tweaked so that we're not using `==` to compare against a potentially multi-element vector.

I'll open a (tiny!) PR to address this ☺️ 

## Minimal example

```r
# Assuming some valid connection `conn` exists already - in my case, a postgres connection
src <- dm:::src_from_src_or_con(conn)
schema <- c("one", "two")

dm:::get_src_tbl_names(src, schema = schema)
```
```
Error in postgresqlExecStatement(conn, statement, ...) : 
  RPosgreSQL error: could not Retrieve the result : ERROR:  operator does not exist: information_schema.sql_identifier = record
HINT:  No operator matches the given name and argument type(s). You may need to add explicit type casts.
```

PR #1534, opened 2022-09-22
Allow for multiple schemata in filter condition

owenjonesuob:
Closes https://github.com/cynkra/dm/issues/1533.

(I opened that issue a moment ago, accidentally from a different account!)

### DIFF
--- a/R/db-helpers.R
+++ b/R/db-helpers.R
@@ -114,7 +114,57 @@ repair_table_names_for_db <- function(table_names, temporary, con, schema = NULL
   quote_ids(names, con_from_src_or_con(con), schema)
 }
 
-get_src_tbl_names <- function(src, schema = NULL, dbname = NULL) {
+
+
+find_name_clashes <- function(old, new) {
+
+  # Any entries in `new` with a higher count than in `old`
+  clashes <- names(purrr::keep(table(new), ~ .x > table(old)[.x]))
+
+  purrr::imap_chr(
+    purrr::set_names(clashes),
+    ~ glue::glue("* {paste0(dQuote(old[new == .x], q = FALSE), collapse = ', ')} => {dQuote(.y, q = FALSE)}")
+  )
+}
+
+
+make_local_names <- function(schema_names, table_names, tidy_names = FALSE) {
+
+  if (tidy_names) {
+
+    check_suggested("snakecase", use = TRUE)
+
+    schema_names_untidy <- schema_names
+    schema_names <- snakecase::to_snake_case(schema_names)
+
+
+    schema_clashes <- find_name_clashes(schema_names_untidy, schema_names)
+
+    if (length(schema_clashes) > 0)
+      stop(paste0(c("Forcing tidy schema names leads to name clashes:", schema_clashes), collapse = "\n"))
+
+
+    table_names_untidy <- table_names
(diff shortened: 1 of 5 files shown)


krlmlr:
Thanks. Can you please check the error at https://github.com/cynkra/dm/actions/runs/3107142929/jobs/5034892662#step:12:173 ?

owenjonesuob:
Apologies, that was my bad! The SQL query needed to change from

```sql
where schema_name = 'something'
```

to

```sql
where schema_name in ('something'[, ...])
```

This happens automatically when `schema` is a character vector, but when it's a sql object - e.g. the `dplyr::sql()` object returned by `dm:::schema_mariadb()` - we need to wrap it manually. The nicest way I could think to do that was to use `glue::glue_sql_collapse()`, but there may be some neater way!

krlmlr:
Thanks. To better understand this, a test would be helpful. Would you like to add one? Ideally this would fail on the current main branch.

R/db-helpers.R
```diff
@@ -146,11 +146,14 @@ get_src_tbl_names <- function(src, schema = NULL, dbname = NULL) {
...
   names_table %>%
-    filter(schema_name == !!schema) %>%
+    # sql expressions need to be wrapped in parentheses
+    filter(schema_name %in% !!(if (inherits(schema, "sql")) glue_sql_collapse(schema) else schema)) %>%
     collect() %>%
     # create remote names for the tables in the given schema (name is table_name; cannot be duplicated within a single schema)
-    mutate(remote_name = schema_if(schema_name, table_name, con, dbname)) %>%
-    select(-schema_name) %>%
+    transmute(
+      local_name = gsub(" ", "_", if (length(unique(schema_name)) == 1) table_name else paste(schema_name, table_name, sep = ".")),
```
krlmlr, line 154: What happens if we don't do this? I'd expect an error.

I think this should be opt-in, perhaps with a glue expression that refers to variables named `schema` and `table`?

krlmlr (commented):
Thanks, good catch. I wonder if we could usefully split the PR -- e.g., first add all tests (skipping failed tests), and then fix and unskip?

owenjonesuob:
Apologies, closed due to git-related clumsiness - I'll open a new PR in just a moment!

owenjonesuob:
Sorry it took me so long to get around to doing so - but I've finally added full testing for the new functionality included in this PR (and CI is now happy 😇)

krlmlr:
Thanks for the update. It might take us some time to review this.

@mgirlich: Would you like to take a look?

krlmlr (commented):
Thanks.

- A few (snapshot?) tests for `find_name_clashes()` and `make_local_names()` would help understand the logic
- Can we keep existing tests unchanged in this PR?
- The conversion to snake case (or to any other case, for that matter) is an interesting feature itself, but does it really have to be tightly coupled with `tidy_names`? The original issue seems to be about the inability to use multiple schemas.

owenjonesuob:
Thanks for the pointers! You're absolutely right, this is really two things in one go, so I've split them out into new separate PRs:
* #1789 - the bugfix for multiple schemas
* #1790 - possible new feature of name repairing

Hopefully this makes it easier to continue the two threads of discussion. I'll close this PR now, since the changes are now covered entirely by those two new PRs.

PR #1540, opened 2022-10-04
start work on cheatsheet vignette

maelle:
TODOS

- [x] figure out it looks good on Firefox only (I might need to update Chrome) -- locally
- [ ] figure out why it looks good in the browser but not when printing?!
- [ ] figure out system dependencies
- [ ] figure out a way to make the font loading work, speaking of these sytem dependencies
- [ ] maybe only serve the HTML version and count on users to print from there (especially as it'd be on a cynkra dot come subdomain hence the font should be served!)
- [ ] check all images are there


maelle:
Reg inserting SVGs inline, some care is needed to include the alternative text in the absence of an img tag. https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_In_HTML_Introduction#best_practices

vignettes/articles/cheatsheet.Rmd
```diff
@@ -0,0 +1,550 @@
...
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  echo = FALSE
+)
+```
+
+[PDF](/dev/cheatsheet.pdf)
```
maelle, line 18: to be fixed

vignettes/articles/cheatsheet.Rmd
```diff
@@ -0,0 +1,550 @@
...
+```{r, results='asis'}
+create_figure(
+  class = "right",
+  alt = "diagram showing the transformation from left two dm objects to right a single dm object represented as three tables in a blue frame",
+  src = "dm-dm.svg",
+  caption = "```r\ndm(dm1, dm2)\n```\n"
+)
+```        
+
+<hr class="d-none d-print-block"></hr>
```
maelle, line 117: taking advantage of bootstrap display classes :smile_cat:

maelle:
It seems pagedjs polyfill is having trouble loading the font. I don't remember this error from earlier this year but then I was not trying to use `pagedown::chrome_print()` :thinking: 

If I am able to get the PDF printing correctly from the menu, I'd go with serving the HTML version (building it via the pkgdown article, linking it from the navbar).

maelle:
On Chrome things look good as HTML, not good when printing (not sure why I don't see the right options)
On Firefox things look a bit shifted because it cannot load the font (expected, I don't use a CORS unblocker there) but the PDF would look exactly the same.
So I'm wondering whether this means I should install the font.
Or should the cheatsheet just use a free similar font instead.

maelle:
https://support.google.com/chrome/thread/3101500/chrome-print-dialogue-not-offering-fit-to-page-landscape-other-printing-options?hl=en

maelle:
that said on Chrome I see more printing options where trying to print any other page

maelle:
In summary, I am a bit stuck and I assume it is because of the font + Chrome.

PR #1646, opened 2022-10-20
feat: Improve error message for `dm_add_pk()` when the `columns` argument is missing

### LINKED ISSUE #1644: Better error message for `dm_add_pk()` when `columns` argument is missing
The current error message is a bit abstruse for a simple mistake like this.

``` r
library(dm)

x <- tibble(x = integer())

dm(x) %>%
  dm_add_pk(x)
#> Error in eval_select_indices(col_expr, colnames(table)): argument "col_expr" is missing, with no default
```

<sup>Created on 2022-10-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
Closes #1644

### DIFF
--- a/R/primary-keys.R
+++ b/R/primary-keys.R
@@ -47,9 +47,11 @@
 dm_add_pk <- function(dm, table, columns, ..., check = FALSE, force = FALSE) {
   check_dots_empty()
   check_not_zoomed(dm)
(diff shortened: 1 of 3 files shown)


krlmlr:
Thanks!

Issue #1642, opened 2022-10-19
Support `dm_add_pk(autoincrement = TRUE)`


krlmlr:
For #1386.

- [x] Argument to `dm_add_pk()`, for simple keys only
- [x] Storage in dm def (watch out for `dm_upgrade()`)
- [x] Creation in `copy_dm_to()` (targeting Postgres and SQL Server first)
- [ ] Learning in `dm_learn_from_db()` (targeting Postgres and SQL Server first)
- [x] `dm_paste()`

``` r
library(dm)
#> 
#> Attaching package: 'dm'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(tidyverse)
library(DBI)

# Setup
con <- dbConnect(RSQLite::SQLite())

dbExecute(con, "CREATE TABLE x (x_id INTEGER PRIMARY KEY AUTOINCREMENT, x_data TEXT) STRICT")
#> [1] 0
dbExecute(con, "CREATE TABLE y (y_id INTEGER PRIMARY KEY AUTOINCREMENT, x_id INTEGER, y_data TEXT, FOREIGN KEY(x_id) REFERENCES x(x_id)) STRICT")
#> [1] 0

# Data model
dm <-
  dm_from_con(con, learn_keys = FALSE) %>%
  dm_add_pk(x, x_id) %>%
  dm_add_pk(y, y_id) %>%
  dm_add_fk(y, x_id, x, x_id)

dm %>%
  dm_draw()
```

![](https://i.imgur.com/Kbo3RHP.png)

``` r

# Desired syntax:
try({
  x <- tibble::tibble(
    x_id = integer(0),
    x_data = character(0),
  )
  y <- tibble::tibble(
    y_id = integer(0),
    x_id = integer(0),
    y_data = character(0),
  )
  dm(
    x,
    y,
  ) %>%
    dm_add_pk(x, x_id, autoincrement = TRUE) %>%
    dm_add_pk(y, y_id, autoincrement = TRUE) %>%
    dm_add_fk(y, x_id, x)
})
#> Error in dm_add_pk(., y, y_id, autoincrement = TRUE) : 
#>   `...` must be empty.
#> ✖ Problematic argument:
#> • autoincrement = TRUE

# Hint: watch out for dm_upgrade()

# Support only simple PKs with autoincrement
```

<sup>Created on 2022-10-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

IndrajeetPatil:
[...]

PR #1654, opened 2022-10-26
Support `dm_add_pk(autoincrement = TRUE)`

IndrajeetPatil:
Related to #1642


R/validate.R
```diff
@@ -64,7 +64,7 @@ dm_validate <- function(x) {
     select(table = ref_table, column = ref_column) %>%
     check_colnames(dm_col_names, "Parent key")
 
-  stopifnot(lengths(def$pks) %in% 0:1)
+  stopifnot(lengths(def$pks) %in% 0:2)
```
krlmlr, line 67: I don't see why this is necessary. Is the check a mistake?

krlmlr:
It might be easier, in the beginning, to keep the behavior of `dm_get_all_pks()` unchanged -- to reduce the snapshot changes in the first iteration.

R/validate.R
IndrajeetPatil: The definition of `new_pk()` is the following:

```r
new_pk <- function(column = list(), autoincrement = logical()) {
  stopifnot(is.list(column), is.logical(autoincrement))
  tibble(column = column, autoincrement = autoincrement)
}
```

This means that the length (`length(new_pk())`) can be `2L`.

Or are you saying that this entire check itself is unnecessary?

krlmlr: I guess the check should be `NROW(new_pk()) %in% 0:1` .

IndrajeetPatil:
`test-upgrade.R` continues to fail because `dm` objects stored in `.rds` files (`tests/testthat/dm/`) don't have `autoincrement` column. So the saved column type (`pks <list<tibble[,1]>>`) differs from upgraded column type (`pks <list<tibble[,2]>>`)

R/primary-keys.R
```diff
@@ -55,18 +58,29 @@ dm_add_pk <- function(dm, table, columns, ..., check = FALSE, force = FALSE) {
   col_expr <- enexpr(columns)
   col_name <- names(eval_select_indices(col_expr, colnames(table)))
 
+  # TODO: Should we support autoincrement for composite primary keys?
```
krlmlr, line 61: We probably won't, no need for this TODO.

R/validate.R
```diff
@@ -64,8 +64,6 @@ dm_validate <- function(x) {
     select(table = ref_table, column = ref_column) %>%
     check_colnames(dm_col_names, "Parent key")
 
-  stopifnot(lengths(def$pks) %in% 0:1)
-
```
krlmlr, line 68: This should be replaced with a correct test, not removed.

krlmlr (commented):
Please edit `dm_upgrade()` and bump the version in `new_dm3()` .

krlmlr:
What is blocking progress here?

IndrajeetPatil:
Struggling to figure out how to create autoincrementing PKs in `copy_dm_to()`. We can discuss during the one-on-one call if I can't figure it out by then.

tests/testthat/_snaps/rows-dm.md
```diff
@@ -94,6 +94,9 @@
       dm_copy %>% dm_rows_update(dm_update_copy) %>% pull_tbl(tf_2) %>% arrange_all()
     Message
       Result is returned as a dm object with lazy tables. Use `in_place = FALSE` to mute this message, or `in_place = TRUE` to write to the underlying tables.
+    Condition
+      Warning:
+      `x` must be a one- or two-column data frame in `deframe()`.
```
IndrajeetPatil, line 99: Note to self: Track down where these warnings are coming from and get rid of them.

IndrajeetPatil:
Minor progress update:

``` r
library(dm)

con <- DBI::dbConnect(RSQLite::SQLite())

x <- tibble::tibble(
  x_id = integer(0),
  x_data = character(0),
)
y <- tibble::tibble(
  y_id = integer(0),
  x_id = integer(0),
  y_data = character(0),
)

dm <- dm(x, y,) %>%
  dm_add_pk(x, x_id, autoincrement = TRUE) %>%
  dm_add_pk(y, y_id, autoincrement = TRUE) %>%
  dm_add_fk(y, x_id, x)

df <- dm:::build_copy_queries(con, dm)

df$sql_table
#> <SQL> CREATE TEMPORARY TABLE x (
#>   `x_id` INT,
#>   `x_data` TEXT,
#>   AUTOINCREMENT PRIMARY KEY (`x_id`)
#> )
#> <SQL> CREATE TEMPORARY TABLE y (
#>   `y_id` INT,
#>   `x_id` INT,
#>   `y_data` TEXT,
#>   AUTOINCREMENT PRIMARY KEY (`y_id`),
#>   FOREIGN KEY (`x_id`) REFERENCES x (`x_id`)
#> )
```

<sup>Created on 2022-11-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

krlmlr:
Thanks, looks good! The SQL Server and PostgreSQL syntax look a bit different. For now, PostgreSQL is most important, SQL Server can wait.

You can `brew install postgresql@14` and `brew services start postgresql@14` (and `stop` when done). See if you can connect by running `psql` on the shell.

TSchiefer:
Have to have a closer look at why the smoke test failed, but in a nutshell:
It seems to work alright now for Postgres. If you use `copy_dm_to()` to copy the test dm `dm_for_autoinc_1` to PG it creates the expected result.
I (temporarily) sacrificed the progress bar for populating the tables, because I needed to use a `for` loop in order to update the child tables of tables with an autoincrement PK before populating them.
I now use `rows_append()` for the tables with an AI-PK.
Still needs quite a bit of work on some details.

R/build_copy_queries.R
```diff
@@ -22,23 +22,83 @@ build_copy_queries <- function(dest, dm, set_key_constraints = TRUE, temporary =
...
+    df_col_types
   }
 
   col_defs <-
     dm %>%
     src_tbls_impl() %>%
     set_names() %>%
     map_dfr(get_sql_col_types, .id = "name") %>%
-    mutate(col_def = glue("{DBI::dbQuoteIdentifier(con, col)} {type}")) %>%
+    mutate(col_def = glue("{DBI::dbQuoteIdentifier(con, col)} {type} {autoincrement_attribute}")) %>%
```
krlmlr, line 98: To get rid of the `trimws()` below, change definition of `autoincrement_attribute` and

```suggestion
    mutate(col_def = glue("{DBI::dbQuoteIdentifier(con, col)} {type}{autoincrement_attribute}")) %>%
```

?

R/db-interface.R
```diff
@@ -218,11 +232,25 @@ copy_dm_to <- function(dest, dm, ...,
     top_level_fun = "copy_dm_to"
   )
 
+
   # populate tables
-  pwalk(
-    queries[c("name", "remote_name")],
-    ticker_populate(~ db_append_table(dest_con, .y, dm[[.x]], progress))
-  )
+  for (i in seq_along(queries$name)) {
```
krlmlr, line 237: Good catch, but I'd prefer ignoring this problem for now. It's fine if we can't use `copy_dm_to()` with a dm with data and autoincrement keys. We can resort to `rows_append()` once we have that, or figure out something even nicer.

R/paste.R
```diff
@@ -168,15 +168,20 @@ dm_paste_select <- function(dm) {
 dm_paste_pks <- function(dm) {
   dm %>%
     dm_get_all_pks_impl() %>%
-    mutate(code = glue("dm::dm_add_pk({tick_if_needed(table)}, {deparse_keys(pk_col)})")) %>%
+    mutate(
+      code = case_when(
+        !is.na(autoincrement) & autoincrement ~ glue("dm::dm_add_pk({tick_if_needed(table)}, {deparse_keys(pk_col)}, autoincrement = TRUE)"),
```
krlmlr, line 173: How can `autoincrement` be `NA` here?

R/primary-keys.R
```diff
@@ -16,12 +16,13 @@
 #' @param force Boolean, if `FALSE` (default), an error will be thrown if there is already a primary key
 #'   set for this table.
 #'   If `TRUE`, a potential old `pk` is deleted before setting a new one.
+#' @param autoincrement If `TRUE`, a sequence of integers (from 1 to number of
+#'   rows) will be generated in column specified in `columns`.
```
krlmlr, line 20: Can we also add an "experimental" badge for this argument?

```suggestion
#' @param autoincrement If `TRUE`, the  column specified in `columns` will be populated
#'   automatically with a sequence of integers.
```

R/upgrade.R
```diff
@@ -39,5 +39,14 @@ dm_upgrade <- function(dm, quiet) {
     dm <- new_dm3(def, zoomed = is_zoomed(dm))
   }
 
+  if (version < 4L) {
+    if (!quiet) {
+      message("Upgrading dm object created with dm <= 1.0.4.") # TODO: Check that package version is correct
+    }
+    def <- unclass(dm)$def
+    def$uuid <- vec_new_uuid_along(def$table)
```
krlmlr, line 47: We need to adapt that code to make sure we create storage for the autoincrement column here.

R/validate.R
krlmlr: This is still an issue. Do we need `map_int(vctrs::vec_size)` ?

TSchiefer:
I will address your comments in a follow-up session. Now I worked on making dm_rows_append() work for PKs on autoincrement columns and the tables referencing them. It seems to work both for SQL Server and PG, in case the tables are created properly on the DBs.
The snapshot should be supplied by GHA, I need to review that result.

I reverted the commits relating to copy_dm_to().

TSchiefer:
I noticed that due to the new column `autoincrement` for the `pks`-column in `def`, the tests for learning (`dm_learn_from_db()` or `dm_from_con()`) from a remote src fail now, cause -- at least for PG -- we don't have the info available, if a key is an AI key, and this leads to `NA` values for `autoincrement`, e.g.:
https://github.com/cynkra/dm/actions/runs/3676951573/jobs/6218444459#step:7:146
https://github.com/cynkra/dm/actions/runs/3676951573/jobs/6218444459#step:7:225

Shall we just replace `NA` with `FALSE`, or shall we try to figure out the nature of the PK already in this PR?

R/validate.R
TSchiefer: We want to test here if we have maximally 1 PK for each table, right?
I am gonna go with your suggestion here, `all(map_int(def$pks, vctrs::vec_size) %in% 0:1)` seems to work.

R/paste.R
TSchiefer: Currently this can happen when learning a dm from a remote data source.

R/build_copy_queries.R
TSchiefer: Is this [a446754](https://github.com/cynkra/dm/pull/1654/commits/a446754a39ff7e2339c61127f58fce23c0e5062e) what you meant here?

R/primary-keys.R
TSchiefer: done in [0086bf3](https://github.com/cynkra/dm/pull/1654/commits/0086bf3e6f3155e8c4c4a504699d24397d9f5019)

R/upgrade.R
TSchiefer: do we need to add a new version 5? the code could be:
```r
def <- unclass(dm)$def
    def$pks <- map(def$pks, mutate, autoincrement = FALSE) %>%
      vctrs::as_list_of(new_pk())
    dm <- new_dm3(def, zoomed = is_zoomed(dm))
```

krlmlr: We only need to update the code that upgrades from version 3 to version 4.

krlmlr:
This PR is already large, we can tackle remaining TODOs in separate PRs.

R/upgrade.R
TSchiefer: done in [0910ce7](https://github.com/cynkra/dm/pull/1654/commits/0910ce7ecb5745ef7d6577c42b951a730459a5d2)

TSchiefer:
I tweaked expect_equivalent_dm by adding an ignore_autoincrement arg in [4c3a95e](https://github.com/cynkra/dm/pull/1654/commits/4c3a95e0e271fc29d8530c6923c186616903a56d).
We can revert that once we can learn autoincrement PKs from remote sources.

TSchiefer:
I would suggest among the TODOs should be:
1. learn AI PKs from remote sources
2. copy_dm_to() should eventually be able to create and fill data when one or more tables with AI PKs are to be copied

krlmlr:
Is this PR ready for review? Happy to!

R/primary-keys.R
```diff
@@ -16,12 +16,15 @@
 #' @param force Boolean, if `FALSE` (default), an error will be thrown if there is already a primary key
 #'   set for this table.
 #'   If `TRUE`, a potential old `pk` is deleted before setting a new one.
+#' @param autoincrement
+#'   `r lifecycle::badge("experimental")`
+#'   If `TRUE`, the  column specified in `columns` will be populated
+#'   automatically with a sequence of integers.
```
TSchiefer, line 22: This doesn't work for local dm's yet. MAybe we should mention that in the documentation. Also this should be another TODO in my opinion.

TSchiefer:
Now also testing dm_rows_append() with AI PKs on SQL Server ([4ef7b1d](https://github.com/cynkra/dm/pull/1654/commits/4ef7b1dbf66f2d51f80fc1c46482f76b27510019))

TSchiefer:
It also works for SQLite, so the test is now also conducted there. For MariaDB `returning` is not supported. It might be possible to work around here, but it promises to be a bit fiddly and should IMO be done in a future PR.

krlmlr:
Implemented in #1698.

PR #1659, opened 2022-11-07
chore: Make sure `{dm}` passes "noSuggests" workflow

IndrajeetPatil:
Done using https://github.com/krlmlr/actions-sync/pull/35

Also see: https://developer.r-project.org/blosxom.cgi/R-devel/NEWS/2022/11/14

cf. https://github.com/krlmlr/actions-sync/issues/22


vignettes/setup/setup.R
```diff
@@ -1,6 +1,12 @@
+are_suggested_pkgs_needed_for_vignettes_available <- function() {
+  pkgs_needed_for_vignettes <- c("tidyverse")
+  all(purrr::map_lgl(pkgs_needed_for_vignettes, rlang::is_installed))
+}
```
IndrajeetPatil, line 4: I know this whole function can be simplified just by using `rlang::is_installed("tidyverse")`, but I am trying to be future-proof here. If `{dm}` vignettes need any additional deps in the future, we can simply add those to the `pkgs_needed_for_vignettes` vector.

.github/workflows/R-CMD-check-hard.yaml
```diff
@@ -0,0 +1,52 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+#
+# NOTE: This workflow only directly installs "hard" dependencies, i.e. Depends,
+# Imports, and LinkingTo dependencies. Notably, Suggests dependencies are never
+# installed, with the exception of testthat, knitr, and rmarkdown. The cache is
+# never used to avoid accidentally restoring a cache containing a suggested
+# dependency.
```
IndrajeetPatil, line 8: This workflow can either be included in this PR, or removed from the current PR and added directly to the actions-sync repo.

krlmlr:
Thanks!

jangorecki:
Is it expected that check of {dm} should be passing simplified check workflow (no vignettes) without needing suggested dependencies to be installed?

PR #1660, opened 2022-11-07
Harmonize vignette titles and entries

IndrajeetPatil:
To avoid warnings while building vignettes:

```r
Warning: Warning: The vignette title specified in \VignetteIndexEntry{} is different from the title in the YAML metadata. The former is "First read: Getting started with dm", and the latter is "Getting started with dm". If that is intentional, you may set options(rmarkdown.html_vignette.check_title = FALSE) to suppress this check.

Warning: Warning: The vignette title specified in \VignetteIndexEntry{} is different from the title in the YAML metadata. The former is "How to: Create a dm object from a database", and the latter is "Create a dm object from a database". If that is intentional, you may set options(rmarkdown.html_vignette.check_title = FALSE) to suppress this check.

...
```

The other option is to build vignettes with the following option: `options(rmarkdown.html_vignette.check_title = FALSE)`.

Not sure which one we should prefer. I personally prefer having the same, detailed titles on the pkgdown website as on CRAN.

### DIFF
--- a/vignettes/dm.Rmd
+++ b/vignettes/dm.Rmd
@@ -1,5 +1,5 @@
 ---
-title: "Getting started with dm"
+title: "First read: Getting started with dm"
 date: "`r Sys.Date()`"
 output: rmarkdown::html_vignette
 vignette: >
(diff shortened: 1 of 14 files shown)


krlmlr:
Can you please push this to a branch whose name starts with `docs` so that we get a pkgdown preview?

IndrajeetPatil:
Thanks. Done in https://github.com/cynkra/dm/pull/1661

PR #1661, opened 2022-11-07
docs: Remove warnings about differences between vignette titles and entries

IndrajeetPatil:
To avoid warnings while building vignettes:

```r
Warning: Warning: The vignette title specified in \VignetteIndexEntry{} is different from the title in the YAML metadata. The former is "First read: Getting started with dm", and the latter is "Getting started with dm". If that is intentional, you may set options(rmarkdown.html_vignette.check_title = FALSE) to suppress this check.

Warning: Warning: The vignette title specified in \VignetteIndexEntry{} is different from the title in the YAML metadata. The former is "How to: Create a dm object from a database", and the latter is "Create a dm object from a database". If that is intentional, you may set options(rmarkdown.html_vignette.check_title = FALSE) to suppress this check.

...
```

The other option is to build vignettes with the following option: `options(rmarkdown.html_vignette.check_title = FALSE)`.

Not sure which one we should prefer. I personally prefer having the same, detailed titles on the pkgdown website as on CRAN.

### DIFF
--- a/.github/workflows/pkgdown-build/action.yml
+++ b/.github/workflows/pkgdown-build/action.yml
@@ -5,5 +5,6 @@ runs:
   steps:
     - name: Build site
       run: |
+        options(rmarkdown.html_vignette.check_title = FALSE)
         pkgdown::build_site()
       shell: Rscript {0}
--- a/vignettes/dm.Rmd
+++ b/vignettes/dm.Rmd
@@ -1,5 +1,5 @@
 ---
-title: "Getting started with dm"
+title: "First read: Getting started with dm"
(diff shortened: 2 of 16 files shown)


krlmlr:
That's what the title looks now:

Screen Shot 2022-11-07 at 11 26 01

The "How to: " needs to go, I don't mind the warnings. Making the titles consistent otherwise could be a useful result of this PR.

IndrajeetPatil:
> I don't mind the warnings. Making the titles consistent otherwise could be a useful result of this PR.

Okay, in that case, there is not much to do here. The titles are already quite consistent, so not much needs to be changed there.

PR #1686, opened 2022-12-12
chore: Use correctly typed missing value for lists

DavisVaughan:
This PR makes your package compatible with the next version of dplyr:

- `if_else()` now uses vctrs, which generally makes it more permissive when there are varying types, but it also means that the inputs must always be vectors. In dm's case, `NULL` was being used as the missing type for lists, when really `list(NULL)` needs to be used for vctrs

In dev dplyr, you could also just use `NA` now but that won't work with CRAN dplyr.

We plan to submit dplyr 1.1.0 on January 27th.

This should be compatible with both dev and CRAN dplyr. It would help us out if you could go ahead and send a patch version of your package in ahead of time! Thanks!

---

Also, I think if you try and check this against dev dplyr right now then you'll see an additional failure related to `left_join()` and unused arguments. I think we are going to fix that in dplyr so it should go away.

There are also some local failures related to join snapshots changing because of the `multiple = "all"` warning that is now thrown when a join results in multiple matches. When 1.1.0 is released you can just update the snapshots or set `multiple = "all"` explicitly

### DIFF
--- a/R/gui-gui_global.R
+++ b/R/gui-gui_global.R
@@ -75,17 +75,17 @@ reactable_column <- function(data, table_name) {
               dplyr::if_else(
                 data$is_pk[index],
                 list(htmltools::span(style = "margin-right: 10px;", title = "Primary key", shiny::icon(verify_fa = FALSE, "key"))),
-                NULL
+                list(NULL)
               ),
               dplyr::if_else(
                 data$is_child_fk[index],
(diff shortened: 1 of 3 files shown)


krlmlr:
Thanks!

krlmlr:
Thanks Davis! We're running dm continuously against the dev version of dplyr, see, e.g., https://github.com/cynkra/dm/actions/runs/3745241278 . I'll take a look later today to make sure we're good when dplyr 1.1.0 is submitted.

PR #1689, opened 2022-12-19
feat: adapt dm structure to support autoincrement PKs

TSchiefer:
subset taken from https://github.com/cynkra/dm/pull/1654


TSchiefer:
@krlmlr Do you have an idea, why in the validate-test, test-update.R is suddenly failing? https://github.com/cynkra/dm/actions/runs/3734991716/jobs/6337699310
In the original PR this didn't happen, and I cannot tell the origin of the error, as locally it works alright.

krlmlr:
The validate run sets an environment variable and/or tweaks the code. Can you recreate the failure locally by applying the same steps as defined in the GitHub Actions YAML file?

TSchiefer:
Problem was, that validation took place during the updating process (e.g. during lifting from version 1 -> 2), and then the validation wasn't successful, since the blueprint obviously wasn't matched.

TSchiefer:
Just wondering: it almost looks as if the tests are running twice without any changes to them? https://github.com/cynkra/dm/blob/main/.github/workflows/validate.yaml#L105-L114
Am I missing something?

krlmlr (commented):
We also need an upgrade test from the most recent version to this version. It would best fit this PR, I believe.

TSchiefer:
It kind of checks already by comparing to the boilerplate `new_dm_def()`: https://github.com/cynkra/dm/blob/1642-f-autoincrement-data-structure/R/validate.R#L29-L34
Should there be a more explicit check?

krlmlr:
Thanks! Can we get the rest of the other PR to CRAN, today, before it closes? 🚄🚄🚄

PR #1694, opened 2022-12-20
remove duplicate step in validate GHA

TSchiefer:
tests were run twice in a row in the same manner, if I am not mistaken.


krlmlr:
Thanks!

PR #1715, opened 2022-12-27
chore: `dm_unpack_tbl()` sets PK before FK

TSchiefer:
Otherwise in the future, there will be an additional unique key, which is the same as the then added PK.


TSchiefer:
"Unfortunately" this didn't make a difference in the past regarding the dm that was created. So it's not possible to add a test that reflects this change in behaviour.

krlmlr:
Thanks!

PR #1716, opened 2022-12-27
feat: New `dm_add_uk()`, `dm_rm_uk()` and `dm_get_all_uks()` functions for explicit support of unique keys

### LINKED ISSUE #622: Stronger support for unique keys
With #561 (closing #402) already some support for the concept of multiple primary keys was introduced.
After this PR FKs to a table can refer to different columns. But there will always be just 1 "official" PK at a time. My suggestion is to make it possible to have several PKs at a time which have "equal rights".

Things that would need to be adapted:
1. `dm_add_pk()` (`dm_rm_pk()` may already be ok)
1. `dm_add_fk()`
1. Drawing a dm
1. print dm to console (`print.dm()`, `glimpse.dm()` (after #605))
1. potentially setting key relations when copying a dm to a DB (might already work though for the supported DBMS)

TSchiefer:
unique key = UK.
Can possibly also be called "weak keys".

- There can be an unlimited number of UKs (also updated doc of PK to clarify that only 1 PK can exist per table)
- adding a foreign key whose parent table doesn't have a PK, or whose PK doesn't match the fiven PK-cols results in creating a UK in the parent table
- UKs are listed in `print.dm()`, should they exist
- UKs are added in `dm_paste()` in case option `keys` is chosen

This PR is already based on https://github.com/cynkra/dm/pull/1715 before its merge (1st commit here) to avoid having to deal with ever changing snapshots.

closes #622

### DIFF
--- a/R/dm.R
+++ b/R/dm.R
@@ -604,7 +604,7 @@ tbl_def_impl <- function(def, idx, keyed) {
     pk <- NULL
   }
 
-  uk_def <- def$uks[[idx]]
+  uks <- def$uks[[idx]]
 
   fks_in_def <-
     def$fks[[idx]] %>%
@@ -630,6 +630,7 @@ tbl_def_impl <- function(def, idx, keyed) {
   new_keyed_tbl(
     data,
     pk = pk,
+    uks = uks,
     fks_in = fks_in,
     fks_out = fks_out,
     uuid = def$uuid[[idx]]
--- a/R/dm.R
+++ b/R/dm.R
@@ -384,8 +384,6 @@ show_dm <- function(x) {
   cat_line("Tables: ", commas(tick(def$table)))
   cat_line("Columns: ", def_get_n_columns(def))
   cat_line("Primary keys: ", def_get_n_pks(def))
-  n_uks <- def_get_n_uks(def)
-  if (n_uks > 0) cat_line("Unique keys: ", n_uks)
   cat_line("Foreign keys: ", def_get_n_fks(def))
 
   filters <- dm_get_filters_impl(x)
@@ -420,10 +418,6 @@ def_get_n_columns <- function(def) {
 
 def_get_n_pks <- function(def) {
   sum(map_int(def$pks, vec_size))
-}
-
-def_get_n_uks <- function(def) {
-  sum(map_int(def$uks, vec_size))
 }
 
 def_get_n_fks <- function(def) {
--- a/R/dm.R
+++ b/R/dm.R
@@ -137,16 +137,17 @@ new_keyed_dm_def <- function(tables = list()) {
   unclassed_tables <- map(tables, unclass_keyed_tbl)
 
   pks_df <- pks_df_from_keys_info(tables[is_keyed])
+  uks_df <- uks_df_from_keys_info(tables[is_keyed])
   fks_df <- fks_df_from_keys_info(tables[is_keyed])
 
-  new_dm_def(unclassed_tables, pks_df, fks_df)
+  new_dm_def(unclassed_tables, pks_df, uks_df, fks_df)
 }
 
 
 new_dm_def <- function(tables = list(),
                        pks_df = tibble(table = character(), pks = list()),
-                       fks_df = tibble(table = character(), fks = list()),
-                       uks_df = tibble(table = character(), uks = list())) {
+                       uks_df = tibble(table = character(), uks = list()),
+                       fks_df = tibble(table = character(), fks = list())) {
   # Legacy
   data <- unname(tables)
   table <- names2(tables)
@@ -602,6 +603,8 @@ tbl_def_impl <- function(def, idx, keyed) {
   } else {
     pk <- NULL
   }
+
+  uk_def <- def$uks[[idx]]
 
   fks_in_def <-
     def$fks[[idx]] %>%
--- a/R/examine-constraints.R
+++ b/R/examine-constraints.R
@@ -120,7 +120,7 @@ kind_to_long <- function(kind) {
 
 check_pk_constraints <- function(dm, progress = NA, top_level_fun = NULL) {
   pks <- bind_rows(
-    list(PK = dm_get_all_pks_impl(dm), UK = dm_get_all_uks_impl(dm) %>% rename(pk_col = uk_col)),
+    list(PK = dm_get_all_pks_impl(dm), UK = dm_get_all_uks_impl(dm) %>% rename(pk_col = uk_col) %>% select(-kind)),
     .id = "kind"
   ) %>%
     distinct(table, pk_col, .keep_all = TRUE)
--- a/R/foreign-keys.R
+++ b/R/foreign-keys.R
@@ -274,7 +274,7 @@ dm_get_all_fks_def_impl <- function(def, parent_table = NULL, ignore_on_delete =
   if (!is.null(parent_table)) {
     idx <- match(parent_table, def_sub$parent_table)
     if (anyNA(idx)) {
-      abort(paste0("Table not in dm object: ", parent_table[which(is.na(idx))[[1]]]))
+      abort_table_not_in_dm(parent_table[which(is.na(idx))], def$table)
     }
     def_sub <- def_sub[idx, ]
   }
--- a/R/foreign-keys.R
+++ b/R/foreign-keys.R
@@ -3,9 +3,8 @@
 #' @description
 #' `dm_add_fk()` marks the specified `columns` as the foreign key of table `table` with
 #' respect to a key of table `ref_table`.
-#' Usually the referenced columns are a primary key in `ref_table`,
-#' it is also possible to specify other columns via the `ref_columns` argument.
-#' In this case `ref_columns` will be added as a new unique key to the referenced table.
+#' Usually the referenced columns are a primary key in `ref_table`.
+#' However, it is also possible to specify other columns via the `ref_columns` argument.
 #' If `check == TRUE`, then it will first check if the values in `columns` are a subset
 #' of the values of the key in table `ref_table`.
 #'
@@ -33,8 +32,12 @@
 #'
 #' @details
 #' It is possible that a foreign key (FK) is pointing to columns that are neither primary (PK) nor
-#' unique keys (UK). This can happen, when the PK or UK is removed ([`dm_rm_pk()`]/[`dm_rm_uk()`])
-#' without first removing the associated FKs.
+#' explicit unique keys (UK).
+#' This can happen
+#'   1. when a FK is added without a corresponding PK or UK being present in the parent table
+#'   1. when the PK or UK is removed ([`dm_rm_pk()`]/[`dm_rm_uk()`]) without first removing the associated FKs.
+#'
+#' These columns are then a so-called "implicit unique key" of the referenced table and can be listed via [`dm_get_all_uks()`].
 #'
 #' @rdname dm_add_fk
 #'
--- a/R/foreign-keys.R
+++ b/R/foreign-keys.R
@@ -93,7 +93,7 @@ dm_add_fk <- function(dm, table, columns, ref_table, ref_columns = NULL, ...,
     # check if either a PK or UK already matches ref_col_name
     all_keys <- dm_get_all_pks_impl(dm, ref_table_name) %>%
       rename(uk_col = pk_col) %>%
-      bind_rows(dm_get_all_uks_impl(dm, ref_table_name))
+      bind_rows(dm_get_all_uks_def_impl(dm_get_def(dm), ref_table_name))
     # setequal() could also be used for matching, but IMHO the order should matter
     matches_keys <- map_lgl(all_keys$uk_col, identical, ref_col_name)
     if (!any(matches_keys)) {
--- a/R/foreign-keys.R
+++ b/R/foreign-keys.R
@@ -5,6 +5,7 @@
 #' respect to a key of table `ref_table`.
 #' Usually the referenced columns are a primary key in `ref_table`,
 #' it is also possible to specify other columns via the `ref_columns` argument.
+#' In this case `ref_columns` will be added as a new unique key to the referenced table.
 #' If `check == TRUE`, then it will first check if the values in `columns` are a subset
 #' of the values of the key in table `ref_table`.
 #'
@@ -29,6 +30,11 @@
 #'   and might be considered by [dm_rows_delete()] in a future version.
 #'
 #' @family foreign key functions
+#'
+#' @details
+#' It is possible that a foreign key (FK) is pointing to columns that are neither primary (PK) nor
+#' unique keys (UK). This can happen, when the PK or UK is removed ([`dm_rm_pk()`]/[`dm_rm_uk()`])
+#' without first removing the associated FKs.
 #'
 #' @rdname dm_add_fk
 #'
--- a/R/foreign-keys.R
+++ b/R/foreign-keys.R
@@ -90,20 +90,6 @@ dm_add_fk <- function(dm, table, columns, ref_table, ref_columns = NULL, ...,
     ref_col_name <- get_key_cols(ref_key)
   } else {
     ref_col_name <- names(eval_select_indices(ref_col_expr, colnames(ref_table_obj)))
-    # check if either a PK or UK already matches ref_col_name
-    all_keys <- dm_get_all_pks_impl(dm, ref_table_name) %>%
-      rename(uk_col = pk_col) %>%
-      bind_rows(dm_get_all_uks_def_impl(dm_get_def(dm), ref_table_name))
-    # setequal() could also be used for matching, but IMHO the order should matter
-    matches_keys <- map_lgl(all_keys$uk_col, identical, ref_col_name)
-    if (!any(matches_keys)) {
-      if (check) {
-        if (!is_unique_key_se(ref_table_obj, ref_col_name)$unique) {
-          abort_not_unique_key(ref_table_name, ref_col_name)
-        }
-      }
-      dm <- dm_add_uk_impl(dm, ref_table_name, ref_col_name)
-    }
   }
 
   # FIXME: COMPOUND:: Clean check with proper error message
--- a/R/paste.R
+++ b/R/paste.R
@@ -184,7 +184,8 @@ dm_paste_pks <- function(dm) {
 
 dm_paste_uks <- function(dm) {
   dm %>%
-    dm_get_all_uks_impl() %>%
+    dm_get_def() %>%
+    dm_get_all_uks_def_impl() %>%
     mutate(code = glue("dm::dm_add_uk({tick_if_needed(table)}, {deparse_keys(uk_col)})")) %>%
     pull()
 }
--- a/R/primary-keys.R
+++ b/R/primary-keys.R
@@ -203,7 +203,7 @@ dm_get_all_pks_def_impl <- function(def, table = NULL) {
   if (!is.null(table)) {
     idx <- match(table, def_sub$table)
     if (anyNA(idx)) {
-      abort(paste0("Table not in dm object: ", parent_table[which(is.na(idx))[[1]]]))
+      abort_table_not_in_dm(table[which(is.na(idx))], def$table)
     }
     def_sub <- def_sub[match(table, def_sub$table), ]
   }
--- a/R/primary-keys.R
+++ b/R/primary-keys.R
@@ -233,9 +233,7 @@ dm_get_all_pks_def_impl <- function(def, table = NULL) {
 #' @param columns Table columns, unquoted.
 #'   To refer to a compound key, use `c(col1, col2)`.
 #'   Pass `NULL` (the default) to remove all matching keys.
-#' @param fail_fk
-#'   Boolean: if `TRUE` (default), will throw an error
-#'   if there are foreign keys addressing the primary key that is to be removed.
+#' @param fail_fk `r lifecycle::badge("deprecated")`
 #'
 #' @family primary key functions
 #'
--- a/R/primary-keys.R
+++ b/R/primary-keys.R
@@ -260,7 +260,11 @@ dm_rm_pk_ <- function(dm, table, columns, ..., rm_referencing_fks = NULL) {
   check_not_zoomed(dm)
 
   if (!is.null(rm_referencing_fks)) {
-    deprecate_soft("0.2.1", "dm::dm_rm_pk(rm_referencing_fks = )")
+    deprecate_soft(
+      "0.2.1",
+      "dm::dm_rm_pk(rm_referencing_fks = )",
+      details = "When removing a primary key, potential associated foreign keys will be pointing at an implicit unique key."
+    )
   }
 
   table_name <- dm_tbl_name_null(dm, {{ table }})
--- a/R/primary-keys.R
+++ b/R/primary-keys.R
@@ -247,37 +247,32 @@ dm_get_all_pks_def_impl <- function(def, table = NULL) {
 #'   dm_rm_fk(flights, origin, airports) %>%
 #'   dm_rm_pk(airports, fail_fk = FALSE) %>%
 #'   dm_draw()
-dm_rm_pk <- function(dm, table = NULL, columns = NULL, ..., fail_fk = TRUE) {
-  if (missing(fail_fk)) {
-    fail_fk <- NULL
+dm_rm_pk <- function(dm, table = NULL, columns = NULL, ..., fail_fk = NULL) {
+  if (!is.null(fail_fk)) {
+    lifecycle::deprecate_soft(
+      "1.0.4",
+      "dm_rm_pk(fail_fk =)",
+      details = "When removing a primary key, potential associated foreign keys will be pointing at an implicit unique key."
+    )
   }
-  dm_rm_pk_(dm, {{ table }}, {{ columns }}, ..., fail_fk = fail_fk)
+  dm_rm_pk_(dm, {{ table }}, {{ columns }}, ...)
 }
 
-dm_rm_pk_ <- function(dm, table, columns, ..., rm_referencing_fks = NULL, fail_fk = NULL) {
+dm_rm_pk_ <- function(dm, table, columns, ..., rm_referencing_fks = NULL) {
   check_dots_empty()
   check_not_zoomed(dm)
 
   if (!is.null(rm_referencing_fks)) {
-    deprecate_soft("0.2.1", "dm::dm_rm_pk(rm_referencing_fks = )", "dm::dm_rm_pk(fail_fk = )",
-      details = "Note the different semantics: `fail_fk = FALSE` roughly corresponds to `rm_referencing_fks = TRUE`, but foreign keys are no longer removed."
-    )
-    if (is.null(fail_fk)) {
-      fail_fk <- !rm_referencing_fks
-    }
-  }
-
-  if (is.null(fail_fk)) {
-    fail_fk <- TRUE
+    deprecate_soft("0.2.1", "dm::dm_rm_pk(rm_referencing_fks = )")
   }
 
   table_name <- dm_tbl_name_null(dm, {{ table }})
   columns <- enexpr(columns)
 
-  dm_rm_pk_impl(dm, table_name, columns, fail_fk)
+  dm_rm_pk_impl(dm, table_name, columns)
 }
 
-dm_rm_pk_impl <- function(dm, table_name, columns, fail_fk) {
+dm_rm_pk_impl <- function(dm, table_name, columns) {
   def <- dm_get_def(dm)
 
   if (is.null(table_name)) {
@@ -306,13 +301,6 @@ dm_rm_pk_impl <- function(dm, table_name, columns, fail_fk) {
   if (length(i) == 0 && dm_is_strict_keys(dm)) {
     abort_pk_not_defined()
   }
-
-  pwalk(list(def$fks[i], def$pks[i], def$table[i]), ~ {
-    is_match <- !is.na(vec_match(..1$ref_column, ..2$column))
-    if (fail_fk && any(is_match)) {
-      abort_first_rm_fks(..3, ..1$table[is_match])
-    }
-  })
 
   # Talk about it
   if (is.null(table_name)) {
--- a/R/primary-keys.R
+++ b/R/primary-keys.R
@@ -242,8 +242,7 @@ dm_get_all_pks_def_impl <- function(def, table = NULL) {
 #' @export
 #' @examplesIf rlang::is_installed("nycflights13") && rlang::is_installed("DiagrammeR")
 #' dm_nycflights13() %>%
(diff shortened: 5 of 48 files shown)


TSchiefer:
currently still missing:
1. glimpse.dm()
2. dm_draw()

Thinking that this could be done in a follow-up PR

R/primary-keys.R
```diff
@@ -7,6 +7,8 @@
 #' If `force == TRUE`, the function will replace an already
 #' set key, without altering foreign keys previously pointing to that primary key.
 #'
+#' @details There can be only one primary key per table in a [`dm`].
```
krlmlr, line 10: Mention the possibility to add multiple additional unique keys?

R/unique-keys.R
```diff
@@ -0,0 +1,269 @@
...
+#'
+#' @inheritParams dm_add_pk
+#'
+#' @details The difference between a primary key (PK) and a unique key (UK) consists in the following:
+#' - when a local `dm` is copied to a database (DB) with `copy_dm_to()`, a PK will be set on the DB by default
+#' - a PK can be set as an `autoincrement` key (also implemented on certain DBMS when the `dm` is transferred to the DB)
+#' - there can be only one PK for each table, whereas there can be unlimited UKs
+#' - a UK will be used, if the same table has an autoincrement PK in addition, to ensure that during delta load processes on the DB
+#'   the foreign keys are updated accordingly. If no UK is available, the insertion is done row-wise, which also ensures a correct
+#'   matching, but can be much slower.
```
krlmlr, line 16: delta load -> `dm_rows_append()`

R/unique-keys.R
```diff
@@ -0,0 +1,269 @@
...
+# both "table" and "column" must be characters
+# in {datamodelr}, a primary key may consist of more than one columns
+# a key will be added, regardless of whether it is a unique key or not; not to be exported
+dm_add_uk_impl <- function(dm, table, column) {
+  def <- dm_get_def(dm)
+  i <- which(def$table == table)
+
+  def$uks[[i]] <- vctrs::vec_rbind(
+    def$uks[[i]],
+    tibble(column = !!list(column))
```
krlmlr, line 74: `new_uk()` ? Same for `dm_add_pk_impl()` ?

R/unique-keys.R
```diff
@@ -0,0 +1,269 @@
...
+#'
+#' @return A logical value: `TRUE` if the given table has a unique key, `FALSE` otherwise.
+#'
+#' @examplesIf rlang::is_installed("nycflights13")
+#' dm_nycflights13() %>%
+#'   dm_has_uk(flights)
+#' dm_nycflights13() %>%
+#'   dm_has_uk(planes)
+#' @export
+dm_has_uk <- function(dm, table, ...) {
```
krlmlr, line 97: Not sure we need that, `dm_has_fk()` is already deprecated (because there can be many FKs but just one PK).

R/upgrade.R
```diff
@@ -46,6 +46,8 @@ dm_upgrade <- function(dm, quiet) {
     def <- unclass(dm)$def
     def$pks <- map(def$pks, mutate, autoincrement = FALSE) %>%
       vctrs::as_list_of(new_pk())
+    def$uks <- list_of(new_uk())
```
krlmlr, line 49: Are we importing unique keys implicitly created by foreign keys here?

tests/testthat/test-unique-keys.R
```diff
@@ -0,0 +1,81 @@
...
+      dm_nycflights_small(),
+      flights,
+      time_hour,
+      weather,
+      time_hour
+    ) %>%
+      dm_rm_uk(weather, time_hour, fail_fk = FALSE)
+  })
+
+  expect_snapshot(
```
krlmlr, line 60: We have `expect_snapshot_error()` here.

R/examine-constraints.R
```diff
@@ -120,7 +120,7 @@ kind_to_long <- function(kind) {
 
 check_pk_constraints <- function(dm, progress = NA, top_level_fun = NULL) {
   pks <- bind_rows(
-    list(PK = dm_get_all_pks_impl(dm), UK = dm_get_all_uks_impl(dm)),
+    list(PK = dm_get_all_pks_impl(dm), UK = dm_get_all_uks_impl(dm) %>% rename(pk_col = uk_col)),
```
krlmlr, line 123: There are subtleties with checking a unique key because multiple `NULL` values do not disturb uniqueness (except, of course, in SQL Server 😠 ).

> A unique constraint on T is satisfied if and only if there do not exist two rows R1 and R2 of T such that R1 and R2 have the same non-NULL values in the unique columns.

https://sqlperformance.com/2020/03/t-sql-queries/null-complexities-part-4-missing-standard-unique-constraint is an excellent overview, quote taken from there.

I suggest that we create a `check_uk_constraints()` and (later) optional arguments for `check_key()` that control this.

R/foreign-keys.R
```diff
@@ -84,6 +84,21 @@ dm_add_fk <- function(dm, table, columns, ref_table, ref_columns = NULL, ...,
     ref_col_name <- get_key_cols(ref_key)
   } else {
     ref_col_name <- names(eval_select_indices(ref_col_expr, colnames(ref_table_obj)))
+    # check if either a PK or UK already matches ref_col_name
+    all_keys <- dm_get_all_pks_impl(dm) %>%
+      rename(uk_col = pk_col) %>%
+      bind_rows(dm_get_all_uks_impl(dm)) %>%
+      filter(table == ref_table_name)
```
krlmlr, line 91: Perhaps:

```suggestion
    all_keys <-
      dm_get_all_pks_impl(dm, table) %>%
      rename(uk_col = pk_col) %>%
      bind_rows(dm_get_all_uks_impl(dm, table))
```

R/foreign-keys.R
```diff
@@ -84,6 +84,21 @@ dm_add_fk <- function(dm, table, columns, ref_table, ref_columns = NULL, ...,
...
+    # setequal() could also be used for matching, but IMHO the order should matter
+    matches_keys <- map_lgl(all_keys$uk_col, identical, ref_col_name)
+    if (!any(matches_keys)) {
+      if (check) {
+        if (!is_unique_key_se(ref_table_obj, ref_col_name)$unique) {
+          abort_not_unique_key(ref_table_name, ref_col_name)
+        }
+      }
+      dm <- dm_add_uk_impl(dm, ref_table_name, ref_col_name)
```
krlmlr, line 100: Do we want to print a message here?

R/foreign-keys.R
```diff
@@ -84,6 +84,21 @@ dm_add_fk <- function(dm, table, columns, ref_table, ref_columns = NULL, ...,
...
+    # check if either a PK or UK already matches ref_col_name
+    all_keys <- dm_get_all_pks_impl(dm) %>%
+      rename(uk_col = pk_col) %>%
+      bind_rows(dm_get_all_uks_impl(dm)) %>%
+      filter(table == ref_table_name)
+    # setequal() could also be used for matching, but IMHO the order should matter
+    matches_keys <- map_lgl(all_keys$uk_col, identical, ref_col_name)
+    if (!any(matches_keys)) {
+      if (check) {
+        if (!is_unique_key_se(ref_table_obj, ref_col_name)$unique) {
```
krlmlr, line 96: Where is `is_unique_key_se()` defined?

R/primary-keys.R
```diff
@@ -217,8 +219,9 @@ dm_get_all_pks_def_impl <- function(def, table = NULL) {
 #' Remove a primary key
 #'
 #' @description
-#' `dm_rm_pk()` removes one or more primary keys from a table and leaves the [`dm`] object otherwise unaltered.
-#' An error is thrown if no private key matches the selection criteria.
+#' If a table name is provided, `dm_rm_pk()` removes the primary key from this table and leaves the [`dm`] object otherwise unaltered.
+#' If no table is given, the `dm` is stripped of all primary keys at once.
+#' An error is thrown if no primary key matches the selection criteria.
 #' If the selection criteria are ambiguous, a message with unambiguous replacement code is shown.
 #' Foreign keys are never removed.
```
krlmlr, line 226: Should we now change that: removal of a primary or unique key removes all associated foreign keys?

R/unique-keys.R
```diff
@@ -0,0 +1,269 @@
...
+#'
+#' @details The difference between a primary key (PK) and a unique key (UK) consists in the following:
+#' - when a local `dm` is copied to a database (DB) with `copy_dm_to()`, a PK will be set on the DB by default
+#' - a PK can be set as an `autoincrement` key (also implemented on certain DBMS when the `dm` is transferred to the DB)
+#' - there can be only one PK for each table, whereas there can be unlimited UKs
+#' - a UK will be used, if the same table has an autoincrement PK in addition, to ensure that during delta load processes on the DB
+#'   the foreign keys are updated accordingly. If no UK is available, the insertion is done row-wise, which also ensures a correct
+#'   matching, but can be much slower.
+#' - a UK can generally enhance the data model by adding additional information
+#' - if a foreign key is added to point at a table without a PK and a UK, a UK is automatically added to that table.
```
krlmlr, line 18:
```suggestion
#' - if a foreign key is added to point at a table without a corresponding PK or UK, a UK is automatically added to that table.
```

tests/testthat/test-unique-keys.R
```diff
@@ -0,0 +1,81 @@
+test_that("unique keys", {
+  skip_if_remote_src()
```
krlmlr, line 2: Can we keep the snapshot and ensure stability by never printing the whole dm (or calling `dm_ptype()` before printing)?

krlmlr (commented):
Thanks, great!

This is a bit more complex than anticipated, most review comments really belong in a new issue. For this iteration, I think we need:

- [x] upgrade: add previously implicit unique keys (need to adapt test, IIRC we don't have this in the current example dm)
- [x] remove `dm_has_uk()`

About everything else can go in a new issue, or as FIXMEs in the code.

We also need to update the articles. Need to find a good small example in relational.fit or elsewhere.

R/foreign-keys.R
TSchiefer: https://github.com/cynkra/dm/blob/main/R/key-helpers.R#L76

TSchiefer:
> upgrade: add previously implicit unique keys (need to adapt test, IIRC we don't have this in the current example dm)

luckily we use `dm_for_filter()` for the update tests, which has one FK pointing to a non-PK. Therefore we don't have to use a new outdated dm object.

R/unique-keys.R
TSchiefer: see 845361b24b768c4cc5538edcb1f9bb3d393f1c73

R/upgrade.R
TSchiefer: now yes, see 75d5ad42b281db4c4303c6f4e9a0efbcca2a1897

R/unique-keys.R
TSchiefer: see e1677a89cd8786aa9cce5c3298864fbf330847ae

R/unique-keys.R
TSchiefer: see 75d5ad42b281db4c4303c6f4e9a0efbcca2a1897

tests/testthat/test-unique-keys.R
TSchiefer: see 753a137931ccc8540f90bb29236e5dc0dbcc72aa

R/primary-keys.R
TSchiefer: see 786ee7f737f92be45df3b002e67fc108ab82f62f

R/foreign-keys.R
TSchiefer: I had that originally, but -- mainly due to dm_for_filter() having a UK created each time it's called -- it screwed up our snapshots. Thinking of it, maybe we should repair `dm_for_filter()`, since I don't think that is intentional.
Will add a cli::cli_alert_info() call.

R/examine-constraints.R
TSchiefer: OK, would postpone this to the next PR

R/primary-keys.R
TSchiefer: Sure, we can do that in a future PR

R/foreign-keys.R
TSchiefer: Will also do this in a future PR; it seems though, that "repairing" dm_for_filter() causes bigger problems with the snapshots than leaving it the way it is and accepting the UK-message in the affected snapshots.

tests/testthat/test-unique-keys.R
TSchiefer: see [debf12d](https://github.com/cynkra/dm/pull/1716/commits/debf12da36db0dbf5d203fb467fd58f97472b8e8)

R/foreign-keys.R
TSchiefer: see [a04dd3b](https://github.com/cynkra/dm/pull/1716/commits/a04dd3b00a90fa9e09312e88247a6067159752f7)

R/unique-keys.R
TSchiefer: see [9cadb3c](https://github.com/cynkra/dm/pull/1716/commits/9cadb3cec52aef8588ce84e5ce7aa5d910de232e)

TSchiefer:
one of the rare times, when the restart of a check leads to a different result: https://github.com/cynkra/dm/actions/runs/3794088165/jobs/6451933983 vs. https://github.com/cynkra/dm/actions/runs/3794088165/jobs/6452134441

R/dm.R
```diff
@@ -145,13 +145,15 @@ new_keyed_dm_def <- function(tables = list()) {
...
-                       fks_df = tibble(table = character(), fks = list())) {
+                       fks_df = tibble(table = character(), fks = list()),
+                       uks_df = tibble(table = character(), uks = list())) {
   # Legacy
   data <- unname(tables)
   table <- names2(tables)
 
   stopifnot(all(pks_df$table %in% table))
   stopifnot(all(fks_df$table %in% table))
+  stopifnot(all(uks_df$table %in% table))
```
krlmlr, line 156:
```suggestion
  stopifnot(all(uks_df$table %in% table))
  stopifnot(all(fks_df$table %in% table))
```

R/unique-keys.R
```diff
@@ -0,0 +1,250 @@
...
+
+  pwalk(list(def$fks[i], def$uks[i], def$table[i]), ~ {
+    is_match <- !is.na(vec_match(..1$ref_column, ..2$column))
+    if (fail_fk && any(is_match)) {
+      abort_first_rm_fks(..3, ..1$table[is_match])
+    }
+  })
+
+  # Talk about it
+  if (is.null(table_name)) {
```
krlmlr, line 215:
```suggestion
  if (is.null(table_name) || quo_is_null(columns)) {
```

R/unique-keys.R
```diff
@@ -0,0 +1,250 @@
...
+    message("Removing unique keys: %>%")
+    message("  ", glue_collapse(glue("dm_rm_uk({tick_if_needed(def$table[i])})"), " %>%\n  "))
+  }
+
+  # Execute
+  def$uks[i] <- list_of(new_uk())
```
krlmlr, line 221: We need more work here -- this always removes all unique keys, we need to be more precise.

krlmlr (commented):
We also want `dm_validate()` to check if there is a unique or primary key for each foreign key. And more tests for adding and removing unique keys, especially for the `fail_fk = FALSE` case where we should now *remove* existing foreign keys. The default `fail_fk = TRUE` should be `fail_fk = NULL` for both foreign and unique key removal.

R/unique-keys.R
```diff
@@ -0,0 +1,250 @@
...
+
+  pwalk(list(def$fks[i], def$uks[i], def$table[i]), ~ {
+    is_match <- !is.na(vec_match(..1$ref_column, ..2$column))
+    if (fail_fk && any(is_match)) {
+      abort_first_rm_fks(..3, ..1$table[is_match])
+    }
+  })
+
+  # Talk about it
+  if (is.null(table_name) || quo_is_null(columns)) {
```
krlmlr, line 215:
```suggestion
  if (is.null(table_name) || quo_is_null(columns)) {
```

R/unique-keys.R
TSchiefer: excellent point, thanks, I'm on it.

TSchiefer:
So I am wondering, after e9f07c5601e372d38a1e3398fbb85cdcad5ee66c we get a prompt for using code that we actually used. Do we really want that? Do we want to give more precise messages, including the columns?
``` r
suppressMessages(devtools::load_all("~/git/cynkra/dm"))

dm <- dm_for_filter()
dm_get_all_uks(dm)
#> # A tibble: 1 × 2
#>   table uk_col
#>   <chr> <keys>
#> 1 tf_6  n

dm_rm_fk(dm, tf_5, m, tf_6, n) %>% 
  # Now, with `dm_rm_uk(tf_6)`, we get a message with the code that we're actually using
  dm_rm_uk(tf_6)
#> Removing unique keys: %>%
#>   dm_rm_uk(tf_6)
#> Tables: `tf_1`, `tf_2`, `tf_3`, `tf_4`, `tf_5`, `tf_6`
#> Columns: 20
#> Primary keys: 6
#> Foreign keys: 4
```

<sup>Created on 2023-01-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

TSchiefer:
just realized it's currently possible to add a UK that's the same as an existing PK. I cannot think of a scenario where that behaviour has an advantage over a no-op with a warning, that such a PK already exists. What do you think?

TSchiefer:
OK, I'll make creation of a UK an error, if the same PK exists.

> We need to give precise messages: the code we generate should itself be unambiguous.

Are you referring here to my reprex above?
Do I understand you correctly, that 
1. in case the user doesn't provide `table` all UKs for all tables should be removed?
2. in case the user doesn't provide `columns` all UKs for the given table should be removed?
3. in either of those cases we should generate the code for removal of each UK individually?

TSchiefer:
With 4e66e4c2e quite a few tests for deconstruct were broken. I believe, that the tests can and should be repaired, but I would say it's OK to postpone this to a future PR as I don't understand the key structure of the dm that's being created from the keyed tibbles. Now the tests are being skipped: 0c0f73f

So far the FKs are not yet removed when removing a PK or a UK. Would you like to have this feature in this PR?

krlmlr:
1.-3. -- exactly, we can model the behavior after `dm_rm_fk()` .

Let's wrap up this PR as soon as possible and finish the rest in separate PRs

TSchiefer:
any idea regarding this error: https://github.com/cynkra/dm/actions/runs/3830860239/jobs/6519443932#step:4:1680?

krlmlr:
Can be ignored, but the validation error looks worrying.

TSchiefer:
I needed to disable a few tests, because we are now testing in validate if all UKs are properly defined and if all FKs are pointing to either a PK or a UK (4e66e4c), and that was not always the case in examples or in tests

I also realized that key tracking is necessary for UKs when zooming a dm (3904400). 2 small tests for that are implemented as well.

TSchiefer:
@krlmlr could you please have a look at https://github.com/cynkra/dm/actions/runs/3836327768/jobs/6530377269? I don't understand where the error is coming from

TSchiefer:
Thanks very much.
I guess I was irritated, because locally everything was working fine. But it also is working fine locally with the updated snapshot.
I don't know how much work it is to post a comment in the PR. It might be useful. Also not sure if in some cases the PR could get spammed with such comments.

TSchiefer:
Think I'm relatively happy now, could you please have another look?

R/primary-keys.R
```diff
@@ -239,6 +244,7 @@ dm_get_all_pks_def_impl <- function(def, table = NULL) {
 #' @export
 #' @examplesIf rlang::is_installed("nycflights13") && rlang::is_installed("DiagrammeR")
 #' dm_nycflights13() %>%
+#'   dm_rm_fk(flights, origin, airports) %>%
 #'   dm_rm_pk(airports, fail_fk = FALSE) %>%
 #'   dm_draw()
```
krlmlr, line 249: If we really must change the example:

```suggestion
#' dm_nycflights13() %>%
#'   dm_rm_fk(flights, origin, airports) %>%
#'   dm_rm_pk(airports) %>%
#'   dm_draw()
```

R/unique-keys.R
```diff
@@ -0,0 +1,279 @@
...
+
+  if (!is_empty(def$pks[[i]]$column) && vctrs::vec_in(def$pks[[i]]$column[1], list(column))) {
+    abort_no_uk_if_pk(table, column)
+  }
+
+
+  def$uks[[i]] <- vctrs::vec_rbind(
+    def$uks[[i]],
+    new_uk(list(column))
+  )
```
krlmlr, line 80: Should we also fail if a uk with the same column names exists? Can we also do:

```suggestion
  if (!is_empty(def$pks[[i]]$column) && identical(def$pks[[i]]$column[[1]], column)) {
    abort_no_uk_if_pk(table, column)
  }


  def$uks[[i]] <- vctrs::vec_rbind(
    def$uks[[i]],
    new_uk(list(column))
  )
```

krlmlr (commented):
I'll check again what happens if we keep the tests unchanged in this PR.

krlmlr:
Working on adding uk support for deconstruct.

R/primary-keys.R
TSchiefer: sure we can do that. 
Actually, in the near future (#1729), the dm_rm_fk() call won't be necessary anymore. Shall we then keep the argument `fail_fk` for `dm_rm_uk()`/`dm_rm_pk()`? I guess if we keep it, the default would need to change to `FALSE` and all FKs would be removed then.

R/unique-keys.R
TSchiefer: Good point, failing if a matching UK already exists makes a lot of sense. Need to adapt the error message then.
Thanks for your suggestion for checking for an existing, matching PK, looks good to me.

TSchiefer:
> I'll check again what happens if we keep the tests unchanged in this PR.

Maybe we should just do it all the way in this PR, including the implementation for removing all FKs pointing to PKs/UKs that are to be removed? Otherwise we always get an unfinished, non-happy state for a while on main.

krlmlr:
That's too big. I think we should, at least for now, allow both implicit and explicit unique keys. They should coexist and be treated equally.

- We can always add a foreign key, this creates an implicit unique key.
- If we create an explicit unique key for an existing foreign key, the database schema should not change.
- If we set a primary key and a unique key with exactly the same columns exists, the unique key is promoted (=removed).

This avoids all those corner cases we'd have to deal with otherwise.

krlmlr:
Can you please make sure that the documentation is in line with the new paradigm?

R/foreign-keys.R
```diff
@@ -84,6 +90,20 @@ dm_add_fk <- function(dm, table, columns, ref_table, ref_columns = NULL, ...,
     ref_col_name <- get_key_cols(ref_key)
   } else {
     ref_col_name <- names(eval_select_indices(ref_col_expr, colnames(ref_table_obj)))
+    # check if either a PK or UK already matches ref_col_name
+    all_keys <- dm_get_all_pks_impl(dm, ref_table_name) %>%
+      rename(uk_col = pk_col) %>%
+      bind_rows(dm_get_all_uks_impl(dm, ref_table_name))
+    # setequal() could also be used for matching, but IMHO the order should matter
+    matches_keys <- map_lgl(all_keys$uk_col, identical, ref_col_name)
+    if (!any(matches_keys)) {
```
krlmlr, line 99: Need to think about the correct code here.

R/primary-keys.R
krlmlr: No need to change anything here.

R/primary-keys.R
krlmlr: I think the right course of action is to deprecate the `fail_fk` argument and just keep the foreign key pointing to a then implicit unique key.

R/unique-keys.R
krlmlr: Are we changing this here or in a future PR?

R/unique-keys.R
```diff
@@ -0,0 +1,263 @@
...
+
+  new_dm3(def)
+}
+
+#' Get all primary keys of a [`dm`] object
```
krlmlr, line 85:
```suggestion
#' Get all explicit unique keys of a [`dm`] object
```

R/unique-keys.R
```diff
@@ -0,0 +1,263 @@
...
+#'
+#' @description
+#' `dm_get_all_uks()` checks the `dm` object for unique keys and
+#' returns the tables and the respective unique key columns.
+#'
```
krlmlr, line 90: Mention implicit and explicit unique keys?

R/unique-keys.R
```diff
@@ -0,0 +1,263 @@
...
+#' @param fail_fk
+#'   Boolean: if `TRUE` (default), will throw an error
+#'   if there are foreign keys addressing the unique key that is to be removed.
+#'
+#' @family primary key functions
+#'
+#' @return An updated `dm` without the indicated unique key(s).
+#'
+#' @export
+dm_rm_uk <- function(dm, table = NULL, columns = NULL, ..., fail_fk = TRUE) {
```
krlmlr, line 161: Let's get rid of `fail_fk` here.

tests/testthat/_snaps/dm.md
```diff
@@ -128,6 +128,7 @@
       Tables: `tf_1`, `tf_2`, `tf_3`, `tf_4`, `tf_5`, `tf_6`
       Columns: 20
       Primary keys: 6
+      Unique keys: 1
```
krlmlr, line 131: Why are we now seeing explicit unique keys here?

tests/testthat/test-deconstruct.R
```diff
@@ -519,6 +519,19 @@ test_that("pks_df_from_keys_info()", {
   })
 })
 
+test_that("uks_df_from_keys_info()", {
+  withr::local_seed(20220715)
+
+  dm <- dm_nycflights13(cycle = TRUE)
```
krlmlr, line 525: Use a better dm object here?

krlmlr (commented):
Thanks! Let's do one or at most two more iterations, and get this in.

tests/testthat/_snaps/dm.md
TSchiefer: Was quick to add and is easy to remove. Now that the implicit UKs still exist, I would tend to remove it.

TSchiefer:
Should `dm_get_all_uks()` also return the implicit UKs? There could be an additional column in the response of the function, logical `explicit`.
We could then also at some point think about a function `dm_make_all_uks_explicit()`.

krlmlr:
We could also include primary keys, and have an additional three-valued column "kind" instead. No strong feelings.

tests/testthat/_snaps/dm.md
TSchiefer: So neither show explicit nor implicit UKs in `show_dm()`?

krlmlr: In a new PR?

TSchiefer:
now `dm_get_all_uks()` has the following output:
``` r
suppressMessages(devtools::load_all("~/git/cynkra/dm"))

dm_for_filter() %>% 
  dm_rm_pk(tf_4, h) %>% 
  dm_get_all_uks()
#> # A tibble: 7 × 3
#>   table uk_col kind       
#>   <chr> <keys> <chr>      
#> 1 tf_1  a      PK         
#> 2 tf_2  c      PK         
#> 3 tf_3  f, f1  PK         
#> 4 tf_5  k      PK         
#> 5 tf_6  o      PK         
#> 6 tf_6  n      explicit UK
#> 7 tf_4  h      implicit UK
```

<sup>Created on 2023-01-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

TSchiefer:
Wanted to mention, that currently there is no message that an explicit UK is set when adding a FK with no corresponding PK in the parent table. But I think that could be done in a later PR.

krlmlr:
There's no longer the need to add an explicit UK when adding a new FK for new columns, these UKs remain implicit. Or were you referring to implicit UKs? We don't need to message for implicit UKs either.

R/foreign-keys.R
krlmlr: I believe 93-105 can go now entirely?

R/paste.R
```diff
@@ -178,6 +182,14 @@ dm_paste_pks <- function(dm) {
     pull()
 }
 
+dm_paste_uks <- function(dm) {
+  dm %>%
+    dm_get_def() %>%
+    dm_get_all_uks_def_impl() %>%
```
krlmlr, line 188: Should we filter for explicit UKs here?

R/primary-keys.R
```diff
@@ -239,39 +242,34 @@ dm_get_all_pks_def_impl <- function(def, table = NULL) {
...
-      details = "Note the different semantics: `fail_fk = FALSE` roughly corresponds to `rm_referencing_fks = TRUE`, but foreign keys are no longer removed."
-    )
-    if (is.null(fail_fk)) {
-      fail_fk <- !rm_referencing_fks
-    }
-  }
-
-  if (is.null(fail_fk)) {
-    fail_fk <- TRUE
+    deprecate_soft("0.2.1", "dm::dm_rm_pk(rm_referencing_fks = )")
```
krlmlr, line 263: We could reuse the same message as for `fail_fk` .

R/unique-keys.R
```diff
@@ -0,0 +1,283 @@
...
+      ), " %>%\n  "
+    ))
+  }
+  # Execute
+  # in case `length(i) > 1`: all tables have all their UKs removed, respectively
+  def$uks[i] <- if (length(i) > 1) {
+    list_of(new_uk())
+  } else {
+    list_of(filter(def$uks[[i]], !ii))
+  }
```
krlmlr, line 265: I don't understand this code, but we'll notice if it causes trouble.

krlmlr (approved):
Thanks. Can we do one more run, and merge when green?

TSchiefer:
> There's no longer the need to add an explicit UK when adding a new FK for new columns, these UKs remain implicit.

Right, I thought you did what needed to be done in that respect after your comment https://github.com/cynkra/dm/pull/1716#issuecomment-1373188198. We can do it this way of course, then there are 2 ways of creating implicit UKs:
1. removing a PK/UK with a FK pointing to it
2. adding an FK without an existing PK/UK

Going to update the docs as well

R/unique-keys.R
TSchiefer: `ii` contains the UK columns that are to be removed. We take the list of the current UKs and filter out `ii`. Well, as you say, let's see

TSchiefer:
I addressed your latest comments, merging now.

PR #1741, opened 2023-01-16
feat!: `dm_get_all_pks()`, `dm_get_all_fks()`, and `dm_get_all_uks()` require unquoted table names as input, for consistency with other parts of the API

TSchiefer:
in order to be more in line with the rest of the package's functions

### DIFF
--- a/R/build_copy_queries.R
+++ b/R/build_copy_queries.R
@@ -25,7 +25,7 @@ build_copy_queries <- function(dest, dm, set_key_constraints = TRUE, temporary =
     # to be a scalar
     pk_col <-
       dm %>%
-      dm_get_all_pks(x) %>%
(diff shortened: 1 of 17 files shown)


krlmlr:
Thanks, the code looks good. This is a breaking change, but consistency is important.

Are there other instances in the API where we're using character vectors but should switch to tidyselect?

TSchiefer:
I went through https://dm.cynkra.com/reference/index.html: it seems like everywhere else where arguments are referring to existing table names or column names, we support tidyselect. 
In the case of `decompose_table()`, for its argument `new_id_column` we even allow unquoted input for a non-existing column name, which most likely is not the best choice.

krlmlr:
Thanks!

PR #1745, opened 2023-01-19
feat: `dm_rows_append()` also works for local dm, with support for autoincrement primary keys

### LINKED ISSUE #1727: support AI PKs also for local dm's
(see https://github.com/cynkra/dm/pull/1654#issuecomment-1351745127)

still left to do after https://github.com/cynkra/dm/pull/1689, https://github.com/cynkra/dm/pull/1698 & co.

TSchiefer:
closes #1727


krlmlr:
Waiting for db checks before reviewing this.

R/rows-dm.R
```diff
@@ -256,6 +256,8 @@ do_rows_insert <- function(x, y, by = NULL, ..., autoinc_col = NULL) {
 do_rows_append <- function(x, y, by = NULL, ..., in_place = FALSE, autoinc_col = NULL) {
   if (is.null(autoinc_col)) {
     return(rows_append(x, y, ..., in_place = in_place))
+  } else if (inherits(x, "data.frame")) {
+    return(rows_append_ai_local(x, y, ..., autoinc_col))
```
krlmlr, line 260: Do we need `rows_append_ai_remote()` too?

Each function should operate on the same level of abstraction. We could also get rid of the `return` statements then.

krlmlr (commented):
`dm_rows_run()` is super complex already. Is there a way to implement this without adding more complexity there? Should we take a step back and simplify that function before attempting this feature?

R/rows-dm.R
```diff
@@ -381,14 +389,22 @@ dm_rows_run <- function(x, y, rows_op_name, top_down, in_place, require_keys, pr
...
+      if (!is.null(get_autoinc_col(x, table, colnames(tbl)))) {
+        tbls <- align_autoinc_fks(tbls, x, table, res$ai_lookup)
+      }
     } else if (!in_place) {
       op_results[[i]] <- res
     }
   }
 
-  if (in_place) {
+  if (in_place && is_src_db(x)) {
```
krlmlr, line 407: We should have `!in_place` always if `!is_src_db()` -- I don't understand the necessity of this and the above changes.

tests/testthat/test-rows-dm.R
```diff
@@ -277,18 +277,111 @@ test_that("dm_rows_append() works with autoincrement PKs and FKS for selected DB
...
+      dm_ai_empty_remote,
+      dm_ai_insert,
+      in_place = TRUE,
+      progress = FALSE
+    ) %>%
+      collect()
+  )
+
+  expect_snapshot(
+    variant = my_test_src_name,
```
krlmlr, line 291: Do we really need the snapshot variant here? It didn't seem necessary before.

krlmlr (commented):
Let's shelve this for now, we need to submit to CRAN very soon.

krlmlr:
Thanks!

PR #1751, opened 2023-01-20
test: Restore database checks on GHA

TSchiefer:
cherry-picked from #1745

not sure if anything needs to be updated, but it seems to have been accidentally deleted earlier


krlmlr:
Can you please start by fixing the duckdb checks? We'll see if we need to change anything else after that.

TSchiefer:
The problem is that I don't get these errors locally, the R CMD check passes using duckdb (and all the other DBMS).
Do you see these errors on main or here when testing on duckdb? (in test-flatten.R (and by association in test-zzx-deprecated.R, but it's essentially the same test))

I saw that there were some problems with a warning on SQLite, which is something I also saw locally. I will push the fix for it, but for now I have no idea, why the same errors happen consistently for all different DBMS on GHA.

krlmlr (commented):
Thanks. Can you please use `skip()` instead of commenting out? You can split tests or add a new test where necessary.

krlmlr:
dplyr 1.1.0 is out, we're seeing weird messages now. Should we fix as part of this PR?

TSchiefer:
Could it be that for all DB checks for some reason different R-pkg versions are being used?
e.g. https://github.com/cynkra/dm/actions/runs/4131905764/jobs/7140105648 doesn't know the arg `multiple = "all"` for joins.

TSchiefer:
Ahh, nevermind... Obviously it's `dbplyr:::left_join.tbl_lazy()` & other joins that don't know about this argument.
Do we need to restrict using the argument to the case of a local test src?

TSchiefer:
So far so good.
Nevertheless https://github.com/cynkra/dm/blob/reinstate-db-checks/tests/testthat/_snaps/duckdb/table-surgery.md is wrong, the resulting tables should look the same as https://github.com/cynkra/dm/blob/reinstate-db-checks/tests/testthat/_snaps/df/table-surgery.md
same for https://github.com/cynkra/dm/blob/reinstate-db-checks/tests/testthat/_snaps/duckdb/flatten.md, which should be the same as https://github.com/cynkra/dm/blob/reinstate-db-checks/tests/testthat/_snaps/df/flatten.md

krlmlr:
Thanks, weird.

I would have thought we could use this opportunity to avoid as many `multiple = "all"` calls in the tests as possible. This requires redesigning the tests, of course.

Merging now, we can revisit in a follow-up PR.

PR #1771, opened 2023-02-07
chore: Establish compatibility with dplyr 1.1.0

TSchiefer:
and also tweak some dm code to embrace these changes.
basically makes dplyr 1.1.0 necessary though. Not sure if it's possible to avoid this.


krlmlr:
Thanks!

PR #1789, opened 2023-02-13
feat: `dm_from_con()` can retrieve multiple schemas, pass a character vector to the `schema` argument

### LINKED ISSUE #1533: `dm_from_con()` with multiple schemata fails with opaque error

owenjonesuob:
Currently, multiple schemas cannot be retrieved due to the use of `==` for comparison, instead of `%in%`.

Additionally, SQL tables within the same schema cannot have the same name; but two tables in _different_ schemas _can_ have the same name. One possible solution is to use `schema.table`, rather than just `table`, as the names of the `dm` object when we ask for multiple schemas - that's the solution implemented here.

---

Closes #1533. See #1534 for additional discussion.


krlmlr:
Thanks, this looks good.

Tables from different schemas can have the same name, but not necessarily do. How about the following approach:

- by default, we use the table name only
- in the case of a clash, we raise a clean error with actionable advice
- we offer a pattern-driven way for naming the tables, akin to the `.names` argument to `dplyr::across()`

I'm not sure yet how this is related to name repair, I suspect we can focus on naming patterns first.

owenjonesuob:
That's a really nice parallel to `dplyr::across()`, I hadn't thought of that!

In order to keep a neat separation between the "allow multiple schemas" bugfix in this PR, and the "local naming" feature in #1790, do you think the following sounds sensible?

In this PR:
* Use table name only as the local name (i.e. the behaviour of current release of {dm})
* If there is a clash, keep only the first table with that name, and raise a **warning**. For example:
  - User specifies `schemas = c("schema1", "schema2")`
  - Tables `schema1.table` and `schema2.table` both exist
  - We have two local names of `table`, one pointing to `schema1.table` and the other to `schema2.table`
  - Keep `table -> schema1.table`, discard `table -> schema2.table`
  - Raise warning that `table` refers to `schema1.table` since `table` existed in multiple schemas (`schema1` and `schema2`), and only the first (positionally) has been kept
  - If the user wants to keep `table -> schema2.table` instead, they could try again with `schemas = c("schema2", "schema1")`
  
  We _could_ raise an error here instead, but I think the only advice we can provide is for the user to rename the one of the tables in the database itself, which is probably not feasible. So an error might completely block this functionality for users, whereas an unambiguous warning would allow them to carefully sidestep the problem by making sure `schemas` is ordered correctly for their use case.

In #1790:
* Add a `.names = NULL` argument to `dm_from_con()`, where `NULL` is a "smart default" like [the one used in `dplyr::across()`](https://github.com/tidyverse/dplyr/blob/af5f194faabdab8d32ab666bccf56b4e59c77575/R/across.R#L456). So the user can specify their own pattern, taking advantage of special variables `.schema` and `.table` if they would like to - but by default:
  - When `length(schema) == 1`, use `"{.table}"`, akin to the "single" default of `"{.col}"` in `across()`
  - Otherwise use `"{.schema}.{.table}"`, akin to the "plural" default of `"{.col}_{.fn}"` in `across()`, but adhering more closely to how entities are usually specified in SQL
* If there is a clash, raise an **error** with actionable advice - we will never have clashes with the "smart default" just described, since table names must be unique within a schema; so "actionable advice" is probably showing the user where the clash happens, and suggesting that they change the spec they have passed to `.names`

owenjonesuob:
I've updated this PR with the changes I suggested earlier - will have a go at the changes for #1790 asap too, on the relevant branch!

R/db-helpers.R
```diff
@@ -150,12 +155,47 @@ get_src_tbl_names <- function(src, schema = NULL, dbname = NULL) {
...
+          ~ cli::format_inline(
+            "Local name {.field {.y}} will refer to {.cls {(.x[1])}}, ",
+            "rather than to {.or {.cls {(.x[-1])}}}"
+          )
+        ) %>%
+          purrr::set_names(rep("*", length(clashes)))
+      ))
+
+      # Keep only first schema for each table_name
+      names_table <- slice_head(names_table, by = table_name)
```
krlmlr, line 193: Nice! I didn't know this shortcut, especially with omitting `n` .

krlmlr (commented):
Rerunning tests just to be sure, then merging.

PR #1799, opened 2023-02-17
chore: Gitpod configuration with databases

Antonov548:


krlmlr:
Thanks! We also need SQL Server installed on GitPod.

krlmlr:
https://learn.microsoft.com/en-us/sql/linux/quickstart-install-connect-ubuntu?view=sql-server-ver16 are the official installation instructons.

PR #1806, opened 2023-02-27
chore: Minimal patch to fix multiple match updates

DavisVaughan:
We are preparing to release dplyr 1.1.1 and dm popped up in the revdep checks.

We realized that we didn't get the `multiple` argument quite right the first time around 😞 . It was too aggressive since it warned on both one-to-many and many-to-many joins. In 1.1.1 we've made two improvements:
- `multiple` now defaults to `"all"`. The options of `NULL`, `"error"`, and `"warning"` are deprecated.
- `relationship` is a new argument to add known constraints onto the join procedure, such as `"many-to-one"`, which replaces `multiple = "error"`.

The default of `relationship` checks to see if there is a `many-to-many` relationship between the keys of `x` and `y` and will warn if one is present. This should be _much_ rarer than what we checked for before, and targets the most dangerous case that we were trying to warn the user about.

You can read all about `relationship` here https://github.com/tidyverse/dplyr/pull/6753, along with the issues linked there.

Unfortunately it does affect dm by breaking some tests. I've done the minimal amount of work to get the tests working again, but I think you may need to do a full review of `multiple = "all"` usage. I think most of the time you will just be able to remove it, as I doubt you do many many-to-many joins here. The annoying part is ensuring the tests pass on CRAN and dev dplyr, so maybe it is worth leaving the other `multiple = "all"` calls in there until dplyr 1.1.1 is out and you can depend on it.

We plan to submit dplyr 1.1.1 in 2-3 weeks.

This should be compatible with both dev and CRAN dplyr. It would help us out if you could go ahead and send a patch version of your package to CRAN ahead of time! Thanks!

### DIFF
--- a/tests/testthat/test-flatten.R
+++ b/tests/testthat/test-flatten.R
@@ -300,25 +300,12 @@ test_that("tests with 'bad_dm' work", {
   skip_if_src("postgres")
 
   # flatten bad_dm() (no referential integrity)
-  #
-  # Warning, because since dplyr 1.1.0 dm_flatten_to_tbl()
-  # issues warnings, when there are multiple rows in `y` to match rows in `x`
-  # This means here, that the PK in the parent table is violating key constraints
-  if (is_db(my_test_src())) {
+  if (is_db(my_test_src()) || utils::packageVersion("dplyr") >= "1.1.0.9000") {
     expect_equivalent_tbl(
       dm_flatten_to_tbl(bad_dm(), tbl_1, tbl_2, tbl_3),
       tbl_1() %>%
         left_join(tbl_2(), by = c("a" = "id", "x")) %>%
         left_join(tbl_3(), by = c("b" = "id"))
-    )
-  } else {
-    expect_warning(
-      expect_equivalent_tbl(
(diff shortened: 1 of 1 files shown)


krlmlr (approved):
Thanks a lot, Davis, much appreciated!

TSchiefer (approved):
Thank you, LGTM.

To me it seems that with https://github.com/cynkra/dm/commit/8aeca84c4dddb1f34196dfd6d412f9d95ebc4206 in #1817 in addition to this PR, we should be fine, at least for the currently implemented tests. 
We should probably think about situations where we get many-to-many relations in the join operations in our functions that are currently not covered by our tests.

krlmlr:
Thanks!

krlmlr:
dm is designed to avoid many-to-many joins in "normal" operation. Only an inconsistent dm will give many-to-many join warnings. We could try and see what happens then, but it doesn't seem too important.

DavisVaughan:
@krlmlr that is kind of what i figured and was hoping for! that should align pretty nicely with the dplyr behavior then

PR #1812, opened 2023-02-28
Fix typo in tests

TSchiefer:
tests were always skipped.


TSchiefer:
this one's a bit puzzling to me: https://github.com/cynkra/dm/actions/runs/4292977588/jobs/7480303465

TSchiefer:
FYI: I updated the DB tests for the DBs other than duckdb now also here, even though there are PRs directly to the main branch. Even though it's not exactly the reason for this PR it seemed related enough to me...

krlmlr:
Thanks!

PR #1817, opened 2023-02-28
chore: Adapt to rlang 1.1.0 changes

TSchiefer:
rlang 1.1.0:
- flatten and squash family will be deprecated and superseded by purrr::list_c() and purrr::list_flatten().

In addition:
In https://github.com/cynkra/dm/pull/1806 Davis mentioned, that we might need some more tweaks in order to adapt to dplyr 1.1.1. 
8aeca84c4 ensures that once dplyr 1.1.1 is out, tests won't fail because of unexpected warnings. It seems to me that we don't need further changes.

The scattered `multiple = "all"` calls probably need to remain at least until then.

### DIFF
--- a/R/bind.R
+++ b/R/bind.R
@@ -5,12 +5,12 @@ dm_bind_impl <- function(dms, repair, quiet, repair_arg = "", caller = caller_en
 
   walk(dms, check_dm)
   walk(dms, check_not_zoomed)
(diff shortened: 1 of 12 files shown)


krlmlr:
The two errors on Postgres and SQL Server look very similar, not snapshot errors. Can you please take another look?

TSchiefer:
should be good now

krlmlr:
Thanks!

PR #1838, opened 2023-03-14
chore: Use roxyglobals

krlmlr:
To avoid editing `globalVariables()` manually.


krlmlr:
This works beautifully, now we only need to remove `global.R` entirely.

PR #1868, opened 2023-03-29
use normal dev version of fledge

TSchiefer:
poking a bit in the dark here, but -- seeing the error messages in the action -- does this make sense, @krlmlr?

1. fledge@cran-watching: https://github.com/cynkra/dm/actions/runs/4551073535/jobs/8024728595
2. run only on main: https://github.com/cynkra/dm/actions/runs/4553712072/jobs/8030620020


krlmlr:
Thanks, this was missing for some reason. Now in main.

PR #1872, opened 2023-04-06
change fonts url to cynkra.com

wael-sadek:
Old URLs use the s3 bucket's url directly. The new URLs will benefit from the cloudfront and also free some redundant resources.

### DIFF
--- a/inst/cheatsheet.css
+++ b/inst/cheatsheet.css
@@ -14,21 +14,21 @@ pre {
 
 @font-face {
   font-family: "frutiger-light", sans-serif;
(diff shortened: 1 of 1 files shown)


krlmlr:
@maelle: Can we use cynkratemplate's CSS here, to avoid the redundant reference to the font resources?

maelle:
Apart from the font, the rules are not necessarily the same as in cynkratemplate. 

## Solution A (more work IMHO)

Split the cynkratemplate CSS into fonts / the rest.
Load it from the cheatsheet here.

## Solution B (not good)

Keep as is

## Solution C

- have a more readable URL for the fonts, that'd stay the same over time.
- document it and the font-face rule in a central cynkra style doc?

krlmlr:
Thanks! Merging now, asking questions later 😁 

We should gradually work towards a long-term improvement. We should watch https://dm.cynkra.com/dev/ to see if there are unexpected changes.

PR #1888, opened 2023-06-08
feat: New `dm_set_table_descriptions()`, `dm_get_table_descriptions()` and `dm_reset_table_descriptions()` to set table labels as persistent attributes of the table object

TSchiefer:
New functions: 

- `dm_set_table_description()`
- `dm_get_table_description()`
- `dm_reset_table_description()`

similar to `dm_set_colors()` and `dm_get_colors()`

---

Tests are in place.


TSchiefer:
Weirdly, if I locally just run the tests in the file `test-dm_deconstruct.R`, there's no error. With devtools::test(), I get the same error as the CI.
Still need to figure out where the snapshot-change in test-json.R (additional digits are written in the snapshot) comes from.

Would venture to say those errors are not caused by changes in this PR.

moodymudskipper:
I find the order `value = table` confusing but I get that we want to be
consistent with `dm_set_colors()`.

Using `across()` would seem to be the tidy way to do it though :
`across(selection, ~"lightgreen")`. I assume it didn't exist when the
function was designed.

Would it be too bad of a breaking change to switch to `table = value` and
keep the other way working but undocumented, with priority on the new way
in case of ambiguity (not likely to ever happen).

`dm_get_colors()` and `dm_get_descriptions()` might also return tables as
names, the formerwould be broken but I suppose nobody uses it in prod. And
we could have a dm_colors_legacy option that we might set to `TRUE` to keep
the old behaviour everywhere.

R/draw-dm.R
```diff
@@ -334,3 +325,76 @@ dm_get_colors <- function(dm) {
...
+
+  dm_get_table_description_impl(dm, table_names)
+}
+
+dm_get_table_description_impl <- function(dm, table_names) {
+  def <- dm_get_def(dm, quiet = TRUE)
+  map(
+    table_names,
+    function(table_name) {
+      attr(def[def$table == table_name,]$data[[1]], "description")
```
krlmlr, line 395: The labelled package uses `"label"` :

```suggestion
      attr(def$data[[which(def$table == table_name)]], "label")
```

Can you use `eval_select_both()` to work with table positions instead of table names?

R/draw-dm.R
```diff
@@ -334,3 +325,76 @@ dm_get_colors <- function(dm) {
...
+      # `eval_select_table()` interprets the name `NULL` as "NULL"
+      # downside is, that no table description can be defined as "NULL"
+      desc <- if (desc == "NULL") {
+        NULL
+      } else {
+        desc
+      }
+      def$data[[which(def$table == table)]] <- structure(
+        def$data[[which(def$table == table)]],
+        "description" = desc
```
krlmlr, line 362:
```suggestion
        "label" = desc
```

This might be one of the cases where a `for` loop is easier to decipher. Also, `eval_select_both()` .

R/draw-dm.R
```diff
@@ -334,3 +325,76 @@ dm_get_colors <- function(dm) {
...
+  check_not_zoomed(dm)
+  selected <- eval_select_table(quo(c(...)), src_tbls_impl(dm), unique = FALSE)
+  def <- dm_get_def(dm, quiet = TRUE)
+  reduce2(
+    selected,
+    names(selected),
+    function(def, table, desc) {
+      # `eval_select_table()` interprets the name `NULL` as "NULL"
+      # downside is, that no table description can be defined as "NULL"
+      desc <- if (desc == "NULL") {
```
krlmlr, line 355: A new `dm_reset_table_description()` seems cleaner.

krlmlr (commented):
How do we retrieve the label of a table after it's been extracted from the dm object?

krlmlr:
The labelled R package now supports setting and resetting labels for data frames. Later, we could use that to set and retrieve attributes. For now, it's enough to make sure that we can retrieve the labels for individual tables via the GitHub version of labelled.

R/draw-dm.R
krlmlr: We now can use the new functions from {labelled}.

krlmlr (commented):
Should the new functions live in a new file? The test in `test-draw-dm.R` would be only integration tests that show that the created `.svg` has the descriptions.

TSchiefer:
> How do we retrieve the label of a table after it's been extracted from the dm object?

Good point, after using {labelled}, we can do so with:
```r
labelled::label_attribute(x)
```

R/draw-dm.R
TSchiefer: Agreed

TSchiefer:
As Antoine suggested (https://github.com/cynkra/dm/pull/1888#issuecomment-1595669894), we could think about switching the order of table and description. I believe it won't work with tidyeval then, but very likely we don't need it here.
I think we should rather not change the syntax for dm_set_colors() though.

krlmlr (commented):
Thanks. Would you like to add examples?

krlmlr (commented):
Other ideas:

- Set font size for other components
- Define labels for `dm_nycflights13()`

TSchiefer:
> Other ideas:
> 
>     * Set font size for other components
> 
>     * Define labels for `dm_nycflights13()`

Implemented both ideas in the latest commits

krlmlr:
Thanks!

PR #1900, opened 2023-06-17
chore: In `copy_dm_to()`, call `collect()` only when copying data, table by table

jangorecki:
avoid `collect()` on `dm`, run tests

### DIFF
--- a/R/db-interface.R
+++ b/R/db-interface.R
@@ -191,15 +191,15 @@ copy_dm_to <- function(dest, dm, ...,
   # constraints
 
   # use 0-rows dm object from now on
-  dmdata <- dm ## we still need it to copy actual data, we will collect tables one by one to reduce peak memory usage
-  dm <- collect(dm_ptype(dm))
+  dm ## we still need it to copy actual data, we will collect tables one by one to reduce peak memory usage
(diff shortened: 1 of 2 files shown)


R/db-interface.R
```diff
@@ -190,7 +190,10 @@ copy_dm_to <- function(dest, dm, ...,
   # FIXME: if same_src(), can use compute() but need to set NOT NULL and other
   # constraints
 
-  dm <- collect(dm, progress = progress)
+  # use 0-rows dm object from now on
+  dmdata <- dm ## we still need it to copy actual data, we will collect tables one by one to reduce peak memory usage
+  dm <- dm |> dm_ptype() |> collect()
+  #dm <- collect(dm, progress = progress)
```
krlmlr, line 196: We can't use `|>` yet, for compatibility. Seems cleaner to introduce a new variable name (and perhaps keep `dm` unchanged).

```suggestion
  ptype_dm <- collect(dm_ptype(dm))
```

krlmlr (commented):
The PR looks good, I wonder why the tests fail.

krlmlr:
Thanks!

PR #1739, opened 2023-01-05
feat: `build_copy_queries()` and `db_learn_from_db()` improvements

samssann:
- `copy_dm_to()` now works with fks that have `on_delete` argument set as `cascade` (closes #1677)
- `dm_get_all_fks()` now correctly shows the `on_delete` column (closes #1677)
- `dm_get_all_pks()` now correctly shows the `autoincrement` column for MSSQL, MariaDB and Postgres (closes #1642)

PR #1904, opened 2023-06-19
copy_dm_to split into atomic processes

jangorecki:


R/db-interface.R
```diff
@@ -187,15 +187,14 @@ copy_dm_to <- function(dest, dm, ...,
...
-  dm ## we still need it to copy actual data, we will collect tables one by one to reduce peak memory usage
   ptype_dm <- collect(dm_ptype(dm))
 
   # Shortcut necessary to avoid copying into .GlobalEnv
   if (!is_db(dest)) {
     return(dm)
   }
 
-  queries <- build_copy_queries(dest_con, ptype_dm, set_key_constraints, temporary, table_names_out)
+  queries <- build_copy_queries(dest_con, dm, set_key_constraints, temporary, table_names_out)
```
krlmlr, line 197: In the end, this should be `dm_sql()` , and the code below should be simplified. Everything needed to run the queries should be contained (including the ticker messages) in the returned object.

krlmlr (commented):
Thanks, this is great!

- The failures on duckdb, maria and postgres look interesting, not sure what happens. Are the tables populated in the right order?
- I'm merging #1739, it's been out there for too long.
- Can you please add snapshot tests for the new functions? Instead of relying on `my_test_src()`, can you please use `simulate_mssql()` and friends so that these tests are less reliant on the database?
- I made improvements to the `Makefile` and already merged into this branch. Please edit `.github/CONTRIBUTING.md` as needed, can be part of this PR or a separate PR.

krlmlr:
@TSchiefer @moodymudskipper: Can you please take on the next review round?

jangorecki:
> The failures on duckdb, maria and postgres look interesting, not sure what happens. Are the tables populated in the right order?

it wasn't, fixed in [1eb7678](https://github.com/cynkra/dm/pull/1904/commits/1eb7678f05ea8fe8edb5e71d16a8b14e3fdd7198)

> Can you please add snapshot tests for the new functions? Instead of relying on my_test_src(), can you please use simulate_mssql() and friends so that these tests are less reliant on the database?

will do

krlmlr:
The fk violation won't be an issue when we add foreign keys via `ALTER TABLE`, but we're not there yet, and it's only a priority for the next step, once this is merged.

tests/testthat/_snaps/dm.md
```diff
@@ -108,8 +108,18 @@
     Code
       writeLines(conditionMessage(expect_error(dm(dm_for_flatten(),
       dm_for_filter_duckdb()))))
-    Output
-      All `dm` objects need to share the same `src`.
+    Condition
+      Warning in `dm_for_filter_duckdb()`:
+      restarting interrupted promise evaluation
+      Warning in `dm_for_filter_duckdb()`:
+      restarting interrupted promise evaluation
```
krlmlr, line 115: I don't think this is right -- `dm_for_filter_duckdb()` should at least work, the error indicates that the insertion order still isn't quite right.

R/build_copy_queries.R
```diff
@@ -220,6 +223,12 @@ build_copy_queries <- function(dest, dm, set_key_constraints = TRUE, temporary =
...
 
+  queries$dml <- dm_dml_load(
+    dm, con,
+    ## queries DF order does not match to table_names order so we need to match to order here
+    table_names = table_names[match(queries$name, names(table_names))]
+  )
+
   ## Reorder queries according to topological sort so pks are created before associated fks
   graph <- create_graph_from_dm(dm, directed = TRUE)
   topo <- igraph::topo_sort(graph, mode = "in")
```
krlmlr, line 234: This code and below computes the correct creation and insertion order. Perhaps this part should move further up in the function, we could reorder the tables in the dm object according to that order.

jangorecki:
Some of the failures like `Constraint Error: NOT NULL constraint failed: tf_1_11_2023_06_22_10_57_04_958379_6511.a` are caused by https://github.com/cynkra/dm/issues/1909
Code currently relies on the attribute autoincrement field from `dm_get_all_pks_impl()`, when TRUE it skips inserting into PK column.

R/build_copy_queries.R
jangorecki: This is not relevant because queries are not executed at that point, so reordering here is fine.

R/copy_dm_to.R
```diff
@@ -0,0 +1,169 @@
...
+#' @examplesIf rlang::is_installed("RSQLite") && rlang::is_installed("dbplyr")
+#' con <- DBI::dbConnect(RSQLite::SQLite())
+#' dm <- dm_nycflights13()
+#' s <- dm_sql(dm, con)
+#' s
+#' DBI::dbDisconnect(con)
+dm_sql <- function(dm,
+                   dest,
+                   set_key_constraints = TRUE,
+                   table_names = set_names(names(dm)),
```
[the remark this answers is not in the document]
jangorecki: Renaming tables (from `name` to `remote_name`) is more problematic because we lose original `name` inside `dm` object.
We still need it match autoincrement info from `dm_get_all_pks_impl()` because INSERT statements depends on autoincrement value.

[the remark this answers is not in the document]
jangorecki: I pushed code towards this change, mostly commented out, in https://github.com/cynkra/dm/pull/1904/commits/636637f3e1081cddcb2e1d8a5c61788ddea78463

Another option is to introduce table "labels" or "remote_name" inside `dm` object. Even for local table.

PR #2022, opened 2023-10-08
feat!: `copy_dm_to()` uses `dm_sql()`. Unique keys (#1887) and autoincrement primary keys (#1725) are created on the database. Data models with cyclic references are supported on databases that allow adding constraints in `ALTER TABLE` statements (at this time, all except DuckDB and SQLite, #664)

### LINKED ISSUE #664: copy_dm_to() and cyclic foreign key constraints
A topo sort is now required in `copy_dm_to()` after #658, we should offer an escape hatch.

### LINKED ISSUE #847: Creation of empty tables on PG DB works but throws error
IIRC, roughly the workflow is the following when calling `copy_dm_to()`:
1. create empty tables on DB
1. set keys
1. insert data

I would like to copy only empty tables, to prepare a DB for later insertion of data.
The result after the call is the desired one: the tables are there and key relations are all set, but the function runs into a dbAppendTable()-call which fails.

I think even if there is another preferrable method for doing this, we could  make this work.

### LINKED ISSUE #1385: copy_dm_to(copy_to = ) is defunct
without a really good alternative.

### LINKED ISSUE #1725: copy_dm_to() should support AI PKs
(see https://github.com/cynkra/dm/pull/1654#issuecomment-1351745127: copy_dm_to() should eventually be able to create and fill data when one or more tables with AI PKs are to be copied)

still left to do after https://github.com/cynkra/dm/pull/1689, https://github.com/cynkra/dm/pull/1698 & co.

### LINKED ISSUE #1887: UKs not being set to DB
The `dm_add_uk()` documentation states: "When a local dm is copied to a database (DB) with `copy_dm_to()`, a PK will be set on the DB by default, whereas a UK is being ignored."

Is there a technical reason why UKs are not being set to the DB, or is this work just something not prioritised high? I could have a look at this, if this is on your roadmap.

krlmlr:
- Closes #1887.
- Closes #1725.
- Closes #847.
- Closes #664.
- Closes #1385.


krlmlr:
@nbenn @TSchiefer @moodymudskipper: Can you please test this branch to see if it works for your projects that use dm? I'd like to send it to CRAN tomorrow.

There are breaking changes to `copy_dm_to()`, but we gain `dm_sql()` and friends that allow finer-grained control over the SQL that writes a dm to a database.

krlmlr:
I can now run the tests locally. Thanks for your work on this!

Issue #1735, opened 2023-01-04
`DiagrammeR` package is a dependency that is not automatically loaded with `dm`


DOH-RPS1303:
```
install.packages("dm")

dm <- dm_nycflights13()

dm
```
Fine up to here:
```
── Metadata ──────────────────────────────────────────────────
Tables: `airlines`, `airports`, `flights`, `planes`, `weather`
Columns: 53
Primary keys: 4
Foreign keys: 4
```
Here's where the issue happens:
```
dm %>%
+   dm_draw()
```
Which returns:
```
Error:
! `dm_draw()` needs the 'DiagrammeR' package. Do you need `install.packages("DiagrammeR")` ?
Run `rlang::last_error()` to see where the error occurred.
```
I already solved this problem for myself by installing the package, but given that this is a necessary dependency for a pretty big part of this package and that it's one of the example code snippets, it'd be nice to have DiagrammeR automatically install with dm

```

krlmlr:
Thanks. This is difficult because CRAN limits the number of packages we can import to 20. We're at 17 now. Theoretically, there's room, and I agree that DiagrammeR might be just important enough to warrant inclusion.

The other issue is that it's quite a heavy dependency, but so are many other packages we're already importing.

Would you like to contribute a PR?

krlmlr:
Thanks. I don't follow: `register_pkgdown_methods()` has an effect only if the `IN_PKGDOWN` environment variable is set. How does that break `R CMD build` ?

Good catch about validation. Has DiagrammeR been validated at some point?

pawelru:
> Thanks. I don't follow: register_pkgdown_methods() has an effect only if the IN_PKGDOWN environment variable is set. How does that break R CMD build ?

I am sorry, I confused CI steps. You are definitely right - this is a `pkgdown` step that failed for me - not R CMD BUILD. Nevertheless, my request stayed the same - to allow draw-free usage without aforementioned packages.

> Has DiagrammeR been validated at some point?

Not until now. At least within an internal env that I am working on (maybe it was not requested up until now?).

krlmlr:
This fell through the cracks, I didn't see an action item here. I believe you can use dm without installing DiagrammeR. What fails for you?

PR #2036, opened 2023-10-10
chore: Improve handling of suggested packages, add link in doc

olivroy:
Addresses #1735, closes #2035.

Briefly, it uses `check_installed()` in the check_suggested function to make it easier to install DiagrammeR. I guess `check_suggests()` could be simplified further. (and get the function by chaining rather than naming.

Mind the small tweak in tests to avoid skipping the test. We have to make it believe that we are not testing to avoid skipping the test. and actually record the message.

Note that check_installed()` errors in non-interactive sessions.

In an unrelated way, I added a link to the data frame vignette in the dm_nyc13 flights as I had issues to get started and this vignette helped me.

### DIFF
--- a/R/check_suggested.R
+++ b/R/check_suggested.R
@@ -42,13 +42,17 @@ check_suggested <- function(packages,
   # If TRUE, fail if package is not installed
   for (pkg in packages) {
     if (!is_installed(pkg)) {
-      if (is.null(message)) {
-        message <- glue("`{top_level_fun}()` needs the '{pkg}' package. Do you need `install.packages(\"{pkg}\")` ?")
-      }
       if (is_testing()) {
+        if (is.null(message)) {
+          message <- glue("`{top_level_fun}()` needs the '{pkg}' package.")
+        }
         testthat::skip(message)
       } else {
-        abort(message)
+        if (is.null(message)) {
+          message <- glue("to use `{top_level_fun}()`.")
+        }
+        # Could use the version argument too.
+        rlang::check_installed(packages, message)
       }
     }
   }
--- a/R/nycflights13.R
+++ b/R/nycflights13.R
@@ -3,8 +3,8 @@
 #' @description
 #' Creates an example [`dm`] object from the tables in \pkg{nycflights13},
 #' along with the references.
-#' See [nycflights13::flights] for a description of the data.
-#' As described in [nycflights13::planes], the relationship
+#' See [`nycflights13::flights`] for a description of the data.
+#' As described in [`nycflights13::planes`], the relationship
 #' between the `flights` table and the `planes` tables is "weak", it does not satisfy
 #' data integrity constraints.
 #'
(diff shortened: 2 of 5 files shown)


krlmlr:
On a related note, `rlang::check_installed()` could be doing what we're doing here -- skip in testing mode and when testthat is installed.

olivroy:
In tests, I believe that it would only error. You would need `skip_if_not_installed()`, but from my local tests, `skip_if_not_installed()` seems to be everywhere it needed to be. And the skipping from the check_suggested function did not seem to be needed.

krlmlr:
I'd rather remove all `skip_if_not_installed()` calls from tests, unless the package is something that is needed only for testing. That's where I see the beauty of our `check_suggested()` implementation -- the checks also (indirectly) check that all routines check for all packages they require.

krlmlr:
See the battery of the "Without xxx" tests here. We're removing suggested packages, one by one, and check that the tests still pass.

olivroy:
Maybe something like that could be useful. But I don't know much about chaining.

```r
f <- function(pkg) {
  rlang::local_interactive(FALSE)
  rlang::try_fetch(
    rlang::check_installed("rsvg2"),
    error = function(e) {rlang::inform("The function is enhanced by the following package, ",parent = e, call = rlang::caller_fn())})
  invisible()
}
f()

The function is enhanced by the following package, 
Caused by error in `f()`:
! The package "rsvg2" is required.
```

krlmlr:
I don't follow.

Would you like to look into removing the loop from the checking routine? I'm even proposing to have `check_installed()` call `skip()`, as you noted, this is a FR for rlang. If not, I'm happy to make this a "standalone", to be added via `usethis::use_standalone()` -- I'm seeing myself applying this pattern everywhere.

olivroy:
I could look into it. I may have time next week.

the idea would just be to add an optional parameter
```r
check_installed("DiagrammeR")
#> The package is required.
check_installed("DiagrammeR", optional = TRUE)
#> The package "DiagrammeR" can be used to enhance the function.
```

But I browsed a little through the use, and most usages in dm seem to be replaceable by `check_installed()` as the functions require the package to function, and error otherwise.


``` r
check_suggest <- function(pkg, call = rlang::caller_env()) {
rlang::local_interactive(FALSE) # To avoid the prompt

rlang::try_fetch(
rlang::check_installed(pkg, reason = "to have an enhanced functionality.", call = call),
error = function(e) {rlang::inform("The function is enhanced by the following package, ", parent = e, call = call)})
invisible()
}
g <- function(x) {
check_suggest(c("ggplot2 (>= 3.5.0)", "rsvg", "sss"))
x**2
}
g(2)
#> The function is enhanced by the following package, 
#> Caused by error in `g()`:
#> ! The packages "ggplot2" (>= 3.5.0) and "sss" are required to have an
#>   enhanced functionality.
#> [1] 4
```

<sup>Created on 2023-10-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The only thing that would need to be changed here is to change required to optional.

krlmlr:
Thanks!

`check_installed()` could be improved following what we do here. I propose we straighten our implementation here, make it a "standalone", and then propose an update to rlang. Happy to review more PRs.

Issue #1348, opened 2022-07-22
Can `check_suggested()` check multiple packages at once?


krlmlr:
See #1344.

olivroy:
#2036 may be a solution.

Because `rlang::check_installed(c("rlang (>= 1.1.1)", "vctrs"), reason = "to use a function", call = error_call)` is vectorized and chained.

krlmlr:
This looks good now. Thank you for your help!

PR #2040, opened 2023-10-13
chore: Improve suggested packages handling

### LINKED ISSUE #2039: Remove `message` argument to `check_suggested()`
Not used.

olivroy:
Close #2039 . (possibly enough for #1735 imo)

It could potentially
Use only loop in testing.

Also, for `dm_gui()`, handle suggested at once will improve the UX significantly.

```r
# Now
> dm_gui(dm = dm)
ℹ The packages "colourpicker", "shinyAce", and "shinydashboard" are required to use `dm_gui()`.
✖ Would you like to install them?

1: Yes
2: No
```

I also return `TRUE` early if all packages are installed. Is that correct?


olivroy:
I just added \pkg to make tests pass

krlmlr:
Thank you for your help. Can you please split this PR into two?

https://code-review.tidyverse.org/author/focused.html

olivroy:
Is there a quick way to do this? I don't know the tricks of git.

In any case, you can always revert the last commit I made. It was just a quick way to pass CI replace {tidyr} by \pkg{tidyr} to suppress a cran note.

krlmlr:
Locally:

```
git checkout -b b-fix-ci
git cherry-pick suggests-follow
```

This cherry-picks the last commit, this is the one I'd like to merge first. Then, open a new pull request.

After that:

```
git checkout suggests-follow
git reset --hard HEAD^
```

This erases the last commit.

For more complex splitting, interactive rebase is helpful.

olivroy:
Thanks for the tip. 

I will try to do that tomorrow if I have time.

olivroy:
@krlmlr It didn`t enterily work. The `git pull` command made me pull all branches from upstream to my branch.

The last git reset doesn't seem to do much to the suggests-follow branch.

Anyhow, I opened #2042 !

``` 
git reset --hard HEAD^
fatal ambiguous argument HEAD
unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
 'git <command> [<revision>...] -- [<file>...]'
```

olivroy:
From my testing, I verified #1348. Indeed, `check_installed()` can work with multiple packages.

R/check_suggested.R
```diff
@@ -39,22 +36,36 @@ check_suggested <- function(packages,
...
-        }
+  # Return early if all packages are installed.
+  if (rlang::is_installed(packages, version = version)) {
+    return(TRUE)
+  }
+
+  # Skip if some packages are not installed when testing
+  # And say which package was not installed.
+  if (is_testing()) {
+    for (pkg in packages) {
```
krlmlr, line 47: Once we have a logical `installed` vector, we can do `packages[!installed]` and bundle the result in a single `message()` or `skip()` call, removing the `for` loops.

R/check_suggested.R
```diff
@@ -39,22 +36,36 @@ check_suggested <- function(packages,
...
 
-  # If TRUE, fail if package is not installed
-  for (pkg in packages) {
-    if (!is_installed(pkg)) {
-      if (is_testing()) {
-        if (is.null(message)) {
-          message <- glue("`{top_level_fun}()` needs the '{pkg}' package.")
-        }
+  # Return early if all packages are installed.
+  if (rlang::is_installed(packages, version = version)) {
```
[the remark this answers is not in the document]
olivroy: > This early exit will work regardless of the value of use ?

Yes.


> I don't like the version argument in this function, in particular since version specifications can be passed to the pkg argument of check_installed() . Should we do an early installed <- map_lgl(...) and take it from there?

`version` must be a vector of the same length as `packages`. will error otherwise. This is how `rlang::check_installed()` works.


If we have a loop then, it would have to be `version[i]` indeed.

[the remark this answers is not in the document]
krlmlr: Let's remove the `version` argument here.

``` r
rlang::check_installed("dplyr (>= 2.0.0)")
#> Error:
#> ! The package "dplyr" (>= 2.0.0) is required.
```

<sup>Created on 2023-10-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/check_suggested.R
olivroy: A prototype function would be the following.
```r
  installed <- purrr::map2_lgl(packages, version, function(pkg, ver) {
    is_installed(pkg, version = ver)
  }
```

krlmlr: Let's get rid of the `version` argument in our function if we don't need it.

olivroy: Yes. and use `check_installed()` syntax instead. I will do this.

R/check_suggested.R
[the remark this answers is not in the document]
olivroy: I did this.

olivroy:
There may be some code duplication, but this seems to work as intended now. getting rid of version, and message argument. 

I removed the `install.packages()` bit because some version may be attached to it, which would create invalid code.

R/check_suggested.R
```diff
@@ -10,27 +10,25 @@
...
-        if (is.null(message)) {
-          # FIXME: Mention version
-          message <- glue("`{top_level_fun}()` is improved by the '{.}' package. Consider `install.packages(\"{.}\")`.")
-        }
-        inform(message)
-      }
-      installed
+    installed <- map_lgl(packages, function(pkg) {
+      is_installed(pkg)
     })
```
krlmlr, line 24:
```suggestion
    installed <- map_lgl(packages, is_installed)
```

Can this move even higher, outside `if (!use)` ? We could return immediately if all packages are installed, in all cases.

R/check_suggested.R
```diff
@@ -10,27 +10,25 @@
...
+
+    if (any(!installed)) {
```
krlmlr, line 26:
```suggestion
    if (!all(installed)) {
```

R/check_suggested.R
```diff
@@ -10,27 +10,25 @@
...
+      pkgs_not_installed <- packages[!installed]
+      message <- "{.fn {top_level_fun}} is improved by the '{.pkg {pkgs_not_installed}}' package. Consider installing them."
```
krlmlr, line 28: How can we distinguish between singular and plural here? What happens if more than one package is not installed?

R/check_suggested.R
```diff
@@ -39,22 +37,34 @@ check_suggested <- function(packages,
...
+  if (rlang::is_installed(packages)) {
+    return(TRUE)
+  }
+
+  # Skip if some packages are not installed when testing
+  # And say which package was not installed.
+  if (is_testing()) {
+    installed <- map_lgl(packages, function(pkg) {
+      is_installed(pkg)
+    })
```
krlmlr, line 50:
```suggestion
```

Redundant now?

R/check_suggested.R
```diff
@@ -39,22 +37,34 @@ check_suggested <- function(packages,
...
-          message <- glue("to use `{top_level_fun}()`.")
-        }
-        # Could use the version argument too.
-        rlang::check_installed(packages, message)
-      }
+  # Return early if all packages are installed.
+  if (rlang::is_installed(packages)) {
+    return(TRUE)
+  }
+
```
krlmlr, line 44:
```suggestion
```

Not needed if we do an early return?

tests/testthat/_snaps/check_suggested.md
```diff
@@ -37,8 +37,16 @@
...
     Condition
       Error:
-      ! The package "iurtnkjvmomweicopbt" is required not installed!
+      ! The package "iurtnkjvmomweicopbt" is required.
+    Code
+      rlang::local_interactive(TRUE)
+      check_suggested(c("iurtnkjvmomweicopbt (>= 0.5)", "xxx", "cli"), NA,
+      top_level_fun = "foo")
+    Message
+      `foo()` is improved by the 'iurtnkjvmomweicopbt (>= 0.5) and xxx' package. Consider installing them.
```
krlmlr, line 49:
```suggestion
      `foo()` is improved by the 'iurtnkjvmomweicopbt (>= 0.5)' and 'xxx' packages. Consider installing them.
```

Can we achieve this output?

tests/testthat/_snaps/check_suggested.md
```diff
@@ -37,8 +37,16 @@
       Error:
       ! The package "iurtnkjvmomweicopbt" is required to use `foo()`.
     Code
-      check_suggested("iurtnkjvmomweicopbt", TRUE, message = "not installed!")
+      check_suggested("iurtnkjvmomweicopbt", TRUE)
     Condition
       Error:
-      ! The package "iurtnkjvmomweicopbt" is required not installed!
+      ! The package "iurtnkjvmomweicopbt" is required.
```
krlmlr, line 43: The quotes used here are inconsistent with the quotes in the other error message.

krlmlr (commented):
Thanks, cool!

b6f5382c016199300e86f5fdbccb5fab3ab06eb4 adds a lot of noise here, can you please send this as a separate PR?

tests/testthat/_snaps/check_suggested.md
olivroy: Will use `"` everywhere.

R/check_suggested.R
olivroy: I tried playing with the order without chaning test results, but it didn't work.

R/check_suggested.R
olivroy: Done. https://cli.r-lib.org/reference/pluralization.html

olivroy:
Reverted the commit + added cli pluralization. Not sure about changing the order as it may result in different results, depending on which packages are installed on the system.

R/check_suggested.R
olivroy: Possibly in a follow-up because this PR aims not to change the default behavior. (except gor removing the version and message argument from check_suggested

krlmlr:
I force-pushed, if needed you can use

```sh
git fetch
git reset --hard origin/suggests-follow
```

to sync locally.

Thanks for recognizing a real shortcoming in our approach to handling dependencies and for pushing through!

olivroy:
Glad you like it and that you validated what I did. Always enriching. Tests still pass locally for me after changes you made.

krlmlr:
Looking into `?check_installed`, I wonder if we could make do with one single `try_catch(check_installed(...), ...)` call. I suspect this won't make the code easier to understand, to debug, or to enhance -- I suggest calling it a pull request and see how this works in practice before changing further.

olivroy:
> Looking into `?check_installed`, I wonder if we could make do with one single `try_catch(check_installed(...), ...)` call. I suspect this won't make the code easier to understand, to debug, or to enhance -- I suggest calling it a pull request and see how this works in practice before changing further.

```r
check_suggest <- function(pkg, call = rlang::caller_env()) {
  rlang::local_interactive(FALSE) # To avoid the prompt
  
  rlang::try_fetch(
    rlang::check_installed(pkg, reason = "to have an enhanced functionality.", call = call),
    error = function(e) {rlang::inform("The function is enhanced by the following package, ", parent = e, call = call)})
  invisible()
}
g <- function(x) {
  check_suggest(c("ggplot2 (>= 3.5.0)", "rsvg", "sss"))
  x**2
}
g(2)
#> The function is enhanced by the following package, 
#> Caused by error in `g()`:
#> ! The packages "ggplot2" (>= 3.5.0) and "sss" are required to have an enhanced functionality.
[1] 4
```

An example I shared earlier that could be what you were thinking about? But as I pointed out, the message shared by `check_installed()` is not 100% correct for the suggested enhancements.

krlmlr:
Thanks for the example, I missed that. The condition object `e` has elements that contain the packages and versions that were missing, so in theory we could put together the same message, but I don't think it's worth the effort at this time.

olivroy:
Tests are failing. This is  exactly what I was referring to in https://github.com/cynkra/dm/pull/2040#issuecomment-1765287647
```r
    check_suggested("dm", FALSE, top_level_fun = "foo")
  Output
-   [1] FALSE
+   [1] TRUE
```

Hence why, if `use = FALSE`, it would be safer to return `FALSE`

krlmlr:
Thanks!

PR #2183, opened 2024-01-18
chore: Oracle database integration to compose file

aavanesy:
Adding necessary lines for the oracle database (free version) to the compose file.

Current free image is hosted on `docker pull container-registry.oracle.com/database/free:latest`

Documentation on the docker image can be found on 
https://www.oracle.com/database/technologies/oracle-database-software-downloads.html#db_free

Versions from tutorials such as `docker pull store/oracle/database-enterprise:12.2.0.1` do not work.
Oracle has discontinued the hosting of these images on docker hub.


aavanesy:
Added a bash script that adds users with all permissions. 
I tried without success to create a user on image creation stage or during docker compose. 
Sleep is set at 30 seconds but it is less, just a defensive number

Also added 
```
db-reset:
	docker-compose down --volumes
	docker-compose build --no-cache
```
I noticed during compose ups that modifications don't show up after restarts. Cleaning cache and volumes helped.

krlmlr:
Solved with external repository, thanks!

PR #2184, opened 2024-01-19
chore: Fix withr 3.0.0 compatibility

krlmlr:


krlmlr:
@olivroy: Looks better now?

olivroy:
All good. This is a surprising behavior, do you think it is worth an issue? I wouldn't know if it is a withr, rlang, or testthat issue however.

krlmlr:
If it's a regression with withr (i.e., something that occurs only with withr 3.0.0), it's worth filing there.

PR #2203, opened 2024-03-28
chore: Use correct option to control verbosity of lifecycle messages in unit tests

IndrajeetPatil:
- Old option deprecated in https://tidyselect.r-lib.org/news/index.html#tidyselect-120
- New option documented in https://lifecycle.r-lib.org/reference/verbosity.html


tests/testthat/helper-testthat.R
```diff
@@ -1,3 +1,2 @@
-options(testthat.progress.verbose_skips = FALSE)
```
IndrajeetPatil, line 1: This was removed in https://github.com/r-lib/testthat/commit/6cbd8190ac283a639ed7cff975f90f2c4d5db16d

PR #2214, opened 2024-05-03
fix: pass naming pattern through to dm_learn_from_db() (#2213)

### LINKED ISSUE #2213: Naming pattern not used in `dm_from_con()` when keys are learned successfully
A new `.names` parameter was added to `dm_from_con()` in v1.0.6 (https://github.com/cynkra/dm/pull/1790) - but it's currently ignored when `learn_keys = TRUE` and the subsequent key-querying is successful.

That's because `dm_from_con()` [returns early](https://github.com/cynkra/dm/blob/e52ba91e26b04a0332039f8754054cb246542f7e/R/dm_from_con.R#L80) if key-querying, via `dm_learn_from_db()`, succeeds.

Helpfully, `dm_learn_from_db()` already accepts a `names_pattern` argument, which is used to format the table names with {glue} in the same way as elsewhere - but we don't pass `.names` through to it in `dm_from_con()`:
https://github.com/cynkra/dm/blob/e52ba91e26b04a0332039f8754054cb246542f7e/R/dm_from_con.R#L71

---

For what it's worth, if key-querying fails, then later in `dm_from_con()` we _do_ pass `.names` through to `get_src_tbl_names()`:
https://github.com/cynkra/dm/blob/e52ba91e26b04a0332039f8754054cb246542f7e/R/dm_from_con.R#L107

owenjonesuob:
Fixes #2213.

I had to move the "smart default" logic for `.names` up one level, from `get_src_tbl_names()` into `dm_from_con()` - but I think that is a better place to keep it anyway, since that's where its behaviour is documented.

I've also made sure `names_pattern` is used consistently as the "internal" name for this argument, wherever we see this "naming pattern" concept.


krlmlr:
@claude: Fix tests.

krlmlr:
@copilot: Fix tests.

krlmlr:
@owenjonesuob: Thank you for your patience.

Issue #1678, opened 2022-11-20
render dm visual in R Shiny


sound118:
Is it possible we can render the visual generated by dm_draw() in R Shiny session?

krlmlr:
Thanks. This has worked in the past, what have you tried?

One alternative is to create an SVG version with the DiagrammeRsvg package.

sound118:
Thanks for the quick reply. I am still studying DM package, will try the approach you mentioned above. By the way, from the github page, it’s said DM package is compatible with postgre database. We are using AWS redshift as our data storage and redshift is essentially built based on postgre. Is DM package currently also compatible with redshift DB?

krlmlr:
I've never tried it with Redshift. I suspect that collecting information on relationships via `dm_from_con(learn_keys = TRUE)` might not work, but otherwise chances are good.

sound118:
@krlmlr 

> Thanks. This has worked in the past, what have you tried?
> 
> One alternative is to create an SVG version with the DiagrammeRsvg package.

Hi Krill, how do I render the SVG object after using export_svg from DiagrammeRsvg package in Shiny?

sound118:
Hi Krill, 
The shiny app deployed with "dm_gui" in R Studio server looks awesome. However, I wanna deploy it in another cloud based server (say shinyapps.io). How can I actually do it? Actually, I wanna reutilize your designed shiny app and with a few parameters input to form "dm" object and feed into dm_gui function to run in shinyapps.io. Is there a quick way to do it? I've tried to drill into your package code, trying to run it directly with ui.R and server.R . It seems not so easy to implement.

krlmlr:
This fell through the cracks. Curious to see what you achieved so far. How can we help?

PR #2215, opened 2024-05-03
feat: add support for Redshift connections

owenjonesuob:
An attempt to add support for [Amazon Redshift](https://aws.amazon.com/redshift/) connections.

Redshift is based on Postgres - which is great, because {dm} already has excellent Postgres support :grin:

So for the most part, we can just borrow existing functionality.

---

There _are_ some differences between Redshift and Postgres - more detains in the [AWS docs](https://docs.aws.amazon.com/redshift/latest/dg/c_redshift-and-postgres-sql.html). One note in particular caught my eye:

> Unique, primary key, and foreign key constraints are permitted, but they are informational only. They are not enforced by the system, but they are used by the query planner. - https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-features.html

I'm not sure if this has any impact on whether we can build a dm by learning the database keys.

I also tried, but failed, to write a Redshift-friendly version of `postgres_column_constraints` since Redshift doesn't support some of the syntax used there (e.g. `array_position()`). This affects `dm_meta_raw()` - for now I have added an implementation based on the final `else` clause.

---

Testing any Redshift functionality directly might be tricky, since it would require a Redshift warehouse in an AWS organisation - i.e. a lot of additional infrastructure (and cost?!). Would we be happy to rely on the fact that it's "nearly Postgres", and therefore mostly covered by Postgres testing? Or can we think of some way to test directly?

---

Requested indirectly in #1678.

### DIFF
--- a/R/build_copy_queries.R
+++ b/R/build_copy_queries.R
@@ -83,7 +83,7 @@ build_copy_queries <- function(dest, dm, set_key_constraints = TRUE, temporary =
       pk_col <- pk_col[[1]]
 
       # Postgres:
-      if (is_postgres(dest)) {
+      if (is_postgres(dest) || is_redshift(dest)) {
         types[pk_col] <- "SERIAL"
       }
 
--- a/R/db-helpers.R
+++ b/R/db-helpers.R
@@ -55,6 +55,12 @@ is_postgres <- function(dest) {
   ))
 }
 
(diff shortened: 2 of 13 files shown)


krlmlr:
Thanks! I have no easy way to test this, I trust this works for you.

Issue #365, opened 2020-05-15
make more use of 'with_mock()' in tests


TSchiefer:
Use `with_mock()` for redefining `tbl_sum.tbl_sql` et al. (cf. helper-print.R) in order to avoid consequences for session when using `testthis::test_this()`

krlmlr:
No priority.

PR #2220, opened 2024-06-18
chore: Drop crayon and mockr dependencies

olivroy:
Small PR to drop mockr in favour of `local_mocked_bindings()` and crayon in favour of cli.

relates to #365


krlmlr:
Thanks!

PR #2245, opened 2024-11-21
docs: Avoid topic clashes for `?mutate` and other verbs

krlmlr:
@maelle: Looks like `@noRd` is the better choice here, need to remember that for duckplyr.


maelle:
@krlmlr why did you choose to stop re-exporting here, should duckplyr also stop re-exporting?

krlmlr:
Good question, not sure yet. On the upside, this problem is independent of the help pages.

PR #2256, opened 2024-11-27
fix: Ensure compatibility with upcoming update of the pixarfilms package

erictleung:
New data has been added to the pixarfilms package and this change updates the genres schema accordingly.

Closes https://github.com/erictleung/pixarfilms/issues/39

### DIFF
--- a/R/dm_from_con.R
+++ b/R/dm_from_con.R
@@ -71,9 +71,9 @@ dm_from_con <- function(
       {
         dm_learned <- dm_learn_from_db(con, ...)
         if (is_null(learn_keys)) {
-          inform(c(
+          cli::cli_inform(c(
             "Keys queried successfully.",
-            i = "Use `learn_keys = TRUE` to mute this message."
+            i = "Use {.code learn_keys = TRUE} to enforce querying keys and to mute this message."
           ))
         }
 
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks. It looks like this needs more work. I have other material that relies on the fact that there's a film with no title in the current data. I'm thinking about copying the current state of the data here, perhaps pixarfilms can have a local, up-to-date version of `dm_pixarfilms()` . What do you think?

erictleung:
@krlmlr ah yes, I think you mentioned something about materials relying on missing data. The missing data happened to be for that one particular time point where Pixar didn't realize the name of the film yet.

And I'll try to occasionally keep the data up to date as Pixar will release more movies. So I'd agree that you can simply copy the current state of the data for dm and then I can simply write an up-to-date dm_pixarfilms() within pixarfilms. I already have a draft of it in a vignette https://github.com/erictleung/pixarfilms/blob/main/vignettes/use_dm.Rmd.

Feel free to close this PR and copy data from, say, the last release tag (https://github.com/erictleung/pixarfilms/releases/tag/v0.2.1), and then I'll create an up to date dm function in pixarfilms. Thanks!

PR #2280, opened 2025-04-05
corrected the font paths, right aligned the top navbar items

talpitoo:
This PR repeats the same patterns made in https://github.com/cynkra/cynkratemplate/pull/58 (once that is merged as the parent template).

### DIFF
--- a/pkgdown/_pkgdown.yml
+++ b/pkgdown/_pkgdown.yml
@@ -155,13 +155,13 @@ reference:
 navbar:
   structure:
     left:
+    right:
     - intro
     - reference
     - tutorials
     - cheatsheet
     - tech
     - news
-    right:
     - search
     - github
   components:
--- a/inst/cheatsheet.css
+++ b/inst/cheatsheet.css
@@ -15,7 +15,7 @@ pre {
 /* NOTE: 'new' frutiger normal=400 */
 @font-face {
     font-family: "frutiger";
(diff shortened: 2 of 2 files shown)


pkgdown/_pkgdown.yml
```diff
@@ -155,13 +155,13 @@ reference:
 navbar:
   structure:
     left:
+    right:
```
maelle, line 158: this I'm not sure the maintainer wants to change. that's something pkgdown website maintainers can change as they will. Unless it looks really awful with the current design?

krlmlr: I wonder if we can achieve the designer's intention by treating all "left" items as if they were "right" items. I'd rather not adapt `_pkgdown.yml` in every package that uses the template.

maelle: Well in that case we'd need to provide a wrapper around `pkgdown::build_site()` within cynkratemplate. That's what rotemplate, the rOpenSci's pkgdown template, does. But it means we'd have to tweak all package repos to use that... so in the end, easier to tweak YAML configs.

krlmlr: Please:

- No wrapper
- No config changes

Pretty sure we can adapt the template, let's discuss there.

maelle: Ok, let me tweak the cynkratemplate PR

krlmlr:
Let's restart as needed.

PR #2356, opened 2026-01-14
chore: Don't refer to removed `dplyr::src_dbi()`

DavisVaughan:
This has been defunct in dplyr for many years, and we've finally removed it.

dplyr will be released on January 31, 2026. If you could please send an update of your package to CRAN before then, that would help us out a lot! Thanks!


krlmlr:
Thanks!

PR #2477, opened 2026-07-21
ci: source dev-package deps from P3M, keep only target dev from r-universe

krlmlr:
## Problem

The scheduled `rcc dev` workflow (`.github/workflows/R-CMD-check-dev.yaml`) fails in the `rcc-dev: tidyr` matrix job. The surface error is:

```
ERROR: dependency ‘tidyr’ is not available for package ‘dm’
```

during `R CMD build` of `dm`. The real cause is one step earlier — the "Install dev version" step failed to install dev `tidyr`:

```
Installing 1 packages: tidyr
ERROR: dependency ‘cpp11’ is not available for package ‘tidyr’
* removing ‘/home/runner/work/_temp/Library/tidyr’
```

## Root cause

The "Install dev version" step ran a single `remotes::install_runiverse("<pkg>", linux_distro = "noble")`. `install_runiverse()` installs the requested dev package **without its dependencies**. Dev `tidyr` has `cpp11` as a `LinkingTo` dependency, and when r-universe has no ready binary for the pinned `noble` distro it serves the dev package as **source**, which must be compiled against `cpp11`. Because the install brought in no dependencies and `cpp11` was not otherwise in the check library, the source build of dev `tidyr` failed, `tidyr` was removed, and `dm`'s subsequent `R CMD build` could not find it.

The earlier `any::cpp11` `extra-packages` entry only papered over this by happening to place `cpp11` in the library — a coincidental workaround, not a fix. The actual bug is that the dev-install step installed the package without its dependencies.

## Fix

Rework the "Install dev version" step into three phases so dependencies are sourced correctly:

1. **Install only the target dev package from r-universe, `dependencies = FALSE`.** `linux_distro = "noble"` is kept so we still pull a fast *binary* build rather than compiling from source when one is available.
2. **Install any newly-introduced hard deps from P3M.** Read the freshly installed dev package's `Depends`/`Imports`/`LinkingTo` off `installed.packages()`, diff against what is already installed, and `install.packages()` the missing ones (e.g. `cpp11` for dev `tidyr`, `purrr` for dev `DBItest`) from Posit Package Manager. This brings dependencies in as *released* binaries instead of dragging dev versions of the whole tree from r-universe. `repos` is set explicitly because phase 1 may have pointed them at r-universe.
3. **Re-install the target dev package from r-universe with `dependencies = TRUE`.** Its hard deps are now satisfied from P3M, so this only ensures the dev package itself is in place and fills any residual gaps.

The `any::cpp11` `extra-packages` workaround added earlier has been **reverted** (`extra-packages` is back to `"any::rcmdcheck r-lib/remotes@f-618-universe ."`), since the deps are now handled in the install step itself.

### `noble` caveat

r-universe has renamed its Linux binary distro from `noble` to `resolute`, so the hardcoded `linux_distro = "noble"` can start returning 404 for some packages (it already broke a dev `tidyr` install). It is kept for now to preserve fast binary installs; it should be updated or dropped once it starts 404ing.

## Verification

`dm`'s `R CMD check` against released dependencies already passes in the `base` job. The `rcc dev` workflow only runs on schedule / tag / `workflow_dispatch`, so this fix is validated by a manual `workflow_dispatch` run on this branch rather than by PR checks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016BRYo2D5L4wpgaTFTHH2B7

### DIFF
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -118,7 +118,14 @@ jobs:
         with:
           cache-version: rcc-dev-${{ matrix.package }}-1
           needs: build, check
-          extra-packages: "any::rcmdcheck r-lib/remotes@f-618-universe ."
+          # cpp11 is a LinkingTo (header-only) dependency of several tidyverse
+          # packages (e.g. tidyr, dplyr).
+          # When r-universe has no ready binary and serves the dev package as
+          # source, it must be compiled, which fails unless cpp11 is present in
+          # the library.
+          # pak does not install LinkingTo dependencies of binary deps, so we
+          # request cpp11 explicitly here.
+          extra-packages: "any::rcmdcheck any::cpp11 r-lib/remotes@f-618-universe ."
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install dev version of ${{ matrix.package }}
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -118,21 +118,49 @@ jobs:
         with:
           cache-version: rcc-dev-${{ matrix.package }}-1
           needs: build, check
-          # cpp11 is a LinkingTo (header-only) dependency of several tidyverse
-          # packages (e.g. tidyr, dplyr).
-          # When r-universe has no ready binary and serves the dev package as
-          # source, it must be compiled, which fails unless cpp11 is present in
-          # the library.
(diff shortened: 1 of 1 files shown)


krlmlr:
Superseded by the Resolute CI migration on `main`. r-universe now serves `resolute` binaries, so dev `tidyr` installs as a binary and no longer needs its `cpp11` `LinkingTo` dep built from source — the failure this PR addressed no longer occurs. Rebasing onto current `main` leaves **no net change**, so I'm closing this as obsolete.

---
_Generated by [Claude Code](https://claude.ai/code)_

